PR #19, opened 2014-10-07
Use bash for install script and retain on failure

wmertens:
The actual install script needs bash, not sh. Use the provided bash from the binary closure.

Also, on failure, retain the downloaded files so the user can fix issues and run the script again. The actual install script prompts the user to do certain things like running things as root if needed.


domenkozar:
This is the exact opposite of https://github.com/NixOS/nixos-homepage/pull/62

domenkozar:
@wmertens this would need to be rebased now and the tarball should be retained based on some flag.

wmertens:
@domenkozar How about this? Reuse downloaded data and verify before using.

(untested)

wmertens:
Do you think the NIX_DL test needs more testing? If someone sets NIX_DL=/ that will result in a wiped root… Maybe safer to request manual deletion when reusing?

samueldr:
[triage]

This is now [part of the `nix` repository](https://github.com/NixOS/nix/blob/master/scripts/install.in).

The PR would need to be re-done there.

And yes, as of today, [the script does cleanup after itself](https://github.com/NixOS/nix/blob/9617a043541d77d79e4f20f9676aae63de72f45d/scripts/install.in#L15).

PR #24, opened 2014-12-18
Add version-specific install scripts.

aaronlevin:
I don't know anything about these `.tt` files or ASP, so this seemed the quickest / easiest way. I only went back to `1.7` because the other scripts don't take into account the `uname` info. 

My desire to have version-specific install scripts is because I am creating a Docker container and installing Nix within that container. The curl-style install is very convenient in a Docker situation. Unfortunately, `RUN` commands in Docker are cached if they don't change, so the command to install nix won't update even if nix has. However, if I hardcode the version in the install, then I can update the `RUN` command when nix's version is updated.

### DIFF
--- a/nix/install-1.7
+++ b/nix/install-1.7
@@ -1,0 +1,45 @@
+#!/bin/sh
+
+# This script installs the Nix package manager on your system by
+# downloading a binary distribution and running its installer script
+# (which in turn creates and populates /nix).
+
+{ # Prevent execution if this script was only partially downloaded
+
+unpack=nix-binary-tarball-unpack
+
+require_util() {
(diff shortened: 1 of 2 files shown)


domenkozar:
@aaronlevin thanks. I think it's better we use `pkgs.dockerTools` these days to build a Nix image for different Nix releases. And the installer script should always just take the latest version.

aaronlevin:
@domenkozar sounds cool. Thanks for replying to my quite old PR :) I will close!

domenkozar:
Just for reference, https://github.com/LnL7/nix-docker is WIP but hopes to achieve that.

PR #28, opened 2015-01-11
use https for search form submission

ryantm:
Change http://www.google.com/search form post to use https instead of http, which will, on some pages, fix the mixed content warning currently appearing. I have not tested this change works.


ryantm:
I tested it now; it works.

PR #30, opened 2015-01-23
Make options search work in non-ES6 browsers.

ragnard:
Use regex instead of ES6 String.prototype.startsWith method which isn't
available in all browsers.


darkehorselll:
I think this iz the root repository for nginx which is a debian OS . UNIEX OR UNIX
What would be the benifiet of merging you???

edolstra:
Thanks, applied!

PR #33, opened 2015-02-15
Mention newer nixpkgs releases in download.html

falsifian:
Is there a better place to link to here? The only list of releases I found was in the table of contents of the NixOS manual, under "Release Notes", because there just happens to be a release notes section for every release.

(Related question: do we have any other information about the release process documented? E.g. how long do they receive support?)

### DIFF
--- a/nixpkgs/download.tt
+++ b/nixpkgs/download.tt
@@ -44,6 +44,10 @@ continuous build system:</p>
 
 <section><h2>Old releases</h2>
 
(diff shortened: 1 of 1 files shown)


edolstra:
14.04, 14.12 etc. are NixOS release branches, not Nixpkgs release branches. (Meaning for instance that they don't include Mac OS X builds.) We don't currently have a stable release branch for Nixpkgs.

falsifian:
Ah, I see. It might be good to add some sort of note, because the current page makes it look like the second-last release was in 2010 or something, i.e. the project might be dead.

PR #36, opened 2015-03-31
update pkgs count

jagajaga:
Doing `nix-env -qa \* | uniq | wc -l` gives me `13142` pkgs now.


jagajaga:
ping

edolstra:
Counting in that way gives numerous duplicates (e.g. thousands of Python packages built for different Python versions).

jagajaga:
You are right.

PR #55, opened 2015-10-15
replace logo with a new hexagon design

gfxmonk:
I've kept the old .svg in case it's needed in the future, but have replaced the existing png exports.

Note: I haven't yet been able to build the site yet, so I can't see precisely how this looks. The new design is now wider than it is tall, so whatever HTML uses these logo may need updating to fit nicely.


gfxmonk:
I've now been able to build locally, and fixed up the layout to accommodate the slightly different size / spacing.

domenkozar:
\o/ Does that mean it's official?

PR #59, opened 2015-11-14
Restore the project menu

zimbatm:
This helps with the naviation and discoverability of the various projects.
Right now the user has to either follow the links or enter the right path in
the URL to navigate between these.

This tries to fix one of the complaint @garbas had during the talk.

screen shot 2015-11-15 at 00 32 14

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -26,8 +26,20 @@ p {
     hyphens: auto;
 }
 
+.brand > a {
+    color: #777;
+    text-decoration: none;
+}
+
+.brand > a:hover {
+    color: #333;
+    text-decoration: none;
+}
+
 .logo {
     margin-top: -5px;
+    padding-right: 5px;
+    line-height: 25px;
     height: 25px;
 }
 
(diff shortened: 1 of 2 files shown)


garbas:
+1 ... not yet the final version, but definitely the step into the right direction
@edolstra what do you think?

edolstra:
+1, I removed the old projects menu because it cluttered the navbar, but turning the NixOS logo into a dropdown solves that nicely. The only problem is that this changes the font of "NixOS" relative to the rest of the navbar. Could that be made consistent?

zimbatm:
Fixed the font's default and hover color.

Before / After:
screen shot 2015-11-19 at 18 49 57
screen shot 2015-11-19 at 18 49 44

The menu is part of the top-left logo so I think that it's fine if it's in a different size. In mobile view it also stands out like that:
screen shot 2015-11-19 at 18 52 44

edolstra:
Thanks, applied!

zimbatm:
Thanks, is it automatically deployed ?

edolstra:
Yes, every 15 minutes iirc.

PR #60, opened 2015-11-16
Update index.tt

ret:
Small oversight, I think.


edolstra:
Ping @svanderburg

svanderburg:
@edolstra Looks fine to me. However, I can't integrate it since I don't have the rights to do so :)

PR #68, opened 2015-12-20
installation script needs bash

phunehehe:
from https://github.com/NixOS/nix/issues/747


phunehehe:
alternatively, https://github.com/NixOS/nix/blob/master/scripts/install-nix-from-closure.sh might be made to use `sh`

PR #81, opened 2016-06-01
install.in: mktemp -d and reliably cleanup tarball dir on exit

aschmolck:
[Copied from #62, since the underlying repo for the PR went away]
This PR should make the default installation method a little bit more robust:

curl https://nixos.org/nix/install | sh
As it is, after a failed attempt to run the above (e.g. due to a keyboard interrupt or failed partial download) the second invocation of the same command will also fail because the nix-binary-tarball-unpack directory won't have been cleaned up and the attempt to mkdir it at the same location will error out.

This PR uses mktemp (in a way that should work at least on OS X and Linux) and an exit trap to fix this.

### DIFF
--- a/nix/install.in
+++ b/nix/install.in
@@ -5,19 +5,22 @@
 # (which in turn creates and populates /nix).
 
 { # Prevent execution if this script was only partially downloaded
-
-unpack=nix-binary-tarball-unpack
+oops() {
+    echo "$0: $@" >&2
+    exit 1
+}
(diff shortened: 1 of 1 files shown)


domenkozar:
I can test on Ubuntu, but I'll have to ask someone to test on OSX.

LnL7:
works fine on darwin

domenkozar:
Tested on Ubuntu and it installs successfully. No traces in `/tmp` if download is ctrl-c in the middle. Thanks!

PR #84, opened 2016-06-20
nix/install: Run the installer after download

aelsabbahy:
Fixes regression introduced by dc30713b14db73de61b85501aa1ce32d7e23fc34

@edolstra This is in reference to the comment I posted on your commit.


kevincox:
:+1: Actually running the script is quite important!

PR #85, opened 2016-06-25
Nix install script: Fix Bash-ism, quoting issues

8573:
In the Nix installation script at `nix/install.in`, replace a Bash-ism
with its POSIX-compatible equivalent, and fix some (minor) quoting
issues found by ShellCheck.

Credit to @wycats for hitting the Bash-ism, and thereby prompting me to
fix it.

NOTE: This will require the PGP signature at `nix/install.sig` to be
updated, which I don't believe I can do.

### DIFF
--- a/nix/install.in
+++ b/nix/install.in
@@ -6,7 +6,7 @@
 
 { # Prevent execution if this script was only partially downloaded
 oops() {
-    echo "$0: $@" >&2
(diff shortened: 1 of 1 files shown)


domenkozar:
cc @edolstra

8573:
Thanks for merging!

Issue #15, opened 2014-08-29
NixOs News RSS or Atom feed


bezirg:
I cannot find an RSS or Atom feed for the http://nixos.org/news.html

I found a /news.xml page, but this is apparently a custom format.

If you build such a feed, can you also add an RSS icon link somewhere on the homepage?

davidak:
that would be helpful.

also add links to the html head section that browsers displays it:

`<link rel="alternate" type="application/rss+xml"
   title="RSS" href="http://www.example.net/feed.rss" />`

`<link href="atom.xml" type="application/atom+xml" rel="alternate" title="Sitewide ATOM Feed" />`

charlieok:
:+1: Would like to get updates on NixOS in my feed reader. Also looked for a feed and didn't find one. Would subscribe if this were added.

ajgrf:
I wrote a small program to generate an Atom feed from the news.xml page. I run it locally but it would probably be easy to adapt if you wanted to offer an official feed. I posted it as a gist [here](https://gist.github.com/ajgrf/ce07620b7ba5c80583a9).

rexim:
:+1: Would like to stay updated on NixOS that way too. :)

noffle:
I've been wanting an RSS feed for news as well, so I took a stab at this today! (https://github.com/NixOS/nixos-homepage/issues/90)

domenkozar:
It's live: https://nixos.org/news-rss.xml

PR #90, opened 2016-07-19
Add News RSS feed to website

noffle:
Hi, this is one of my first NixOS PRs. :wave:

This PR addresses https://github.com/NixOS/nixos-homepage/issues/15 and is broken up into 3 logical pieces:

1. Replaces the `<year>/<month>/<day>` tags in `news.xml` with the single tag `<pubDate>`. This format is easy to parse in XSLT 1.0 (no date functions) using only substrings, and happens to also be the expected format for RSS 2.0.

2. Adds an XSLT transform, `news-rss.xsl` that produces the RSS 2.0 feed `news-rss.xml` from `news.xml`.

3. Updates `index.tt` to include
  i. a `<link>` tag for the RSS feed
  ii. an RSS icon + link to the RSS feed beside the "News" heading
  iii. an RSS icon + link to the RSS feed beside the "Blog Posts" heading


domenkozar:
cc @edolstra

domenkozar:
Thanks!

edolstra:
Thanks, merged. However there are two issues:
* The `<description>` elements contain HTML, which is apparently not allowed. So the generated RSS doesn't validate according to https://validator.w3.org/feed/, and the descriptions don't show up in Firefox.
* `blogs.xml` is very big because it contains the complete contents of every blog post, so it may not be very suited as a feed.

noffle:
Thank you, @edolstra. I also appreciate the feedback:

1. You're right: all of the HTML is escaped in e.g. `blogs.xml`. It's a side effect of us using XML for the news data but in inlining HTML /wo escaping we're treating it like XML data. I'll see how hard this is to do in XLST.
2. I'd be fine with capping `blogs.xml` at, say, 20-25 entries.

PR #93, opened 2016-08-07
download.tt: add link to the packages

KaiHa:
Add a link to the distribution specific packages to show up an alternative to piping the curl output to the shell.


domenkozar:
I think link to distribution packages was removed recently due to problems. @edolstra what do we need to fix here to get this going again?

Issue #86, opened 2016-06-27
Install script not working on RHEL anymore


PierreR:
Since dc30713b1 the install script fails on `rhel` with:
```
you do not have 'shasum' installed, which I need to verify the binary tarball
```
On `rhel 7.2`, `shasum` is not available (at least not directly) but there are  `sha512sum` and cie.

domenkozar:
@PierreR so you have `sha256sum`?

domenkozar:
cc @edolstra, any reason not to use `sha512sum`?

edolstra:
`sha512sum` doesn't exist on OS X.

PierreR:
One way to make the script generic would be to check if `sha512sum` exists and resort to `shasum -a 512` if it doesn't.

This is described here:

https://github.com/silpion/ansible-java/pull/15

copumpkin:
Or just call `openssl dgst`, which should be present on almost all systems (Mac OS and all Linuxes I've ever tried)

copumpkin:
@edolstra great, thanks!

PR #94, opened 2016-08-23
Don't require shasum for the Nix installer

aneeshusa:
Recently, the installer script was updated to use sha512sum or openssl
if available, falling back only to shasum if necessary. Since shasum is
no longer needed and there is separate code that checks for each
utility, we no longer need to `require_util` shasum.

Refs #86.


aneeshusa:
I'm guessing the `nix/install.sig` file also needs to be updated after merging.

edolstra:
Thanks, merged!

PR #96, opened 2016-09-21
add Azure blobs to the download page

rbvermaa:


nixos/download.tt
```diff
@@ -111,7 +111,79 @@ $ ec2-run-instances [% ami %] --region [% region %] -k <em>my-key-pair</em>
...
+    --nic-name $VM --vnet-name $VM --vnet-address-prefix 10.0.0.1/16 --vnet-subnet-name $VM \
+    --vnet-subnet-address-prefix 10.0.1.1/24 --public-ip-name $VM --public-ip-domain-name $VM \
+    --admin-username azureuser
+
+# show machine properties and log in
+$ azure vm show $GROUP $VM
+$ ssh azureuser@"$VM"."$LOCATION".cloudapp.azure.com
+</pre>
+
+</section>
```
edolstra: This really bloats the download page, which is not intended to serve as a manual. IMHO, extensive instructions should be moved to the NixOS manual.

mhsjlw:
Not sure where else to put this, but this blob doesn't actually deploy. The VHD isn't rounded to the nearest MB therefore Azure fails to deploy it. It'd be nice if someone could resize it (round up) so that I don't have to do this from scratch :smile:

PR #113, opened 2016-12-03
Add list of Nix consulting companies

zimbatm:
Would it make sense to publicize the consulting companies that offer nix support?

There is a bit of a conflict of interest but it might also help adoption.

/cc @edolstra @domenkozar @jagajaga @qknight @seitz


jagajaga:
Thank you!

qknight:
thank you

PR #78, opened 2016-03-05
WIP: Displaying package parameters

fkz:
See also http://thread.gmane.org/gmane.linux.distributions.nixos/19714
Demo: http://schmitthenner.eu/nixos-homepage/nixos/packages.html

Not yet ready for review but I wanted to start collecting a TODO list at a central place.
# TODO for being able to publish
- [x] call the right patched nix version (instead of the hardcoded ~/... path)
- [x] revert the gunzip (which I need locally because I don't know how pregzipped files work with apache)
- [x] <del> use allow-unsafe-native-code-during-evaluation instead of patching nix (so we don't have to recompile nix) </del> not possible because we added comment (docstring) parsing to the parser of nix now
- [ ] make displaying stable and unstable versions more general
- [x] rebase
# Other nice additions
- [ ] show default value of optional parameters
- [x] investigate the "some error happened" packages
- [x] don't show failed packages (or mark them somehow)
- [ ] show more metadata like comments before parameters (or use more elaborate docstrings etc.)
- [ ] investigate what to do with more elaborate packages, e. g. `vim_configurable` that uses `composableDerivation`
- [x] maybe enable search for parameters
- [ ] make it more beginner friendly: show expression you should insert into `.nixpkgs/config.nix`

Issue #57, opened 2015-11-06
Add possibility to link to packages


matthiasbeyer:
More and more projects link to their packages on `github/nixos/nixpkgs` - it would be nice (in manner of user experience) to link to the package on `nixos.org/nixos/packages`. Currently this is not possible.

A simple solution would be to pass the search term via HTTP GET, I think. I would file the PR for this myself, though I'm by no means a JS guy, so I guess I'll struggle hard to accomplish this.

bjornfor:
+1. Achlinux already has this, ex. this random package: https://aur.archlinux.org/packages/microscheme/. With NixOS I have to manually enter a search term into http://nixos.org/nixos/packages.html. But I want a _link_, so that I can pass it around :-)

davidak:
duplicate of https://github.com/NixOS/nixos-homepage/issues/23

davidak:
You can now get a Permalink: https://nixos.org/nixos/packages.html?attr=pipenv&channel=nixos-19.03&query=pipenv

samueldr:
You can drop the channel name and/or the query from the package name to get a more permanent permalink that links to whatever the default channel is. The important bit is the `attr` argument.

 * https://nixos.org/nixos/packages.html?attr=pipenv

Issue #49, opened 2015-09-14
"Search NixOS packages" doesn't show License


davidak:
checked the first 15 packages, always "License: Not specified" but it is in the Nix expression.

for example:

https://github.com/NixOS/nixpkgs/blob/release-15.09/pkgs/games/0ad/default.nix#L118
https://github.com/NixOS/nixpkgs/blob/release-15.09/pkgs/games/pong3d/default.nix#L20
https://github.com/NixOS/nixpkgs/blob/release-15.09/pkgs/applications/audio/aacgain/default.nix#L42

https://github.com/NixOS/nixos-homepage/blob/master/nixos/packages.tt#L194

source for data `packages.json.gz`:

```
"license":[  
            {  
               "fullName":"GNU General Public License v2.0 only",
               "shortName":"gpl2",
               "spdxId":"GPL-2.0",
               "url":"http://spdx.org/licenses/GPL-2.0"
            },
            {  
               "fullName":"GNU Library General Public License v2.1 only",
               "shortName":"lgpl21",
               "spdxId":"LGPL-2.1",
               "url":"http://spdx.org/licenses/LGPL-2.1"
            },
            {  
               "fullName":"MIT License",
               "shortName":"mit",
               "spdxId":"MIT",
               "url":"http://spdx.org/licenses/MIT"
            },
            {  
               "fullName":"Creative Commons Attribution Share Alike 3.0",
               "shortName":"cc-by-sa-30",
               "spdxId":"CC-BY-SA-3.0",
               "url":"http://spdx.org/licenses/CC-BY-SA-3.0"
            },
            {  
               "fullName":"zlib License",
               "shortName":"zlib",
               "spdxId":"Zlib",
               "url":"http://spdx.org/licenses/Zlib"
            }
         ],
```

output comma seperated list of spdxId.

garbas:
this is now fixed

PR #120, opened 2016-12-20
packages: include unfree and broken packages in the search index

LnL7:
This should help a little bit with https://github.com/NixOS/nixpkgs/issues/17126, hopefully a some more users will find packages because of this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,7 @@ endif
 nixpkgs/packages.json.gz:
 	nixpkgs=$$(nix-instantiate --find-file nixpkgs -I nixpkgs=$(NIXPKGS)); \
 	(echo -n '{ "commit": "' && cat $$nixpkgs/.git-revision && echo -n '","packages":' \
-	  && nix-env -f '<nixpkgs>' -I nixpkgs=$(NIXPKGS) -qa --json --arg config '{}' \
+	  && nix-env -f '<nixpkgs>' -I nixpkgs=$(NIXPKGS) -qa --json --arg config '{ allowUnfree = true; allowBroken = true; }' \
 	  && echo -n '}') \
 	  | sed "s|$$nixpkgs/||g" | gzip -9 > $@.tmp
 	gunzip < $@.tmp | python -mjson.tool > /dev/null


ToxicFrog:
It looks like, due to a separate issue with the JS, such packages won't be clearly marked as "unfree", since license information for most (all?) packages is not displayed. This should be a fairly easy fix (and in fact FIXME #49 already exists for it) to look for `info.meta.license.fullName` if `info.meta.license` is an Object rather than a String.

edolstra:
Sorry, but Nixpkgs is intended as a free software distribution, so it should not promote the availability of unfree packages. It also shouldn't show broken packages unless they are conspicuously marked as broken.

LnL7:
The reason for this change is that somebody tried to package sublime text, and discovered after 3 hours that we already have it.

copumpkin:
@edolstra there's a difference between promoting and actively hiding. All we want is a good user experience, and hiding existing work from users because of a philosophical stance doesn't seem helpful in the long run. It causes wasted work, as @LnL7 mentioned, and also causes people to incorrectly conclude that Nix and nixpkgs is less mature and capable than it really is.

It seems possible to at least openly acknowledge that we have the packages in the system while still not building or distributing them over Hydra and getting people to opt into installing them, right? I'm not looking for magic "oh shit my system is full of unfree software how did that happen???" situations, but there has to be a happy medium between forcing people to clone and grep nixpkgs to find out that we have e.g., Spotify or Steam and installing it on their system by default, right?

ToxicFrog:
@edolstra I'm not convinced that showing them in the index clearly marked as "nonfree" counts as "promoting" them; if you're that concerned, maybe make it an opt-in "show packages with nonfree licenses" tickbox on the page somewhere rather than making it on by default. Even debian, who are fairly hardcore about this, [don't go so far as to completely hide nonfree packages from the index](https://packages.debian.org/search?keywords=steam&searchon=names&suite=stable&section=all). Neither does SUSE, although they relegate nonfree packages to a separate repo.

I do agree that this shouldn't be merged until displaying package licenses is fixed.

ToxicFrog:
Also, even if they are listed in the index and someone doesn't realize they're nonfree, there's still no chance of accidentally installing nonfree software, because as soon as you try it'll error out and tell you to enable `allowUnfree` if you really mean to do this.

jcrben:
@edolstra how would you feel about hinting to users that an unfree package exists, especially if they get zero results? that's a suggestion by @alexfmpe over at https://github.com/NixOS/nixpkgs/issues/17126#issuecomment-369242742

davidak:
I think an option to "Show proprietary packages" like in https://github.com/NixOS/nixos-homepage/pull/78 would be a good solution. Maybe with tooltip to clearify what free software is and why you might not want to use proprietary software.

samueldr:
Hi, I've been watching the thread here for a while.

I have a solution I have been working off and on (mostly off) for that problem, which is a re-implementation of the packages searcher with more features, including regex search, switching channels and showing unfree packages.

Here's a demo: https://nix.samueldr.com/explorer/

I haven't yet made "official moves" to get this on the website since I still need to

 1. Cleanup the code.
 1. Integrate it into the website build.
 1. Misc minor improvements.

The code isn't *bad*, but I want to at least put my eyeballs on every lines once more. [The code isn't hidden away](https://github.com/samueldr/nixos-packages-explorer).

Let's make this comment a commitment to try harder to upstream this effort.

davidak:
@samueldr great!

This still don't fixes the issue i just commented in https://github.com/NixOS/nixos-homepage/issues/57.

https://nix.samueldr.com/explorer/?channel=nixos-unstable&page=1&query=screen&unfree=true

Also, regex is not something a normal user is able to use. Instead it should not be necessary to use it.

samueldr:
Let's say I can subtly insert a way to permalink to a particular attribute :sweat_smile:, as this would be a better experience than using a regex. Using such link could also unfold the result row automatically.

garbas:
I'm closing this PR due being inactive and also since many things changes since this was opened. we will probably continue on this topic here -> https://github.com/NixOS/nixos-homepage/pull/300

PR #123, opened 2017-01-04
Add a Security page for NixOS

grahamc:
I think we can close https://github.com/NixOS/nixpkgs/issues/13515 after merging this. The list of members isn't intended to be final.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ HTML = index.html news.html \
   nix/index.html nix/about.html nix/download.html \
   nixpkgs/index.html nixpkgs/download.html \
   nixos/about.html nixos/download.html nixos/support.html nixos/community.html nixos/packages.html nixos/options.html \
-  nixos/screenshots.html nixos/foundation.html \
+  nixos/screenshots.html nixos/security.html nixos/foundation.html \
   patchelf.html hydra/index.html \
   disnix/index.html disnix/download.html disnix/docs.html \
   disnix/extensions.html disnix/examples.html disnix/support.html \
--- a/layout.tt
+++ b/layout.tt
@@ -76,6 +76,7 @@
                 <li><a href="[%root%]nixos/packages.html">Packages</a></li>
                 <li><a href="[%root%]nixos/options.html">Options</a></li>
                 <li><a href="[%root%]nixos/community.html">Community</a></li>
+                <li><a href="[%root%]nixos/security.html">Security</a></li>
               </ul>
               <ul class="nav pull-right">
                 <!--
--- a/nixos/security.tt
+++ b/nixos/security.tt
@@ -44,6 +44,12 @@
     <li>
       <p>Rob Vermaas
          <a href="mailto:rob.vermaas@gmail.com">rob.vermaas@gmail.com</a>
+         <br />
+         <strong>GPG Key:</strong><tt>
+            <a href="https://pgp.mit.edu/pks/lookup?op=get&amp;search=0xE114A5F264A8AE8E">0xE114A5F264A8AE8E</a>
+         </tt><br />
+         <strong>GPG Fingerprint:</strong>
+         <tt>96BF 75A5 3DEE 1F21 5F0C  979C E114 A5F2 64A8 AE8E</tt>
       </p>
     </li>
   </ul>
--- a/nixos/security.tt
+++ b/nixos/security.tt
@@ -39,6 +39,12 @@
     <li>
       <p>Domen Kožar
          <a href="mailto:domen@dev.si">domen@dev.si</a>
+         <br />
+         <strong>GPG Key:</strong><tt>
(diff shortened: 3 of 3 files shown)


grahamc:
Looks like this:

domenkozar:
My key: https://keybase.io/ielectric

grahamc:
@domenkozar before I do that, can you make a signed commit on github?

grahamc:
BTW @domenkozar, @rbvermaa, @fpletz, we _should not_ sign each other's keys until we meet personally. Just a reminder :) https://www.debian.org/events/keysigning

grahamc:
Updated with a GPG key for Domen, it looks like this:

vcunat:
> we should not sign each other's keys until we meet personally

Actually, having good history of commits under the key seems more trustworthy or proving ownership of the github account with good history (comments etc.).  When meeting someone in person, how do you really verify that it's "who you think it is"?

grahamc:
Yes, I've considered that. ... here are some thoughts:

Unfortunately, Rob nor Domen have good commit history with their key. On this list, only Franz and I do.

I spent a lot of time last night reading about different signing policies. I wanted something like "I know this key does good work and I trust their work, but I can't assert I've verified they're actually Franz Pletz (if that is their real name...!)."

I would like to sign each of these keys to indicate some trust together, but without also having proven identity I don't think that is good GPG etiquette.

I wonder if it is possible to add a UID for "github.com/grahamc" "github.com/domenkozor" "github.com/fpletz" etc. to our keys? Maybe that is something we could sign without in-person identity verification?

dckc:
Have you considered keybase?

e.g. https://keybase.io/dckc

grahamc:
I have ... and I am on Keybase: https://keybase.io/graham I dunno, I suppose I would defer to @fpletz on this one, who I think has the most operational experience with GPG.

fpletz:
@dckc How is trusting Keybase any different than trusting Github, Trump or God to verify my key? You just can't be sure without doing it yourself.

> When meeting someone in person, how do you really verify that it's "who you think it is"?

@vcunat You can prove that you're in possession of the private key by signing a given text and let the other party verify it with the public key they think is yours. In person of course. :smile: 

> I would like to sign each of these keys to indicate some trust together, but without also having proven identity I don't think that is good GPG etiquette.

@grahamc Given that we're dealing with security stuff here and @edolstra previously voiced some concerns about giving pseudonymous contributors commit access, we should absolutely stick to real names at least for the security team for now.

It's also unfortunate that I had to generate a new key a few years ago and haven't managed to attend a keysigning party since. My old key had so many signatures from my Debian times. :cry: 

FWIW and for the record, my real name is indeed Franz Pletz and I am happy to prove that by providing some governmental issued ID in person. Not sure though if that means anything because most of you probably don't know how to check if a German passport or driver's license is fake. :wink: 

> I spent a lot of time last night reading about different signing policies. I wanted something like "I know this key does good work and I trust their work, but I can't assert I've verified they're actually Franz Pletz (if that is their real name...!)."

Strictly speaking, when signing another key (or UID to be exact) you should verify that you trust that this key belongs to the person you think it belongs to (as defined by the UIDs you sign). As we don't know each other in real life this is a bit difficult and unintuitive to achieve. 

So given the Git track record I think it would be safe for you to actually sign for example my key right away because that was our main interaction channel in the past. We could also for instance use IRC as another verification channel if we trust that Freenode and Github couldn't possibly be compromised by the same evil actor. But well, as we haven't met in person we're trusting the broken TLS CA system anyway in the end.

But in any case you should also check that the other party is in control of the email addresses you sign. This is commonly achieved for keysigning parties by sending the signatures to the email addresses of the respective UID after the event instead of uploading the signature to key servers right away. At the party you normally verify that the UID matches some form of official and trustworthy ID and tick a box that you will sign it. There are tools to automate the whole process, like for printing forms to tick off fingerprints or for signing keys and mailing them.

We should do a keysigning party at the next NixCon!

> I wonder if it is possible to add a UID for "github.com/grahamc" "github.com/domenkozor" "github.com/fpletz" etc. to our keys? Maybe that is something we could sign without in-person identity verification?

This would be an option but it's not what the UID field is designed for. Not sure about abusing this.

Another option: One could just sign a short declaration with one's GPG key that says that the signer is in control of a particular Github account and, to be absolutely sure, push that to a repo in your account.

dckc:
> @dckc How is trusting Keybase any different than trusting Github, Trump or God to verify my key?

  - It integrates evidence from multiple sources
  - It's conveniently packaged

> You just can't be sure without doing it yourself.

Even doing it yourself you can be fooled.

Security is about wisely evaluating trade-offs. A large network of nix developers connected via various sorts of evidence (some of which are made more convenient by KeyBase) may provide better risk mitigation than a small network of nix developers who have met in person.

copumpkin:
@fpletz keybase by design doesn't make itself a trusted third party, or makes itself as little of one as possible. You should check out how it works.

PR #129, opened 2017-02-23
Update logo with version that doesn't require specific font installed

leenaars:
The logo had a soft-coded font in it, which meant that people downloading the logo first had to install that font to render the logo correctly. If not the software (browser, office suite, etc) would fall back to some random font on the system - which may not look very nice and certainly is not likely to result in what was intended by the designer. 

The text part of the logo has been converted to paths, thus avoiding all problems and resulting in a pretty logo everywhere.


edolstra:
Thanks, merged!

PR #137, opened 2017-05-05
NixOS option search improvements

zraexy:
Fixes a bunch of issues:
- Every time the query changed by a character a new entry would be put in history.
- The URL would not update after the last character was removed from the search box.
- If you typed in a valid element id e.g. `search` it would try to jump to that element (which misaligns the search box).
- If you typed in` +` and reloaded the page it would be converted into a space.
- If the query had more then two words it would not replace the additional spaces with pluses.

### DIFF
--- a/nixos/options.tt
+++ b/nixos/options.tt
@@ -123,7 +123,16 @@ function refilter() {
       return words.every(match);
     });
 
-    document.location.hash = words.join(' ').replace(/ +/, '+');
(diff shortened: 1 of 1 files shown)


zraexy:
Alternatively we could change behavior so that the name of the option that is currently expanded is copied to the hash.

Mic92:
Any comments from the people with merge permissions?

edolstra:
Thanks @zraexy!

PR #138, opened 2017-05-05
nixos/wiki: init to handle wiki requests

grahamc:
Looks like https://codepen.io/gchristensen/project/full/XxvnoD/

re https://github.com/NixOS/nixos-org-configurations/pull/30

cc @copumpkin

### DIFF
--- a/nixos/wiki.html
+++ b/nixos/wiki.html
@@ -1,0 +1,31 @@
+<h1>The NixOS wiki has been shut down.</h1>
+
+<h2>For official documentation and help, please check:</h2>
+
+<ul>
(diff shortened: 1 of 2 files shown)


nixos/wiki.html
```diff
@@ -0,0 +1,31 @@
+<h1>The NixOS wiki is currently down.</h1>
```
edolstra: "is currently down" implies a temporary issue, like a server problem. That's likely to cause people to constantly ping us about this. Maybe "The NixOS wiki has been shut down.".

grahamc: Good catch. Fixed.

grahamc:
Ready again.

PR #141, opened 2017-05-25
add news entry for NixCon 2017

cko:


vcunat:
Obsoleted by #143, I guess.  (Just a different formulation of basically the same.)

cko:
yep.

PR #146, opened 2017-07-14
Display license information in nix package search

ToxicFrog:
A lot of packages have a license object or array thereof rather than a
license string. With this patch, the package search will display all
listed licences, as links to their SPDX page.

Signed-off-by: Ben Kelly <bk@ancilla.ca>

### DIFF
--- a/nixos/packages.tt
+++ b/nixos/packages.tt
@@ -153,6 +153,45 @@ function refilter() {
   updateTable();
 };
 
+// Try to figure out a human-readable name for a license object.
+// Use the SPDX ID if present; failing that try the long or short names. If all
+// else fails fall back on URL.
+function licenseName(license) {
+  return license.spdxId
+    || license.fullName
+    || license.shortName
+    || license.url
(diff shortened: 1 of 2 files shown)


nixos/packages.tt
```diff
@@ -153,6 +153,18 @@ function refilter() {
...
+// Given a license object, or array of license objects, produce suitable
+// HTML for it.
+function licenseHTML(license) {
+  if (typeof license == 'string') {
+    return license;
+  }
+  if (Array.isArray(license)) {
+    return license.map(licenseHTML).join(', ');
+  }
+  return '<a href="' + license.url + '">' + license.spdxId + '</a>';
```
edolstra: It's probably a good idea to handle the case where `url` or `spdxId` are missing, and to escape them.

ToxicFrog:
Updated. Should now properly handle licenses containing html-special characters, and licenses with a name but no URL (and vice versa).

edolstra:
Thanks!

PR #148, opened 2017-07-16
Verified install steps

grahamc:
First commit improves the places a shell command is on the website, where the `$` of the prompt is placed with CSS. This means that when you copy-paste the line it is impossible to accidentally copy the `$` of the shell.

Second commit creates a `.tt` file containing the sha sums of the nix/install file, making it available in templates.

Third commit updates the "quick" install popup thing to include a sha verification, and a signature verification. I think it is valuable to put these verification steps right up front in order to:

1. demonstrate we care
2. demonstrate they're available, without having to click to the "additional options" page

however doesn't put GPG in the mandatory path for users unless they would like to use it. Additionally:

1. this PR adds a second link to the GPG key, on a secondary domain if someone truly paranoid would like to validate the signing key matches in both places, and
2. adds the `gpg2 --recv-keys` line right there, to make it trivially easy to follow.

Note the light trickery in the sha256 validation line:

> echo "aa37b53577c96ad745c0d84a871de3e398a1cd9c32987d0e662d92c56cdcc4ef  install-nix-1.11.12" | (shasum -p 256 || sha256sum -c) 2> /dev/null

If shasum doesn't exist, it'll fall back to sha256sum automatically.

 All in all, the changes end up looking like this:

screenshot-2017-7-16 nix_ the purely functional package manager

This work is inspired by trying to make my work's security team happy enough with the instructions.

Here is the output of a trial run of these quick instructions:

```
grahamc@Morbo> curl -o install-nix-1.11.12 https://nixos.org/nix/install                                                                % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  2358  100  2358    0     0   5937      0 --:--:-- --:--:-- --:--:--  5954

grahamc@Morbo> echo "aa37b53577c96ad745c0d84a871de3e398a1cd9c32987d0e662d92c56cdcc4ef  install-nix-1.11.12" | (shasum -p 256 || sha256sum -c) 2> /dev/null
install-nix-1.11.12: OK

grahamc@Morbo> gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE                                                              gpg: key 0x8170B4726D7198DE: "Eelco Dolstra <edolstra@gmail.com>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1

grahamc@Morbo> curl -o install-nix-1.11.12.sig https://nixos.org/nix/install.sig                                                        % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   488  100   488    0     0   1349      0 --:--:-- --:--:-- --:--:--  1359

grahamc@Morbo> gpg2 --verify ./install-nix-1.11.12.sig                                                                                gpg: assuming signed data in './install-nix-1.11.12'
gpg: Signature made Thu 13 Jul 2017 09:44:48 AM EDT
gpg:                using RSA key B541D55301270E0BCF15CA5D8170B4726D7198DE
gpg: Good signature from "Eelco Dolstra <edolstra@gmail.com>" [unknown]
gpg:                 aka "Eelco Dolstra <eelco.dolstra@logicblox.com>" [unknown]
gpg: WARNING: This key is not certified with a trusted signature!
gpg:          There is no indication that the signature belongs to the owner.
Primary key fingerprint: B541 D553 0127 0E0B CF15  CA5D 8170 B472 6D71 98DE
```

### DIFF
--- a/nix/about.tt
+++ b/nix/about.tt
@@ -188,7 +188,7 @@ href="https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/
 Nix expression</a>:</p>
 
 <pre>
-<span class="command">nix-shell '&lt;nixpkgs>' -A pan</span>
+<span class="command">nix-shell '&lt;nixpkgs&gt;' -A pan</span>
 </pre>
 
 <p>You’re then dropped into a shell where you can edit, build and test
--- a/nix/about.tt
+++ b/nix/about.tt
@@ -88,8 +88,8 @@ there after an upgrade.  This means that you can <em>roll back</em> to
 the old version:</p>
 
 <pre class="code">
-<span class="muted">$</span> nix-env --upgrade <em>some-packages</em>
-<span class="muted">$</span> nix-env --rollback
+<span class="command">nix-env --upgrade <em>some-packages</em></span>
+<span class="command">nix-env --rollback</span>
 </pre>
 
 
@@ -98,7 +98,7 @@ the old version:</p>
 <p>When you uninstall a package like this…</p>
 
 <pre class="code">
-<span class="muted">$</span> nix-env --uninstall firefox
+<span class="command">nix-env --uninstall firefox</span>
 </pre>
 
 <p>the package isn’t deleted from the system right away (after all,
@@ -107,7 +107,7 @@ other users).  Instead, unused packages can be deleted safely by
 running the <em>garbage collector</em>:</p>
 
 <pre class="code">
-<span class="muted">$</span> nix-collect-garbage
+<span class="command">nix-collect-garbage</span>
 </pre>
 
 <p>This deletes all packages that aren’t in use by any user profile or
@@ -137,7 +137,7 @@ store.</p>
 source, so an installation action like</p>
 
 <pre class="code">
-<span class="muted">$</span> nix-env --install firefox
+<span class="command">nix-env --install firefox</span>
 </pre>
 
 <p><em>could</em> cause quite a bit of build activity, as not only
@@ -188,18 +188,18 @@ href="https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/
 Nix expression</a>:</p>
 
 <pre>
-<span class="muted">$</span> nix-shell '&lt;nixpkgs>' -A pan
+<span class="command">nix-shell '&lt;nixpkgs>' -A pan</span>
 </pre>
 
 <p>You’re then dropped into a shell where you can edit, build and test
 the package:</p>
 
 <pre>
-<span class="muted">[nix-shell]$</span> tar xf $src
-<span class="muted">[nix-shell]$</span> cd pan-*
-<span class="muted">[nix-shell]$</span> ./configure
-<span class="muted">[nix-shell]$</span> make
-<span class="muted">[nix-shell]$</span> ./pan/gui/pan
+<span class="nix-shell-command">tar xf $src</span>
+<span class="nix-shell-command">cd pan-*</span>
+<span class="nix-shell-command">./configure</span>
+<span class="nix-shell-command">make</span>
+<span class="nix-shell-command">./pan/gui/pan</span>
 </pre>
 
 <p>Since Nix packages are reproducible and have complete dependency
--- a/nix/download.tt
+++ b/nix/download.tt
@@ -7,17 +7,17 @@
 on Linux and Mac OS X is to run the following in a shell (as a user
 other than <tt>root</tt>):
 
-<pre><span class="muted">$</span> curl <a href="[%root%]nix/install">https://nixos.org/nix/install</a> | sh</pre>
+<pre><span class="command">curl <a href="[%root%]nix/install">https://nixos.org/nix/install</a> | sh</span></pre>
 
 Make sure to follow the instructions output by the script.
 
 You may want to verify the integrity of the installation script using GPG:
 
 <pre>
-<span class="muted">$</span> curl -O <a href="[%root%]nix/install">https://nixos.org/nix/install</a>
-<span class="muted">$</span> curl -O <a href="[%root%]nix/install.sig">https://nixos.org/nix/install.sig</a>
-<span class="muted">$</span> gpg2 --verify ./install.sig
-<span class="muted">$</span> sh ./install
+<span class="command">curl -O <a href="[%root%]nix/install">https://nixos.org/nix/install</a></span>
+<span class="command">curl -O <a href="[%root%]nix/install.sig">https://nixos.org/nix/install.sig</a></span>
+<span class="command">gpg2 --verify ./install.sig</span>
+<span class="command">sh ./install</span>
 </pre>
 
 The <a href="[%root%]edolstra.gpg">signing key</a> has fingerprint <tt>B541 D553 0127 0E0B CF15  CA5D 8170 B472 6D71 98DE</tt>.</p>
--- a/nix/download.tt
+++ b/nix/download.tt
@@ -4,31 +4,36 @@
 
 <p>The latest and recommended version of Nix is
 <strong>[%latestNixVersion%]</strong>. The quickest way to install it
-on Linux and Mac OS X is to run the following in a shell (as a user
-other than <tt>root</tt>):
-
-<pre><span class="command">curl <a href="[%root%]nix/install">https://nixos.org/nix/install</a> | sh</span></pre>
-
-Make sure to follow the instructions output by the script.
-
-You may want to verify the integrity of the installation script using GPG:
+on Linux and macOS is to run the following in a shell (as a user
+other than <tt>root</tt>):</p>
 
 <pre>
-<span class="command">curl -O <a href="[%root%]nix/install">https://nixos.org/nix/install</a></span>
-<span class="command">curl -O <a href="[%root%]nix/install.sig">https://nixos.org/nix/install.sig</a></span>
-<span class="command">gpg2 --verify ./install.sig</span>
-<span class="command">sh ./install</span>
+<span class="command">curl https://nixos.org/nix/install | sh</span>
 </pre>
 
-The <a href="[%root%]edolstra.gpg">signing key</a> has fingerprint <tt>B541 D553 0127 0E0B CF15  CA5D 8170 B472 6D71 98DE</tt>.</p>
+<p>Make sure to follow the instructions output by the script.</p>
+
+<h3>GPG Verification</h3>
+<p>You may want to verify the integrity of the installation script
+using GPG:</p>
+
+<pre>
+<span class="command">curl -o install-nix-[%latestNixVersion%] https://nixos.org/nix/install</span>
+<span class="command">curl -o install-nix-[%latestNixVersion%].sig https://nixos.org/nix/install.sig</span>
+<span class="command">gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE</span>
+<span class="command">gpg2 --verify ./install-nix-[%latestNixVersion%].sig</span>
+<span class="command">sh ./install-nix-[%latestNixVersion%]</span>
+</pre>
+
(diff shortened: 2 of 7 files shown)


grahamc:
# This is a work in progress, please don't merge yet. I would love feedback though.

nixos/packages.tt
```diff
@@ -33,7 +33,7 @@
     <tr>
       <th>Install command:</th>
       <td>
-        <tt><span class="muted">$</span> nix-env -iA nixos.<span class='attrname'></span></tt> <em class='muted'>(NixOS channel)</em>
+        <tt><span class="command">nix-env -iA nixos.<span class='attrname'></span></span></tt> <em class='muted'>(NixOS channel)</em>
```
newfurniturey: Small style nit, even though double-quotes are the general "style of the web", _this_ specific file seems to use single-quotes for all HTML tag attributes. For consistence, could you make this:

```html
- <span class="command">
+ <span class='command'>
```

... that, or make a separate PR to update the rest of this file to use double quotes (which is my _personal_ recommendation) =P

Makefile
```diff
@@ -119,18 +119,25 @@ docs/papers-in.html: docs/papers.xml docs/bib2html.xsl
...
+	  --pre_process=nix-release.tt --pre_process=nix/install-hash.tt --pre_process=nixos-release.tt --pre_process=common.tt \
 	  $< > $@.tmp
 	xmllint --nonet --noout $@.tmp
 	mv $@.tmp $@
 
+nix/install-hash.tt: nix/install
+	echo "[%-" > $@
+	echo " nixInstallHashsha256 = \"$$(sha256sum nix/install | cut -d' ' -f1)\";" >> $@
+	echo " nixInstallHashsha1 = \"$$(sha1sum nix/install | cut -d' ' -f1)\";" >> $@
+	echo " nixInstallHashmd5 = \"$$(md5sum nix/install | cut -d' ' -f1)\";" >> $@
```
newfurniturey: As the commands presented only use `shasum` and `sha256sum`, why provide an MD5 hash here too? The more the merrier of course, but it feels like it's just additional noise given the multiple more-modern choices available.

nix/about.tt
```diff
@@ -188,18 +188,18 @@ href="https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/networking/
 Nix expression</a>:</p>
 
 <pre>
-<span class="muted">$</span> nix-shell '&lt;nixpkgs>' -A pan
+<span class="command">nix-shell '&lt;nixpkgs>' -A pan</span>
```
newfurniturey: Since this appropriately uses `&lt;` for `<`, can it be updated to _also_ use `&gt;` for `>` (for consistency)?

```html
- <span class="command">nix-shell '&lt;nixpkgs>' -A pan</span>
+ <span class="command">nix-shell '&lt;nixpkgs&gt;' -A pan</span>
```

nix/download.tt
```diff
@@ -4,23 +4,30 @@
...
 <p>The latest and recommended version of Nix is
 <strong>[%latestNixVersion%]</strong>. The quickest way to install it
-on Linux and Mac OS X is to run the following in a shell (as a user
+on Linux and macOS is to run the following in a shell (as a user
 other than <tt>root</tt>):
 
-<pre><span class="muted">$</span> curl <a href="[%root%]nix/install">https://nixos.org/nix/install</a> | sh</pre>
+<pre>
+<span class="command">curl -o install-nix-[%latestNixVersion%] https://nixos.org/nix/install</span>
+<span class="command">echo "[%nixInstallHashsha256%]  install-nix-[%latestNixVersion%]" | (shasum -p 256 || sha256sum -c) 2> /dev/null</span>
```
newfurniturey: Insanely small nit-pick, but there's an extra space between these two:

```
"[%nixInstallHashsha256%]  install-nix-[%latestNixVersion%]"
```

=P

Makefile
grahamc: I only actually used the sha256sum, the other two were remnants from an experiment. Cleaned up.

nix/download.tt
```diff
@@ -4,23 +4,30 @@
...
+<span class="command">sh install-nix-[%latestNixVersion%]</span>
+</pre>
```
newfurniturey: Will there be a case when `sha256sum` is available but `shasum` is not (or vice-verse)? Or, would there be a package difference that would cause `shasum -p 256` to generate a _different_ hash from `sha256sum -c`?

I ask because if not for both, which I think would be the case, we could actually lessen this to only include `sha256sum -c` for a shorter line overall (or just `shasum -p 256` if we're worried that `sha256sum` isn't avialable). That, or if precedence is for `sha256sum`, perhaps flip the order to `(sha256sum -c || shasum -p 256)`.

nix/download.tt
grahamc: Indeed! mandatory formatting: without the 2 spaces, shasum complains and won't validate the file.

nix/download.tt
grahamc: Tragically, macOS only has shasum out of the box, where as most linuxes have sha256sum, and some don't have shasum. For super bonus points, some don't have either and you have to use openssl >.> I'm omitting support for thaht, because I'm going to assume they can figure out how to check a hash (or not.)

nix/download.tt
newfurniturey: Whoah; I never realized! _brilliant!_

grahamc:
Thank you for the review, @newfurniturey! I've addressed your feedback (including the ' -> " situation.)

nix/download.tt
```diff
@@ -4,23 +4,30 @@
...
 
 <pre>
-<span class="muted">$</span> curl -O <a href="[%root%]nix/install">https://nixos.org/nix/install</a>
-<span class="muted">$</span> curl -O <a href="[%root%]nix/install.sig">https://nixos.org/nix/install.sig</a>
-<span class="muted">$</span> gpg2 --verify ./install.sig
-<span class="muted">$</span> sh ./install
+<span class="command">curl -o install-nix-[%latestNixVersion%] https://nixos.org/nix/install</span>
+<span class="command">curl -o install-nix-[%latestNixVersion%].sig https://nixos.org/nix/install.sig</span>
+<span class="command">gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE</span>
+<span class="command">gpg2 --verify ./install-nix-[%latestNixVersion%].sig</span>
```
newfurniturey: Not to expand the install steps much, but with this line it's important that the two curl commands above are executed as-is (specifically, the two output files are named as they're indicated.

`gpg2 --verify`, if only given the `.sig` file, looks for a file with the _same name_, less the `.sig` to be the "signed file".

Though `man gpg2` will tell anyone that, it may be an unexpected error and "fun" install process if the end-user is unaware and decides to curl into a different filename for either step...

nix/index.tt
```diff
@@ -21,16 +21,31 @@
...
 
+    <p>If you would like to verify the installer with GPG, import the
+    <a href="[%root%]/edolstra.gpg">signing key</a> (also on
+    <a href="https://github.com/NixOS/nixos-homepage/blob/master/edolstra.gpg">GitHub</a>)
+    and verify with the signature:</a>
+
+    <pre>
+<span class="command">gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE</span>
+<span class="command">curl -o install-nix-[%latestNixVersion%].sig https://nixos.org/nix/install.sig</span>
+<span class="command">gpg2 --verify ./install-nix-[%latestNixVersion%].sig</span>
```
newfurniturey: To echo the comment above, this command requires/assumes there is a file signed with this signature by the name of `install-nix-[%latestNixVerison%]` (e.g. "same filename, without the `.sig`"). Not sure if it's necessary to note, based on the population that would verify via gpg - but something to keep in mind =]

nix/download.tt
grahamc: Hmm... I think how to use gpg is out of scope of this. Do you have any specific suggestions on how to handle the problem you've described? I guess what I'm hoping is if someone doesn't know how to use gpg, they won't muck about with the provided commands 😬.

grahamc:
OK I think this is ready to go. When I try the packages.html locally I get a data loading error, which I suspect is because I'm running it from `file://` and not on a proper webserver.

edolstra:
I kinda feel that if we're going to put a lot of instructions under the "Get Nix" popup, we might as well make "Get Nix" a direct link to the download page, and provide detailed steps there. The popup was intended to provide a one-liner installation command. However, I'd prefer to keep the simple `curl | bash` under the popup and put detailed instructions on the download page.

grahamc:
Okay, that is alright -- I'll update this PR to restore the simple popup, and it more clear there are verification instructions.

nix/about.tt
```diff
@@ -88,8 +88,8 @@ there after an upgrade.  This means that you can <em>roll back</em> to
 the old version:</p>
```
grahamc: This diff is nothing but:
1. `<span class="muted">$</span> my command` to `<span class="command">my command</span>`
2. updated a `>` to `&gt;`

nixops/index.tt
```diff
@@ -43,8 +43,8 @@ should be instantiated (in this case, as VirtualBox virtual machines
 running on your desktop). You can then run:</p>
```
grahamc: This diff is nothing but updating `<span class="muted">$</span> my command` to `<span class="command">my command</span>`

nixos/community.tt
```diff
@@ -9,14 +9,14 @@
     well):</p>
```
grahamc: This diff is nothing but updating `<span class="muted">$</span> my command` to `<span class="command">my command</span>`

nixos/packages.tt
```diff
@@ -1,18 +1,18 @@
-[% WRAPPER layout.tt title='Search NixOS packages' menu='nixos' %]
```
grahamc: This diff is mostly updating single quotes to be more internet-standard double quotes, plus  `<span class="muted">$</span> my command` to `<span class="command">my command</span>` or two.

nix/download.tt
```diff
@@ -64,10 +67,7 @@ href="../hydra">continuous build system</a>:</p>
 <ul>
   <li><a
   href="https://hydra.nixos.org/job/nix/trunk/release#tabs-constituents">The
-  latest development release of Nix</a> (<a
-  href="https://hydra.nixos.org/job/nix/trunk/release/latest-finished/eval/job/tarball/download-by-type/file/source-dist">source
-  tarball</a> and <a
-  href="https://hydra.nixos.org/job/nix/trunk/release/latest-finished/eval/job/tarball/download-by-type/doc/manual">manual</a>)</li>
+  latest development release of Nix</a></li>
```
grahamc: Removed because Hydra doesn't serve these anymore.

nix/download.tt
```diff
@@ -4,31 +4,36 @@
...
 </pre>
 
-The <a href="[%root%]edolstra.gpg">signing key</a> has fingerprint <tt>B541 D553 0127 0E0B CF15  CA5D 8170 B472 6D71 98DE</tt>.</p>
+<p>The <a href="[%root%]edolstra.gpg">signing key</a> has fingerprint <tt>B541 D553 0127 0E0B CF15  CA5D 8170 B472 6D71 98DE</tt>.
+It is also available <a href="https://github.com/NixOS/nixos-homepage/blob/master/edolstra.gpg">on GitHub</a>.</p>
 
 <p>The installation script requires that you have <tt>sudo</tt> access
-to <tt>root</tt>, unless the directory <tt>/nix</tt> already exists
-and is writable by you. You can uninstall Nix simply by running <tt>rm
+to <tt>root</tt>. You can uninstall Nix simply by running <tt>rm
```
grahamc: The part about `/nix` existing / owned by you isn't true when on macOS, so I just deleted it for simplicity.

edolstra:
Thanks, merged!

grahamc:
Way cool! thank you a lot! :D

PR #149, opened 2017-07-16
Remove screenshots

grahamc:
Most of the screenshots are of really out of date software ... and vista? :) Anyway, we don't put a ton of work in customized branding, so the screenshots don't do much to show what NixOS is all about.


grahamc:
thank you!

PR #153, opened 2017-08-17
Nix pills

grahamc:
1.  Correct root path in manuals

Previous behavior added an extra `../` to the path, which works on web
(extra ../'s get removed once it gets to the root) but broke locally.

2.  When there are navigation links in docbook, present the top and bottom links

3. Distribute the Nix Pills on the website

The about 60% of the nix pills have been ported to docbook. You could merge this now, or wait until they're fully ported. Also, of course, if you wish I did this differently, that'd be fine too :)

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -45,8 +45,6 @@
                 Nix
               [% ELSIF menu == 'nixpkgs' %]
                 Nixpkgs
-              [% ELSIF menu == 'nix-pills' %]
-                Nix Pills
               [% ELSIF menu == 'nixops' %]
                 NixOps
               [% ELSIF menu == 'disnix' %]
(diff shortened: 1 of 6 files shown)


grahamc:
@edolstra the Pills have all been ported. This is fully Ready To Go!

grahamc:
Please hold, actually. Verifying some changes to the makefile.

grahamc:
Ok, the patches I've made fix an issue where realpath failed if the website hadn't been generated before. Now it works OK and  should be 🆗

layout.tt
```diff
@@ -59,6 +61,7 @@
             <ul class="dropdown-menu">
               <li><a href="[%root%]">NixOS</a></li>
               <li><a href="[%root%]nix">Nix</a></li>
+              <li><a href="[%root%]nix-pills">Nix Pills</a></li>
```
edolstra: It seems strange to put this in the drop down menu as a top-level project. Shouldn't it just be linked from the docs page?

grahamc: It could be, I wasn't sure which top level project it would belong to for the menu, and was worried about it being confusing to click docs in nix/nixpkgs/nixos and end up in a different section. Then I put it in its own section. What do you recommend?

grahamc:
I moved the pills to nixos/support

PR #156, opened 2017-08-22
On chapter pages, use the value of the 'Home' link as the title

grahamc:
This fixes https://nixos.org/nixos/nix-pills/why-you-should-give-it-a-try.html's header being "NixOS Manual"


grahamc:
Note that because this is the only multi-page doc on the site, it could have been hard-coded, but also I'm not aware of any other doc this impacts.

PR #157, opened 2017-08-25
Note entry point to top level packages.

Fresheyeball:


garbas:
this is covered in nixpkgs manual since it is quite an internal thing. we might consider documenting this in nixpkgs manual better.

PR #160, opened 2017-09-05
Build the options page in Elm

grahamc:
Just a last minute PR before vacation. Incorporates Elm in to the build phase, including pinned elm dependencies. PR version of http://gsc.io/nixos/nixos/options.html

I'll be back in a week if you have questions :)


grahamc:
Of course keep in mind I built this to learn elm, not specifically with the
intent of it merging and being used for NixOS. If you like it, I'll address
feedback and make it mergable. If not, it won't hurt my feelings and you
can close the PR :)

zraexy:
Looks good. The only issue I see is that it introduces a couple of regressions:

- Every time a character is typed in the search box a new entry is pushed to history.
- The back/forward buttons don't update the page.

It might also be good to always keep the query in the URL even when an option is selected so that if the page is refreshed the query is not lost

grahamc:
> The back/forward buttons don't update the page. 

Good catch!

> Every time a character is typed in the search box a new entry is pushed to history.

Not sure how to fix this with Elm, but I'll research when I get back.

> It might also be good to always keep the query in the URL even when an option is selected so that if the page is refreshed the query is not lost

This was a tough choice, but I think I settled on the right implementation: what if the query no longer matches the selected option when they come to it? Same for page number: what if it is on a different page? Once selected, IMO the query is not significant, just the selected option. Also, related options show up for free due to the alpha sorting.

grahamc:
@zraexy can you test again? gsc.io/nixos/nixos/options.html code needs cleaning up, but should work ok now w.r.t. back buttons and history pollution.

grahamc:
I've found an issue when your query has no results. Do you see other issues?

elm-src/.gitignore
```diff
@@ -0,0 +1,3 @@
+elm-stuff/*
+!elm-stuff/exact-dependencies.json
```
zwilias: This is not recommended. Couple of things to keep in mind:
- this file is used to keep track of the entire dependency tree including transitive dependencies
- elm does semver (and does it well)
- if you require version pinning, the recommended approach is by selecting a specific version in `elm-package.json` by restricting the range (i.e. `1.1.1 <= v < 1.1.2` to pin to 1.1.1)

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+
+debounce =
+    Debounce.debounceCmd debounce_cfg
+
+
+debounce_cfg : Debounce.Config Model Msg
+debounce_cfg =
+    Debounce.config
+        .debounce_state
+        -- getState   : Model -> Debounce.State
```
zwilias: I think putting the comments on the preceding line will clarify things, and make the "trailing" comment thing go away

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+            [ onClick (ChangePage page) ]
+         else
+            [ disabled True ]
+        )
+        [ text txt ]
+
+
+view : Model -> Html Msg
+view model =
+    if List.length model.options > 0 then
```
zwilias: of `options` could get very long, `if not <| List.isEmpty model.options then` may be clearer/safer. `length` is `O(n)`

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+            [ thead []
+                [ tr []
+                    [ th [] [ text "Option name" ]
+                    ]
+                ]
+            , tbody []
+                (List.concat (List.map (optionToTd model.selected) (fetch_page model.page model.matchingOptions)))
+            ]
+        , ul [ class "pager" ]
+            [ li []
```
zwilias: You could potentially format these a bit different (i.e. `li [] [ (changePageIf (not (model.page <= 1)) 1 "« First") ]` could go on a single line, pretty sure `elm-format` would allow it, too :)

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+            [ input [ id "search", class "search-query span3", autofocus True, value model.query, onInput ChangeQuery ] []
+            ]
+        , hr [] []
+        , p [ id "how-many" ]
+            [ em []
+                [ text
+                    (if (List.length model.matchingOptions) == 0 then
+                        "Showing no results"
+                     else
+                        (String.concat
```
zwilias: Specific reason you seem to (generally) prefer `List.concat` and `String.concat` over the `++` concatenation operator? `++` feels little more idiomatic, to me.

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+            toString x
+
+        GFloat x ->
+            toString x
+
+        GList list ->
+            nix_list list
+
+        GDict dict ->
+            if ((Dict.member "_type" dict) && ((Dict.get "_type" dict) == Just (GString "literalExample")) && (Dict.member "text" dict)) then
```
prikhi: You could turn this from an `if/then` to a `case/of`:

```
case (Dict.get "_type" dict, Dict.get "text" dict) of
  (Just (GString "literalExample"), Just (GString txt) ->
    txt
  (Just (GString "literalExample"), _ ) -> 
    "Literal example failed to product text"
  _ ->
     elseBranchCodeHere
```

elm-src/NixOptions.elm
```diff
@@ -0,0 +1,50 @@
...
+
+filtered : NixOptions -> List String -> NixOptions
+filtered options terms =
+    List.filter (option_filter terms) options
+
+optionDecoder : Decode.Decoder Option
+optionDecoder =
+    decode Option
+        |> required "declarations" (Decode.list Decode.string)
+        |> optional "default" (Decode.lazy (\_ -> gsonDecoder)) gnull
```
zwilias: no need for `lazy` here

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+                        txt
+
+                    otherwise ->
+                        "Literal example failed to produce text"
+            else
+                let
+                    entries =
+                        Dict.toList dict
+                in
+                    if List.length entries == 0 then
```
prikhi: `if List.isEmpty entries then`

elm-src/NixOptions.elm
```diff
@@ -0,0 +1,50 @@
...
+        |> required "description" Decode.string
+        |> optional "example" (Decode.lazy (\_ -> gsonDecoder)) gnull
```
zwilias: no need for `lazy` here

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+                        "{ }"
+                    else
+                        String.concat
+                            (List.concat
+                                [ [ "{\n" ]
+                                , indent
+                                    (List.concat
+                                        (List.map nix_enc_dict entries)
```
prikhi: `List.concat ( List.map ...` can be replaced with `List.concatMap`

elm-src/NixOptions.elm
```diff
@@ -0,0 +1,50 @@
...
+    List ( String, Option )
+
+option_sort_key : ( String, Option ) -> String
+option_sort_key ( name, option ) =
+    name
+
+option_filter : List String -> ( String, Option ) -> Bool
+option_filter terms ( name, option ) =
+    List.any (term_matches name) terms
+        || List.any (term_matches option.description) terms
```
zwilias: `List.any (\term -> term_matches name term || term_matches option.description term) terms`

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+                , [ "''" ]
+                ]
+            )
+    else
+        String.concat [ "\"", x, "\"" ]
+
+
+nix_list : List Gson -> String
+nix_list list =
+    if List.length list == 0 then
```
prikhi: `List.isEmpty`

elm-src/NixJSON.elm
zwilias: it's a little weird to have `nix_list` as a separate function, yet inline this one. It looks like it could easily live in a separate function, and use a little refactoring :)

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+            (List.concat
+                [ [ "''\n" ]
+                , [ x ]
+                    |> segments_to_lines
+                    |> indent
+                , [ "''" ]
+                ]
+            )
+    else
+        String.concat [ "\"", x, "\"" ]
```
zwilias: I really do think `"\"" ++ x ++ "\""` reads easier, but perhaps that's just personal preference

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+
+
+nix_list : List Gson -> String
+nix_list list =
```
zwilias: How about something along these lines?:

```elm
nix_list : List Gson -> String
nix_list list =
  let
    present : List String -> String
    present strings =
      case strings of
        [] -> ""
        [ x ] -> x
        _ -> "\n" ++ (String.join "\n  " strings) ++ \"n"
  in
  "[" ++ (List.map nixFromGson list |> present) ++ "]"
```

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+import NixOptions exposing (..)
+import NixJSON exposing (Gson, gnull, gsonDecoder, nixFromGson)
+import Paginate exposing (..)
+import Html exposing (node, Attribute, Html, a, text, em, table, tr, th, td, tbody, thead, input, ul, li, button, h1, h2, div, pre, p, hr)
+import Html.Attributes exposing (placeholder, src, disabled, href, class, autofocus, value, id)
+import Html.Events exposing (onInput, onClick)
+import Http
+import Json.Encode exposing (encode, Value, null)
+import Dict exposing (Dict, empty, foldr, filter)
+import List
```
zwilias: No need for the default imports (as listed here: http://package.elm-lang.org/packages/elm-lang/core/latest)

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+
+clamped : Model -> Model
+clamped model =
+    { model | page = (clamp 1 (total_pages model.matchingOptions) model.page) }
+
+
+update_options : Model -> NixOptions -> Model
+update_options model unsorted_options =
+    let
+        options =
```
zwilias: General-ish remark: type signatures everywhere prevents a lot of mistakes down the line; even in `let..in` expressions :)

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+
+        query =
+            Maybe.withDefault "" (QueryString.one QueryString.string "query" qs)
+
+        page =
+            Maybe.withDefault 1 (QueryString.one QueryString.int "page" qs)
+
+        selected =
+            QueryString.one QueryString.string "selected" qs
+    in
```
zwilias: Seems like there's some shared code between `updateFromUrl` and `init`..

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+
+update : Msg -> Model -> ( Model, Cmd Msg )
+update msg model =
+    case msg of
+        FetchedOptions (Ok options) ->
+            let
+                newModel =
+                    update_options model options
+            in
+                ( newModel, updateUrl newModel )
```
zwilias: Perhaps an alternative is to rewrite `updateUrl` so it's type-signature can be `updateUrl : Model -> (Model, Cmd Msg)`. If you do that, then these branches can be written as `model |> update_options options |> updateUrl`. (you'd have to flip the params for `updateOptions`, but in general putting the "core" datastrcture a function works on as the last argument is good form, since it's pipeline-friendly)

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,150 @@
...
+                                , indent
+                                    (List.concat
+                                        (List.map nix_enc_dict entries)
+                                    )
+                                , [ "}" ]
+                                ]
+                            )
+
+
+nix_string : String -> String
```
zwilias: It took me a while to realize this, but what's with the `snake_case` functions? This feels very weird. Some sort of disambiguation between exposed and non-exposed functions? I'd say that doesn't really make sense - whether a function is exposed or not doesn't make a difference to _how_ you'd use it within the defining module.

I feel like I'm missing something?

elm-src/index.elm
```diff
@@ -0,0 +1,489 @@
...
+    Sub.none
+
+
+
+-- HTTP
+
+
+getOptions : Cmd Msg
+getOptions =
+    Http.send FetchedOptions (Http.get "./options.json" decodeOptions)
```
zwilias: In this case (and quite a few others), I personally like using `<|` to remove some parents:

```
Http.send FetchedOptions (Http.get "./options.json" decodeOptions)
Http.send FetchedOptions <| Http.get "./options.json" decodeOptions
```

zwilias: Or, perhaps, even more sensical

```
Http.get "./options.json" decodeOptions
    |> Http.send FetchedOptions
``` 

i.e. build the request, then send it

zraexy:
Although the debouncing does help some, a longer delay is needed in order to be sure that typing has finished. Currently, any hesitation whatsoever causes a new entry to be pushed into history; I think it should at least wait for 1/3 of a second. Unfortunately, the delay cannot be increased too much as it would feel sluggish. I'm honestly not sure that it's even possible to strike a good balance with the current setup.

I noticed something else: While the old search requires that all terms are found in order for the result to be included, this only requires that one is. Is that intentional?

elm-src/NixOptions.elm
grahamc: I used `lazy` in both these cases because they're a bit expensive to decode, there are thousands of them, and a user may only ever need to look at 15 of them. Is this a sensible use of lazy?

elm-src/NixJSON.elm
grahamc: Nice! Thank you!

elm-src/.gitignore
grahamc: So the reason I do this, is so Nix can do a pure build of the application. I'm parsing this file to pre-fetch dependencies at build time. See https://github.com/NixOS/nixos-homepage/pull/160/files#diff-42fbbaa4d2f9ce0b8c95980d6d43cc96R5 to about line 55.

elm-src/NixOptions.elm
zwilias: No, the decoded result is always fully evaluated. `lazy` applies to constructing the decoder, but makes no difference to actually executing the decoder.

elm-src/.gitignore
zwilias: Right. Seems weird to bypass elm-package that way, but I don't know how Nix builds work or what the constraints are. I guess what I'm asking is this: what's the added value of this over `elm-package install --yes`?

elm-src/NixOptions.elm
```diff
@@ -22,8 +22,8 @@ option_sort_key ( name, option ) =
 
 option_filter : List String -> ( String, Option ) -> Bool
 option_filter terms ( name, option ) =
-    List.any (term_matches name) terms
-        || List.any (term_matches option.description) terms
+    List.all (term_matches name) terms
+        || List.all (term_matches option.description) terms
```
zwilias: Hm, so now (all terms appear in name) OR (all terms appear in description). Is that the intended semantic?

elm-src/index.elm
grahamc: Good catch. Thanks!

elm-src/.gitignore
grahamc: A build under Nix isn't allowed to access the internet unless it can declare ahead of time the contents of what it is going to make.

In other words, there are:

 -  "fixed output" builds (we call them "derivations") where it downloads something from the internet. To have a fixed output derivation, it has to start out by saying what the sha256 of the result is going to be. If what it creates doesn't match the sha256, the build is failed and the result is deleted. In this case, before starting the inputs are unknown (the network!) so the output must be known.
 - standard builds where there is no network access, but the output is not limited. In this case all the inputs are known ahead of time, so the output doesn't have to be known.

This is part of how Nix accomplishes its reproducible build guarantees.

elm-src/NixOptions.elm
prikhi: Seems like you want:
```
List.all (\term -> term_matches name term || term_matches option.description term) terms
```
This ensures every term is either in the name or description.

elm-src/index.elm
grahamc: No. Way!

elm-src/index.elm
grahamc: `options` is about 5,000 items and will probably only go up (but slowly.)

zwilias: I think I'd either write this using `if List.isEmpty model.options then`, or using a `case..of`

```elm
case model.options of
    [] ->
        do stuff

    _ ->
        do other stuff
```

elm-src/index.elm
grahamc: Ignorance :)

elm-src/NixJSON.elm
grahamc: Yes! So obvious now, thank you :) I think I originally made nix_list a separate function because I reused it elsewhere. This, still, should be its own :)

elm-src/NixJSON.elm
grahamc: I think it wasn't clear to me how I should  name my functions, and then effectively picked at random 😳 I'll clean these up :) thank you!

elm-src/NixOptions.elm
grahamc: 👍 Got it! Thank you!

elm-src/.gitignore
domenkozar: Yes, Nix separates download phase from build phase.

zwilias: Right. Got it. I suspect this process could be made a fair bit prettier in 0.19.

domenkozar: Indeed - I'm checking every week for an alpha :)

Makefile
```diff
@@ -236,3 +236,11 @@ nix/install.sig: nix/install
 		mv $@.tmp $@; \
 	fi
 	touch $@
+
+nixos/options.html: elm
+
+elm: elm-src/index.elm
+	nix-build ./elm-src --out-link ./elm
```
edolstra: Calling `nix-build` here breaks calling the Makefile from a Nix expression (as we do e.g. in https://github.com/NixOS/nixops/blob/master/examples/nix-homepage.nix).

edolstra:
Hm, I'm not super-enthusiastic about. It replace something that already worked with something that is significantly more complex (e.g. elm-src/default.nix), and the Elm code is actually longer than the Javascript code it replaces.

Also, I note that `elmPackages.elm` has a 1.6 GiB closure size. `nixos.org` doesn't have a lot of disk space in /nix so this is a pretty heavy dependency.

domenkozar:
> and the Elm code is actually longer than the Javascript code it replaces.

Yeah Elm is a bit more verbose, but it's a pure langauge so it doesn't fail at runtime (except when you compare functions).

> Also, I note that elmPackages.elm has a 1.6 GiB closure size. nixos.org doesn't have a lot of disk space in /nix so this is a pretty heavy dependency.

This is because Elm has multiple executables, we could actually fix that now that haskell multiple-outputs were merged. Elm 0.19 (due to be released soon) will have one executable and the closure should be ~50MB.

grahamc:
We'll revisit this when Elm 0.19 comes out.

Makefile
grahamc: But ... what about all the other nix-build calls in the Makefile? 😕

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,158 @@
...
+        , Decode.int |> Decode.map GInt
+        , Decode.float |> Decode.map GFloat
+        , Decode.list (Decode.lazy (\_ -> gsonDecoder)) |> Decode.map mkGsonList
+        , Decode.keyValuePairs (Decode.lazy (\_ -> gsonDecoder)) |> Decode.map mkGsonObject
+        ]
+
+
+mkGsonList : List Gson -> Gson
+mkGsonList data =
+    GList (List.map (\x -> x) data)
```
jb55: `List.map (\x -> x) data` is equivalent to `data`

elm-src/NixJSON.elm
```diff
@@ -0,0 +1,158 @@
...
+indent lines =
+    List.map (\line -> String.append "  " line) lines
+
+
+nix_enc_dict : ( String, Gson ) -> List String
+nix_enc_dict ( key, value ) =
+    [ "\""
+    , key
+    , "\" = "
+    , (nixFromGson value)
```
jb55: nit: redundant parens

elm-src/Paginate.elm
```diff
@@ -0,0 +1,27 @@
...
+import List.Extra
+
+
+which_page : List a -> (a -> Bool) -> Maybe Int
+which_page options matcher =
+    let
+        pos =
+            List.Extra.findIndex matcher options
+    in
+        case pos of
```
jb55: This whole case could be replaced by:

```elm
Maybe.map (\p -> ceiling (toFloat p / 15.0)) pos
```

domenkozar:
0.19 is out :-)

davidak:
What's the status?

My opinion is that the nixos website code should be consistent to be maintainable, so use as few dependencies as possible. Don't use JavaScript **and** Elm, use only one of them. Since JS is more popular, we might easier find contributors, but i understand that Elm has other advantages. Just don't make a mess of the project and keep it maintainable.

Maybe adding a new language/technology could go through the RFC process?

grahamc:
I thought I had closed this already.

grahamc:
Also, I don't think the website is under the RFC process right now.

PR #165, opened 2017-10-01
Delete the reference to Nixpkgs Monitor, it is down

grahamc:


Phreedom:
On Sunday, October 01, 2017 02:11:13 Vladimír Čunát wrote:
> vcunat approved this pull request.
> 
> /cc @Phreedom, just in case.

I can't fix it right away although I hope that eventually I'll be able to. We 
can always add it back.

vcunat:
BTW, there's a nice similar-style tool with a bit different focus points https://repology.org/

PR #172, opened 2017-11-11
NixOS layout: add manual, remove options+packages

lheckemann:
The previous layout lacked a manual link in the navigation bar at
the top, which suggested by analogy with Nix/Nixpkgs/NixOps that
NixOS did not in fact have a manual. To make space for it, I've
removed the "Packages" and "Options" links, which I've made sure
to still mention on the Support page along with the manual.

cc @srhb @teto


teto:
Thanks for the PR.
Was removing the "packages" necessary ? You can move it to nixpkgs topbar maybe ?

lheckemann:
Done, along with adjusting the URL at which it lives. There should probably also be a redirect at the old URL to avoid breaking people's links/browser history, but I'm not sure what the preferred approach here is.

grahamc:
I worry that the Options and Packages links are highly used, and removing them will make it harder for new users.

PR #185, opened 2018-02-24
Cache the release tarball

### LINKED ISSUE #184: Caching the release tarball
It's annoying to have to wait ~30s to re-download the release tarball after it throws an error (e.g. because of relics from previous installations). Could the release tarball be downloaded to a deterministic path in `/tmp`?

chrismwendt:
Previously, the user would have to wait ~30s to re-download the release tarball if installation throws an error (e.g. because of relics from previous installations).

Now, the tarball is downloaded to a deterministic temp file and only re-downloaded if it doesn't exist yet or the checksum doesn't match.

Closes #184

### DIFF
--- a/nix/install.in
+++ b/nix/install.in
@@ -10,17 +10,18 @@ oops() {
     exit 1
 }
 
-tmpDir="$(mktemp -d -t nix-binary-tarball-unpack.XXXXXXXXXX || \
-          oops "Can\'t create temporary directory for downloading the Nix binary tarball")"
-cleanup() {
-    rm -rf "$tmpDir"
-}
-trap cleanup EXIT INT QUIT TERM
-
 require_util() {
     type "$1" > /dev/null 2>&1 || which "$1" > /dev/null 2>&1 ||
         oops "you do not have '$1' installed, which I need to $2"
 }
+
+require_util curl "download the binary tarball"
+require_util bzcat "decompress the binary tarball"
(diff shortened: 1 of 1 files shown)


edolstra:
Using a fixed-name directory in /tmp creates a potential security problem. Might be better to use $XDG_RUNTIME_DIR on Linux.

chrismwendt:
@edolstra What kind of security problem do you have in mind? It seems to me that the worst case would be where an attacker overwrites the install script in the unpacked release directory with an arbitrary executable.

`$XDG_RUNTIME_DIR` would avoid that threat, but is it present (and not set to something insecure like `/tmp`) on all platforms?

Using `mkdir` as stated in http://www.linuxsecurity.com/content/view/115462/151/ would work, but it seems rather difficult to make sure the release directory was created by the current user in the case where the release directory already exists.

How about copying the tarball to a secure directory (i.e. one created by `mktemp -d`) before computing the SHA and extracting? I made that change and pushed it as a commit to this PR.

samueldr:
[triage]

This is now [part of the `nix` repository](https://github.com/NixOS/nix/blob/master/scripts/install.in).

The PR would need to be re-done there.

And yes, as of today, [the script does cleanup after itself](https://github.com/NixOS/nix/blob/9617a043541d77d79e4f20f9676aae63de72f45d/scripts/install.in#L15).

Issue #121, opened 2016-12-20
nixos-homepage has no license


grahamc:
Someone was talking about sending patches to this repo, but isn't able to unless a license is clearly specified, due to their work.

ToxicFrog:
That might have been me, but even if it wasn't, it's an issue for me too.

edolstra:
Anybody have a preference for a license? Maybe CC BY-SA?

ToxicFrog:
I thought CC-* were discouraged for code? Why not the same license as nixpkgs (MIT, looks like?)

edolstra:
Our homepage is mostly not code.

grahamc:
CC BY-SA seems fine to me.

davidak:
I also think CC BY-SA is a good license for websites. Should i create a PR with it?

PR #188, opened 2018-03-01
Add direct link to Manual for NixOS

markus1189:
I know that it is linked under the `Community` page, but all the others like `nix`, `nixpkgs` etc all have a direct link to the Manual on their site.

I missed this direct link on the `NixOS` often enough now that I think it would be worth it to add it.

Thoughts?

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -77,6 +77,7 @@
                 <li><a href="[%root%]nixos/options.html">Options</a></li>
                 <li><a href="[%root%]nixos/community.html">Community</a></li>
                 <li><a href="[%root%]nixos/security.html">Security</a></li>
+                <li><a href="[%root%]nixos/manual">Manual</a></li>
               </ul>
               <ul class="nav pull-right">
                 <!--


edolstra:
The navbar is getting awfully crowded. Maybe we can remove the link to the Security page? It doesn't seem like something that the average visitor needs to click on frequently. @grahamc

grahamc:
Agreed, we can move that to a sub-page (though I'm not sure which one,) or even the footer.

vcunat:
Well, "support" contains documentation, "community" as well.  Instead of direct manual link, we might consider "documentation".  Perhaps the categories might better get redesigned from scratch.

vcunat:
IMHO "support", "community" and "security" might be all together, as they feel closely related, but I don't know how to call that.

markus1189:
It felt weird to have the `Manual` link on all other sites except NixOS

c0bw3b:
> [...] or even the footer.

yes I tend to look for Legal / Privacy Policy / ToS / **Security** ... in the **footer** of some org homepage. 

Some examples:
https://coreos.com/
https://www.seafile.com/en/home/
https://sandstorm.io/

\+ this footer would also be a good place for #121 solution

markus1189:
closing due to age

Issue #134, opened 2017-04-20
No NixOS Manual Link on the Menu Bar & Lack of Consistent Parallelism in Menu Interface


ixxie:
I love the way the menu works on the nixos.org homepage - the idea of parallel menus for each element of the Nix ecosystem is brilliant.

However, the menus of these different elements are sometimes quite radically different. It is very understandable that some of them have less elements since they have less content; what is less understandable is when they have the same element reachable in different ways.

Case in point: in order to reach the NixOS manual, I must first click support and then press the manual; for Nix, NixOps, Nixpkgs, and Hydra I can reach the manual directly from the menu. Finding the Disnix manual is even more confusing than the NixOS one, although I understand if a pdf manual should not be a menu link.

If the maintainers of the site choose to take this under consideration and restructure the menu, a link to the github page of each project should also be considered, as it is becoming increasingly standard.

BlessJah:
Disnix is also affected: https://github.com/NixOS/nixos-homepage/issues/140

BlessJah:
![screenshot](http://i.imgur.com/gZCH9UU.png)

garbas:
This issues has been now address and hopefully provides an easier way to navigate.

Issue #106, opened 2016-10-23
consolidate all manuals into a Guide to the Nix ecosystem


FlorentBecker:
The manuals for nix, nixpkgs, disnix and nixops are hardly discoverable from the front page. Moreover, it is often not easy, given a question about one of the nix\* tools, where its answer will lie: changing the behavior of nixos might involve using a function from nix's base library. The manual linked from the front page should be a "Guide to the Nix ecosystem" linking to the other manuals, and there should be more links between the various manuals, at least in the "descending" direction (ie nixops > nixos > nixpkgs > nix).

garbas:
This has now been partially address on Learn page, where we made manual more discoverable, by exposing parts of the table of content of the manuals.

"Get started" button on the landing page will most likely point to some of the guides that are being worked on as part of #443.

Issue #56, opened 2015-10-20
Add Package search to Nix homepage


davidak:
Select channel: Linux or Mac, stable or unstable

<table class="source-tableeditor">
<tbody>
<tr>
<td></td>
<td>stable</td>
<td>unstable</td>
</tr>
<tr>
<td>Linux</td>
<td>nixos-17.09</td>
<td>nixos-unstable</td>
</tr>
<tr>
<td>Mac</td>
<td>nixpkgs-17.09-darwin</td>
<td>nixpkgs-unstable</td>
</tr>
</tbody>
</table>

since unstable channels are very active, the packages.json.gz should get generated at least once a day from github source, better every hour.

garbas:
this is being fixed now

davidak:
no, the request was to add a "packages" page with search to Nix page similar to NixOS, but directed to people who use Nix on macOS or linux distro other than NixOS. it don't make sense that this users go to NixOS page...

![Screenshot from 2020-03-16 23-49-39](https://user-images.githubusercontent.com/91113/76806177-048eee80-67e1-11ea-86d9-63c5243f87a4.png)

(thanks for cleaning up the issues btw :+1:)

PR #189, opened 2018-03-05
Splits the navigation in two navbars...

samueldr:
### Description of the changes

 * Makes the different projects more obvious.
 * Marks a clearer hierarchy in the website.

In addition, since the DOM had to be manipulated, and we
have a bit more space:

 * Re-adds github and twitter links for nixos project.

The dropdown for the projects is, as a side-effect from manipulating
the DOM, now gone from the mobile view. It is now a flat list of
projects, followed by the page list of that specific section. This
may help with issues like #106.

Furthermore, with the added room, a direct link to the manual has been added to the NixOS website, like #134 and #188 proposes.

* * *

### Motivation for the changes

This is to make the list of projects more approachable at a glance, and to make navigation more explicit between projects.

Though, another reason is that the horizontal real-estate was becoming a bit scarce, since the project name and the search box were present on the same row than the page list for the project. With this change, it is possible to add direct links to resources as needed.

* * *

### Expectations

 * Push back from needing a bit more vertical real-estate on the website.

* * *

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -380,6 +380,17 @@ div.docbook div.toc dd {
     min-height: 30px
 }
 
+/**
+ * Fixes searchbar on larger mobile responsive breakpoints.
+ */
+@media (max-width: 979px) {
+	.navbar.navbar-primary .search-query {
+		box-sizing: border-box;
+		display: block;
+		width: 100%;
+		height: 30px;
+	}
+}
 
 /**
  * Fixes logo metrics on desktop.
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -1,7 +1,7 @@
 /* Account for the top navbar. */
 
 body {
-    padding-top: 50px;
+    padding-top: 70px;
 }
 
 @media (max-width: 979px) {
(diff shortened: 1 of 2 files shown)


grahamc:
(without commenting on the merits of the change,) this needs rebasing before it can be tested

samueldr:
*rebased*

grahamc:
I think it looks like a pretty good idea, but the double gradient is a really bizarre to me. Maybe a visual change on that?

edolstra:
Not a big fan because it's more cluttered and many of the projects are not really relevant for a first-time visitor. (E.g. why should they care about NixOps/Hydra/Disnix?)

davidak:
Following the argument of @edolstra, what about mentioning NixOps/Hydra/Disnix only on the startpage of nixos.org in the style of Declarative/Reliable/DevOps-friendly, Combine Nix and Nixpkgs (https://github.com/NixOS/nixos-homepage/issues/199), add Package search to Nix homepage (https://github.com/NixOS/nixos-homepage/issues/56) and move the Nix site to nix.org (that we might be able to buy).

So NixOS users and Nix users on Linux and macOS have a different home.

That way we get rid of the menu completely. (but break many links)

samueldr:
*Closing as this **won't be accepted as-is**.*

This was mostly an off-the-cuff proposal I did to gather opinions and comments.

PR #190, opened 2018-03-05
Centers/Aligns `get *` buttons with the snowflake.

### LINKED ISSUE #67: "Get Nix[OS]" buttons are not centered under the new logo
Probably fairly easy to fix, but it looks a little awkward right now.

samueldr:
This fixes #67.

* * *

The fix is only applied on "non-mobile" responsive views, as it otherwise adds a *weird margin* to the button, as the button is further from the logo on those resolutions.

An alternative look (e.g. full width) could be defined for the narrower responsive resolutions if desired.

* * *

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -343,3 +343,13 @@ div.docbook div.toc dd {
 .license li:last-child:after {
     content: "";
 }
+
+/* Centers the `get` button with the image.
+ * 196px is the width of the image.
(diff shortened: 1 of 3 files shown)


grahamc:
Tested and LGTM

grahamc:
Thanks Sam and Eelco!

samueldr:
Thanks!

Issue #89, opened 2016-07-17
Option description is truncated


danbst:
For example, `boot.initrd.network.enable`:

```
    boot.initrd.network.enable = mkOption {
      type = types.bool;
      default = false;
      description = ''
        Add network connectivity support to initrd. The network may be
        configured using the <literal>ip</literal> kernel parameter,
        as described in <link
        xlink:href="https://www.kernel.org/doc/Documentation/filesystems/nfs/nfsroot.txt">the
        kernel documentation</link>.  Otherwise, if
        <option>networking.useDHCP</option> is enabled, an IP address
        is acquired using DHCP.
      '';
    };
```

is viewed like

```
Description:        Add network connectivity support to initrd. The network may be configured using the ip kernel parameter, as described in
Default value:      false
Example value:      Not given
Declared in:        nixos/modules/system/boot/initrd-network.nix
```

garbas:
This should not be the case anymore

PR #191, opened 2018-03-05
Fixes xml issues in description field for options.html

samueldr:
### Fixes links in description fields

To verify:

 * Search using the <input> for `<link`.
 * An example option with link: `boot.initrd.luks.mitigateDMAAttacks`

I say this finishes fixing #89 assuming that [the commit that fixed it *partially*](https://github.com/NixOS/nixos-homepage/pull/97) has this description:

> links are still not rendered as HTML links

This now renders the links.

### Styles literal and options

To verify:

  * `boot.extraTTYs` has both.

* * *

If there are extra tags I missed, don't hesitate to open an issue and tag me in it.

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -282,6 +282,14 @@ div.search-details td.nix {
     font-family: monospace;
 }
 
+#search-results .description.docbook p > .docbook-important {
+	margin-top: 0.5rem;
+}
+
+#search-results .description.docbook .docbook-important {
+	border-left: #F89406 solid 0.3rem;
+	padding-left: 0.5rem;
+}
 
 /* Manual. */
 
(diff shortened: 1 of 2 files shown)


grahamc:
How did you test this? When I try it I get an error saying it can't load option data.

samueldr:
Ah, I was running it out of an HTTP server, otherwise it becomes way harder to test json stuff:

```
exec ruby -run -ehttpd . -p8000
```

The equivalent python directive (and other language directives) should work too. That's a really basic HTTP server.

grahamc:
Tested and LGTM

samueldr:
Hi!

[Saw this earlier today](https://logs.nix.samueldr.com/nixos-dev/2018-04-16#1523871687-1523871695;) and thought I could sneak in a fix!






* * *

Furthermore, while searching for other stray XML tags, I have seen that `<important>` weren't parsed. This means that the important bits were run-in with the main paragraph! See `virtualisation.virtualbox.host.enableHardening`.

edolstra:
Thanks!

PR #192, opened 2018-03-07
options, packages: fetch a compressed version only on nixos.org

grahamc:
Also change the Makefile to produce packages.json, options.json as
well as options.json.gz and packages.json.gz

This makes it quite a bit easier to test the website locally, and you don't even need a server to do it. Just visiting the HTML file in your browser will do the trick.


samueldr:
Tested, and verified it loads the `.json` file locally.

The `document.domain` bit for nixos.org should be good enough, though it would fail if e.g. it was `staging.nixos.org`, `www.nixos.org` or if served anywhere else. It may be more wise to use non-gzip on `127.0.0.1` and default to gzip in other cases?

grahamc:
When we have a staging.nixos.org or a www.nixos.org we can fix it ;)

PR #206, opened 2018-04-07
doc: overrides `hljs` overflow rules for better linking.

samueldr:
This sets the `pre`'s `overflow-x` to `visible`, which is the initial value, which is overridden by the new `hljs` dependency. Anchor links inside boxes with an `overflow` or `position` set cannot account for the top navbar's height in the first 80px of an element, this is a limitation of how CSS works. This could be worked-around using JavaScript (I think), but the previous behaviour did not cause any issues AFAIK.

###### Before

After clicking on `(1)`

###### After

After clicking on `(1)`

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -326,6 +326,10 @@ div.docbook div.toc dd {
     margin-left: 2em;
 }
 
+div.docbook pre.hljs {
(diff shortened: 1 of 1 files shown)


grahamc:
LGTM

PR #211, opened 2018-04-17
install: use tar -f -

### LINKED ISSUE #169: install.in: tar in debian uses $TAPE
when TAPE is set, tar implicitly sets -f $TAPE, breaking the install. The installer should unset TAPE or, better, use `-f -` in the tar call.
https://github.com/NixOS/nixos-homepage/blob/master/nix/install.in#L59

woffs:
fixes #169

### DIFF
--- a/nix/install-1.11.16
+++ b/nix/install-1.11.16
@@ -57,7 +57,7 @@ fi
 
 unpack=$tmpDir/unpack
 mkdir -p "$unpack"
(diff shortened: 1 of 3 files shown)


edolstra:
Not sure if we should change the install scripts for old versions, since the Makefile doesn't currently regenerate the .sig files.

edolstra:
Thanks, done!

PR #220, opened 2018-05-21
Replace nix-devel by Discourse

zimbatm:
This is a proposal to make the switch to reduce the number of discussion platforms that we use.

### DIFF
--- a/disnix/support.tt
+++ b/disnix/support.tt
@@ -8,8 +8,7 @@
   href="irc://irc.freenode.net/#nixos"><tt>#nixos</tt></a> on <a
   href="https://freenode.net/"><tt>irc.freenode.net</tt></a>.</li>
 
-  <li>You can use the <strong>mailing list</strong> <a
-  href="https://groups.google.com/forum/#!forum/nix-devel"><tt>nix-devel</tt></a>.</li>
+  <li>You can use the <a href="[%root%]nixos/support.html#discourse"><tt>Discourse</tt> forum</a>.</li>
 
   <li>You can <strong>report bugs</strong> in the <a
   href="https://github.com/svanderburg/disnix/issues">Disnix issue
--- a/news.xml
+++ b/news.xml
@@ -2,6 +2,17 @@
 
 <news>
   <item>
+    <pubDate>Tue Aug 14 2018 00:00:00 GMT</pubDate>
+    <title>
+      NixOS Discourse forum
+    </title>
+    <description>
+      The <tt>nix-devel</tt> mailing list is now replaced by our discourse forum instance which is also usable by email:
+      <a href="https://discourse.nixos.org"><tt>discourse.nixos.org</tt></a>.
+    </description>
+  </item>
+
+  <item>
     <pubDate>Mon May 21 2018 00:00:00 GMT</pubDate>
     <title>
       NixCon 2018
--- a/nixos/community.tt
+++ b/nixos/community.tt
@@ -23,7 +23,7 @@
 please open a <a
 href="https://github.com/NixOS/nixpkgs/pulls"><strong>pull request on
 GitHub</strong></a>, or <strong>send a patch</strong> to the <a
-href="[%root%]nixos/support.html#mailing-list"><tt>nix-devel</tt> mailing list</a>. If you
+href="[%root%]nixos/support.html#discourse"><tt>Discourse</tt> forum</a>. If you
 want to contribute regularly, you may want to ask for commit access to
 our GitHub repositories (please ask <a
 href="https://nixos.org/~eelco/">Eelco</a>, or on the <tt>#nixos</tt>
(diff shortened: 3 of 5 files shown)


edolstra:
Seems a bit premature. We should first see how well Discourse works out.

ryantm:
We added a Discourse section to the community page in https://github.com/NixOS/nixos-homepage/pull/218

zimbatm:
@edolstra do you have any specific concerns or is X more time enough? If the latter it would be good to set a deadline.

garbas:
:+100: on discourse.nixos.org, but it would be very nice to import mailing list as we do the switch (https://discourse.nixos.org/t/import-the-mailing-list/66)

zimbatm:
How about now? Most of the traffic has died on the mailing list

news.xml
```diff
@@ -16,6 +16,17 @@
     </description>
   </item>
 
+  <item>
+    <pubDate>Mon May 21 2018 00:00:00 GMT</pubDate>
+    <title>
+      Nix-devel mailing list moved
```
edolstra: The subject should be something like "NixOS Discourse forum" since we didn't "move" the mailing list.

nixos/community.tt
```diff
@@ -23,7 +23,7 @@
 please open a <a
 href="https://github.com/NixOS/nixpkgs/pulls"><strong>pull request on
 GitHub</strong></a>, or <strong>send a patch</strong> to the <a
-href="[%root%]nixos/support.html#mailing-list"><tt>nix-devel</tt> mailing list</a>. If you
+href="[%root%]nixos/support.html#mailing-list"><tt>Discourse</tt> mailing list</a>. If you
```
edolstra: Do we actually want people posting patches on Discourse? TBH I would prefer to remove the whole "send a patch" thing.

nixos/community.tt
```diff
@@ -84,7 +84,7 @@ you:</p>
 <h3>Mailing list</h3>
 
 <p>NixOS development is coordinated on the <a
-href="[%root%]nixos/support.html#mailing-list"><tt>nix-devel</tt>
+href="[%root%]nixos/support.html#mailing-list"><tt>Discourse</tt>
```
edolstra: Is this actually true? I think it is more coordinated on GitHub and IRC if anything.

disnix/support.tt
```diff
@@ -8,8 +8,7 @@
   href="irc://irc.freenode.net/#nixos"><tt>#nixos</tt></a> on <a
   href="https://freenode.net/"><tt>irc.freenode.net</tt></a>.</li>
 
-  <li>You can use the <strong>mailing list</strong> <a
-  href="https://groups.google.com/forum/#!forum/nix-devel"><tt>nix-devel</tt></a>.</li>
+  <li>You can use the <a href="[%root%]nixos/support.html#mailing-list"><tt>Discourse</tt> mailing list</a>.</li>
```
edolstra: In several places this PR refers to Discourse as a mailing list. Why not just "Discourse" or the "Nix community forum"? Just because it has a mail gateway doesn't make it a mailing list.

news.xml
zimbatm: Thanks, fixed

nixos/community.tt
zimbatm: As long as there are people willing to apply the patches it doesn't hurt

zimbatm: The rule still applies that a github maintainer needs to be assigned to each packages and modules

nixos/community.tt
zimbatm: fixed in 76a4f50159e3ae9694692e3b377a4d0cc5ccfe77

disnix/support.tt
zimbatm: Okay. I changed all references to "Discourse" or "Discourse forum" in 76a4f50159e3ae9694692e3b377a4d0cc5ccfe77

nixos/support.tt
```diff
@@ -14,16 +14,13 @@
...
     <ul>
 
-    <li><a href="https://groups.google.com/forum/#!forum/nix-devel"><tt><strong>nix-devel@<wbr/>googlegroups<wbr/>.com</strong></tt></a>
+    <li><a href="https://discourse.nixos.org/"><tt><strong>discourse.nixos.org</strong></tt></a>
     is for all user and development questions and discussions. To
-    subscribe, visit the <a
-    href="https://groups.google.com/forum/#!forum/nix-devel">nix-devel
-    Google Group</a> or send an email to <a
-    href="mailto:nix-devel+subscribe@googlegroups.com">nix-devel+subscribe@googlegroups.com</a>.</li>
+    subscribe, visit the page and subscribe with GitHub or email/password.</li>
```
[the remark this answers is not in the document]
zimbatm: The new sentence is:
> Discourse is a forum where we have discussions and help users out with their questions. Visit the page and register with GitHub or email/password: https://discourse.nixos.org/

edolstra:
Thanks @zimbatm!

PR #222, opened 2018-05-30
Revert "Nix 2.0.3 released"

shlevy:
A bug in the installer prevents installing when the calling shell
has set -u.

This reverts commit 9124e272a0f53673ce8747d2054e31716b26e3ec.


shlevy:
Note this is only a partial revert, the actually 2.0.3 installer and its signatures are kept.

PR #228, opened 2018-06-30
add coqPackages, idrisPackages, nodePackages, quicklispPackages to package list

ryantm:
I looked over all the package sets at the top level to find ones not in the current package listing; this change adds the coq, idris, node, and quicklisp sets. The uncompressed archive got 1 MB bigger from this change, and the compressed one got less than 100 KB bigger.

### DIFF
--- a/discover-more-packages-config.nix
+++ b/discover-more-packages-config.nix
@@ -4,12 +4,16 @@
 
 {
 
-  packageOverrides = super: {
-
-    haskellPackages = super.recurseIntoAttrs super.haskellPackages;
-
-    rPackages = super.recurseIntoAttrs super.rPackages;
-
-  };
-
+  packageOverrides = super:
+    builtins.foldl'
+     (s: packageSet: s //
+       { "${packageSet}" = super.recurseIntoAttrs super."${packageSet}"; })
+     {} [
+          "coqPackages"
+          "haskellPackages"
+          "idrisPackages"
+          "nodePackages"
+          "quicklispPackages"
+          "rPackages"
+        ];
 }


vcunat:
Well, alternatively we can consider actually building (some of) them: https://github.com/NixOS/nixpkgs/issues/45849

ryantm:
@vcunat Is building them related to whether we list them?

vcunat:
Yes, they should get listed, at least if we use the usual `recurseIntoAttrs` to build them.

ryantm:
Are you saying that we should not list packages we do not build?

samueldr:
⚠️ not sure we should do this. Even thinking about proposing to revert the change *for the packages listing on the website*. If this is all for the ~~popularity contest~~ repology data source, we may want to generate those separately.

See #245

ryantm:
There are serious reasons to get this data into Repology, r-ryantm has some chance to update these packages, and we can receive more feedback about how our versions of these packages compare to other packaged versions that exist.

I'm only doing this PR to get the data into Repology, I personally don't care too much about it appearing on the website.

samueldr:
Yeah, and sorry if it sounded dimissive with the "popularity contest" joke. The ability to detect package issues is not to be overlooked.

I would even prefer having that data available to the end-users, but not at the cost of UX :/.

davidak:
@samueldr do 100 KB really make a noticable difference for UX?

Merge conflict because the file was renamed to https://github.com/NixOS/nixos-homepage/blob/master/packages-config.nix

samueldr:
I don't really know. I'm not sure where the deeper performance issues lie. It's likely that there's just too much to index for the current approach at searching the package set.

I'm not sure I have the time to check using the developer tools the different between before and after this PR. If the effects are negligible, then I guess it should be fine. Though, I'm not sure what's a good target machine to test, a top of the line machine with ample ram is likely to get a different experience than a low-to-mid tier machine from a couple years back :/. And then, there's mobile use, which shouldn't be discounted, ideally.

garbas:
I'm closing this PR but we will enable this subpackage lists in https://github.com/NixOS/nixos-search

PR #229, opened 2018-07-01
Removes useless escapes from `ppNix` in options page.

samueldr:
This is safe since all calls to `ppNix` are made using `.text()`.

 * https://github.com/NixOS/nixos-homepage/blob/f74049df65db247ce156022155a966c272ffb96c/nixos/options.tt#L324-L328

And *anyway* the previous escapes were done only one time, making it
trivial to bypass

```
> "<<<<<".replace("<", "&gt;")
→ "&gt;<<<<"
```

This has further been manually verified (with an hand-modified options.json) to
not cause opportunities for HTML injection.

image

image

### DIFF
--- a/nixos/options.tt
+++ b/nixos/options.tt
@@ -210,7 +210,7 @@ function ppNix(indent, v) {
     // their content as it cannot be serialize properly. So if we detect such
     // example, escape characters and return the text without pretty-printing.
     if (v._type == "literalExample") {
-      v = ((v.text || '') + '').replace('<', '&lt;').replace('>', '&gt;');
+      v = ((v.text || '') + '');
       return v;
     }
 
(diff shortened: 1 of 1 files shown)


samueldr:
As reported on IRC by @infinisil

 * https://logs.nix.samueldr.com/nixos/2018-07-01#1530476707-1530476774;

nixos/options.tt
```diff
@@ -243,7 +243,7 @@ function ppNix(indent, v) {
   }
 
   if (typeof v == "string") {
-    v = v.replace('<', '&lt;').replace('>', '&gt;');
+    v = v;
```
edolstra: I guess `v = v` is redundant?

samueldr: You're right, I removed the same pattern from both location (and tested both cases) but didn't actually look at what was left at the second location. It's now fixed.

PR #233, opened 2018-08-23
about.tt: update number of packages

magnetophon:


samueldr:
Hi, I feel a bit silly asking for the change, but the thousands separator is [traditionally a comma, but new standards and practices prefer using a thin space instead](https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping).

Since the website is in English, even though it's ridiculous to think it would be confused for 40 packages, I think it would be preferrable to use a thin space (`&thinsp;`) instead of the colon.

magnetophon:
@samueldr NM, I'll do both changes in one commit and make a new PR.

magnetophon:
[There](https://github.com/NixOS/nixos-homepage/pull/235) you go!

PR #234, opened 2018-08-23
index.tt: update number of packages

magnetophon:


samueldr:
Hi, I think the two PRs come from using the github interface, I guess it's fine. See my comment on #233.

Here, a comma was previously used, and has now been replaced with a colon. This PR could realistically set the tone for using a comma in the other PR instead of a thin space. Whichever is chosen, it probably should match in both PRs.

PR #238, opened 2018-09-01
Use nix-build to build the site

grahamc:
Suuuuper impure but what is make in a loop?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,154 @@
+{ pkgs ? import <nixpkgs> {} }:
+with rec {
+  loadNixpkgs = version:
+    let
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -0,0 +1,154 @@
...
+
+  hydra = let
+      src = builtins.fetchGit {
+        url = "https://github.com/nixos/hydra.git";
+        ref = "master";
+      };
+      imported = import "${src}/release.nix" {
+        hydraSrc = src;
+      };
+    in imported.manual;
```
grahamc: the above two aren't great. These two, and the other impure parts, could be made as a pre-nix-build step.

grahamc:
The results of this, could, say, be uploaded to s3.

Just playing around, really.

PR #242, opened 2018-09-27
RSS fixes

jameysharp:
Here are an assortment of fixes for the `news-rss.xml` RSS feed. Most of these commits just fix issues reported by http://www.feedvalidator.org although there are a couple of notable side effects.

1. Since feed items did not include a `<guid>` before, adding one causes all items to appear unread. This is unfortunate but avoids bigger potential problems later.

2. Since the date format in `news.xml` was not consistent with the RSS specification (which requires RFC-822 format), the HTML version of the news changed from "month day year" order to "day month year". This is fixable in a couple of different ways if it matters to you.

As long as I was doing all that, I also took care of my personal crusade, which is RFC5005 "Feed Paging and Archiving" support. Hardly anybody supports it, but it's really easy to do if you're sticking your complete history in the feed anyway, so I figured I might as well.


samueldr:
Hi! Thanks for the contribution!

This repository is much less frequented and has much fewer people with approval and merge rights; don't be alarmed (yet) by the lack of traffic around this PR.

* * *

:+1: eyballed each commits; good separation, easy to understand. I also verified using `nix-shell --run "make && python2 -m SimpleHTTPServer 8000"` that everything built as expected. I also verified that the escapes in the feed are as expected.

LGTM :)

PR #247, opened 2018-10-25
Fixes the package/options search user experience

caugner:
* Doesn't discard input during load anymore.
* Makes the input more responsive by debouncing the filtering.


caugner:
This PR has been superseded by #209.

PR #250, opened 2018-11-09
Drops aliases from the packages listing

### LINKED ISSUE #225: Remove aliases from package search?
Package search for "inotify-tools" gives 2 results. It's the exact same package. The second came from an alias.

https://nixos.org/nixos/packages.html#inotify-tools

Does it make sense to list aliases there or should using the original name be promoted?

Same behavior in https://github.com/NixOS/nixos-homepage/pull/209:

https://nix.samueldr.com/explorer/?channel=nixos-18.03&query=inotify-tools

samueldr:
This promotes the canonical names, and reduces confusion as to why some packages are listed multiple times.

Fixes #225

* * *

This also renames the configuration file as it becomes more general purpose. Also re-formats it.

cc @ryantm @davidak 

* * *

Before:

> image

After:

>  image

Listed packages count goes from 49403 to 49126. Not a huge amount, but still useful for the end-users.


ryantm:
This is going to break my [pending PR](https://github.com/NixOS/nixos-homepage/pull/228); but this looks good to me. Please merge!

PR #259, opened 2019-01-27
add instructions for google cloud platform

phunehehe:
Mostly copied it from the section about Azure. The `make` targets are tested but I'm not too sure how to render the document.


edolstra:
Thanks, it renders fine. Two comments:
* 16.03 is pretty old, maybe we can provide a newer version?
* The download page is not really the best place for extensive installation instructions. Maybe this could be added to the NixOS manual?

phunehehe:
I'm not sure where you see 16.03? In the `gce-images.json` file generated on my computer `latest` points to 18.09.
I don't really have an opinion about where this information should be. Currently the NixOS manual only has a passing line mentioning Amazon, whereas the download page has instructions to get NixOS through various ways. That's the only reason I feel it's natural to add GCP there.

phunehehe:
@garbas I notice that new GCP images aren't being generated, but thanks for picking this PR up \o/

PR #263, opened 2019-02-05
Minify packages.json with jq

tilpner:
Removing whitespace from packages.json (jq -c .) reduces the gzip-ed size from 3.1M to 2.7M, and the extracted size from 62M to 36M, which might help with parsing and decompressing it quicker.


tilpner:
I kept the call to mjson.tool, which seems to be used for JSON syntax checking. While jq should also validate the syntax, I don't know if it would report that error as neatly/at all.

PR #267, opened 2019-03-11
Replaced Cloudfront URLs with NixOS.org URLs

gurjeet:
In commit 70bcfbb the prefix of the download URLs was changed from

`https://nixos.org/releases/nixos/...`

to

`https://d3g5gsiof5omrk.cloudfront.net/nixos/...`

This doesn't seem ideal since the URL is pointing to a non-NixOS URL. I have
verified that the replacing the Cloudfront URL with `https://releases.nixos.org/nixos/...`
seems to work for all the URLs we care about.

Hence changing the prefix back to NixOS.org's site seems right thing to do.

### DIFF
--- a/nixos/download.tt
+++ b/nixos/download.tt
@@ -30,7 +30,7 @@ installer as well as X11, Plasma 5 Desktop and several applications.  It’s a
 <em>live CD</em>, so it allows you to get an impression of NixOS (and
 the Nix package manager) before installing it.</p>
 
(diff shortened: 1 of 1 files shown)


gurjeet:
bump...

/cc @edolstra

gurjeet:
@domenkozar, thanks for assigning a reviewer to the issue!

@grahamc, I understand it's a minor update, but it'd be great to get a closure on the issue.

grahamc:
Thank you!

PR #270, opened 2019-03-27
Build unstable manuals

oxij:
For NixOS/nixpkgs#58098.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ rwildcard=$(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2) $(filter $(subst 
 
 default: all
 
+# HTML stuff
 
 HTML = index.html news.html \
   nix/index.html nix/about.html nix/download.html \
@@ -18,6 +19,9 @@ HTML = index.html news.html \
   disnix/extensions.html disnix/examples.html disnix/support.html \
   nixops/index.html
 
+html: $(HTML) favicon.png $(subst .png,-small.png,$(filter-out %-small.png,$(wildcard nixos/screenshots/*))) \
+
+all: html
 
 ### Prettify the NixOS manual.
 
@@ -98,10 +102,9 @@ $(NIXPKGS_MANUAL_IN):
 	nix-build -o $@ '<nixpkgs/doc>' -I nixpkgs=$(NIXPKGS)
 
 
-all: $(HTML) favicon.png $(subst .png,-small.png,$(filter-out %-small.png,$(wildcard nixos/screenshots/*))) \
-  nixpkgs/packages.json.gz \
-  nixpkgs/packages-unstable.json.gz \
-  nixos/options.json.gz
+all: nixpkgs/packages.json.gz \
+     nixpkgs/packages-unstable.json.gz \
+     nixos/options.json.gz
 
 
 ### Prettify the Hydra manual.
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,18 @@ NIXOS_MANUAL_OUT = nixos/manual
 all: $(NIXOS_MANUAL_OUT)
 
 $(NIXOS_MANUAL_OUT): $(NIXOS_MANUAL_IN) bootstrapify-docbook.sh bootstrapify-docbook.xsl layout.tt common.tt
-	./bootstrapify-docbook.sh $(NIXOS_MANUAL_IN)/share/doc/nixos $(NIXOS_MANUAL_OUT) 'NixOS manual' nixos https://github.com/NixOS/nixpkgs/tree/master/nixos/doc/manual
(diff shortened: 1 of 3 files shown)


nixos/support.tt
```diff
@@ -11,6 +11,8 @@
     the list of configuration options by running <tt>man
     configuration.nix</tt>. You can also <a href="options.html">search
     the list of options online</a>.</p>
+
+    <p>There is also <a href="[%nixosManual%]/unstable/">a version for the <strong>nixos-unstable</strong> channel</a>.</p>
```
samueldr: Most of the file is soft-wrapped at around 75 chars. One simple newline would be welcome (imo).

```diff
diff --git a/nixos/support.tt b/nixos/support.tt
index b6b6763..5ccb27d 100644
--- a/nixos/support.tt
+++ b/nixos/support.tt
@@ -12,7 +12,8 @@
     configuration.nix</tt>. You can also <a href="options.html">search
     the list of options online</a>.</p>
 
-    <p>There is also <a href="[%nixosManual%]/unstable/">a version for the <strong>nixos-unstable</strong> channel</a>.</p>
+    <p>There is also <a href="[%nixosManual%]/unstable/">a version for the
+	<strong>nixos-unstable</strong> channel</a>.</p>
   </div>
 
   <div class="span4" id="discourse">
```

Makefile
```diff
@@ -90,18 +101,24 @@ NIXPKGS_MANUAL_OUT = nixpkgs/manual
...
-	./bootstrapify-docbook.sh $(NIXPKGS_MANUAL_IN)/share/doc/nixpkgs $(NIXPKGS_MANUAL_OUT) 'Nixpkgs manual' nixpkgs https://github.com/NixOS/nixpkgs/tree/master/doc
-	ln -sfn manual.html $(NIXPKGS_MANUAL_OUT)/index.html
+	@echo rm -rf $@
+	./bootstrapify-docbook.sh $(NIXPKGS_MANUAL_IN)/stable/share/doc/nixpkgs $(NIXPKGS_MANUAL_OUT)/stable 'Nixpkgs manual' nixpkgs https://github.com/NixOS/nixpkgs/tree/master/doc
+	ln -sfn manual.html $(NIXPKGS_MANUAL_OUT)/stable/index.html
+	# put unstable manual under $(NIXOS_MANUAL_OUT)/unstable
+	./bootstrapify-docbook.sh $(NIXPKGS_MANUAL_IN)/unstable/share/doc/nixpkgs $(NIXPKGS_MANUAL_OUT)/unstable 'Nixpkgs manual' nixpkgs https://github.com/NixOS/nixpkgs/tree/master/doc
+	ln -sfn manual.html $(NIXPKGS_MANUAL_OUT)/unstable/index.html
+	# alias for stable
+	ln -sfn stable/index.html $(NIXPKGS_MANUAL_OUT)/index.html
```
samueldr: This will orphan a `https://nixos.org/nixpkgs/manual/manual.html` page, that seems to be un-indexed, but still will exist.

“But there's an `rm`!” or so I thought initially, but it's a no-op print of `rm -rf`; `@echo` is an unintuitive construct in Makefiles. (And I'm not targeting you, this seems borrowed from another target.)

What about fixing the `rm -f`? I'm not sure it's a good idea, this will, periodically [(every twenty minutes or so)](https://github.com/NixOS/nixos-org-configurations/blob/517cabda83b1e0678e39e4dd8c221308e547d8ca/nixos-org/webserver.nix#L230) cause the manuals to briefly vanish from the server while they're being generated. Tried it locally and it's trivial to see it; this will end up confusing users, I think.

What should be done? Not sure. About the orphaned manual.html file? Maybe manually remove it. Should we `mv manual.html index.html` instead of symlinking to prevent unintended files to be present? Could we generate into a temp dir and as-atomically-as-possible pivot the manual to it instead of `rm` ing for the shortest amount of time the manuals?

Same comments about `rm` applies to the nixos manual, except there isn't a manual.html file there.

samueldr:
In addition to the review comments, I'm not sure the `nixos.org/*/manual/[index.html] -> nixos.org/*/manual/stable/[index.html]` symlink is entirely good. I may be overthinking URLs, but I'm thinking this could potentially cause issues where this is seen as duplicated content on the site. It'd be better if this could be handled through a permanent HTTP redirection, since IMO having a `/stable` and `/unstable` discrete location is good to convey which manual exactly it is; it might slightly reduce confusion as to why stuf on master isn't in the `/stable` manual.

Though, the annoying bit is how HTTP redirections aren't handled through configuration in this repo; it would need to be handled in the nixos org configurations, and might get even more involved to implement if/when the website is ever moved directly on S3.

samueldr:
With #505 done, I think this can be closed.

If anything is missing, feel free to say.

PR #273, opened 2019-04-29
Mobile friendliness

domenkozar:
Google Console reports:

2019-04-29-104348_1221x396_escrotum

I hope this resolves at least a few :)

cc @samueldr

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -2,6 +2,8 @@
 
 body {
     padding-top: 50px;
+    /* a bit more mobile friendly */
+    font-size: 16px;
 }
 
 @media (max-width: 979px) {
@@ -97,6 +99,11 @@ pre {
 .nix-shell-command:before {
     color: #999;
     content: " [nix-shell]$ ";
+}
+
+/* mobile friendly */
+div.main {
+    word-wrap: break-all;
 }
 
 div.main li {
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -103,7 +103,7 @@ pre {
 
 /* mobile friendly */
 div.main {
-    word-wrap: break-all;
+    overflow-wrap: break-word;
 }
 
 div.main li {


samueldr:
I'm thinking this won't fix anything in there, that the 778~780 pages are the [mailing list archives](https://nixos.org/nix-dev/) and the [misc. pages under ~eelco](https://nixos.org/~eelco/) (and maybe a few other stragglers like non-bootstrapped generated manuals).

I say this considering the error "Viewport not set", while it is definitely set for the main website.

Is there a list of pages that failed validation, or only that unhelpful count?

domenkozar:
Forgot to mention: there is a list of pages, so this aims for fixing the manual.

Viewport error however applies there as well, so I'm a bit puzzled (although manual was indeed too long due to word wrapping)

css/nixos-site.css
```diff
@@ -99,6 +101,11 @@ pre {
     content: " [nix-shell]$ ";
 }
 
+/* mobile friendly */
+div.main {
+    word-wrap: break-all;
```
samueldr: `break-all` will break all words eve
n where it mak
es no sense to the read
er.
image

The hash that is highlighted is the cause of the breaking. What could instead be done is to add zero-width breaks into the long hash(es) of the page, when they are not already in pre-formatted elements.

samueldr: To further explain, [`break-all`](https://www.w3.org/TR/css-text-3/#word-break-property) is intended more for when typesetting assumes it will be broken at the character boundary, the main example being some styles in the CJK families.

domenkozar:
```suggestion
    overflow-wrap: break-word;
```

@samueldr this seems to do the job: https://jsfiddle.net/4ouc90p2/

samueldr: Yes, just tested, this is fine. It might cause the same weird break in a word if somehow we had one of those big impossibly long one where there is no other place to break, but at that point it's a non-issue.

css/nixos-site.css
```diff
@@ -2,6 +2,8 @@
 
 body {
     padding-top: 50px;
+    /* a bit more mobile friendly */
+    font-size: 16px;
```
samueldr: Since this is intended for mobile devices, let's put it into a media query:

```
/* Landscape phone to portrait tablet */
@media (max-width: 767px) { ... }
 
/* Landscape phones and down */
@media (max-width: 480px) { ... }
```

My main issue is with how everything was made with the default font size in mind, with absolute font size units, which will only bring the unchanged font sizes up.

domenkozar: I understand your cautiousness, but I've checked the rendered homepage and there don't seem to be problems.

Note that 16px is the default in Bootstrap 4, so I think this goes into the right direction making text a bit easier to read.

samueldr: It's [more involved than 16px](https://github.com/twbs/bootstrap/blob/09c99fe6e79ad397fdef262017d7b2bc49377b15/scss/_variables.scss#L278); basically it relies on the browser's settings, and most defaults will end up being equivalent to 16px.

While I do agree generally with the "default 16px" being better, my previous point still stands that everything has been conceived with bootstrap's default font-size in mind, not only our additions, but bootstraps' defaults.

This includes some subtle, but imo breaking changes like the line-height set in absolute units to 20px, and some misc paddings and margins. The line-height, especially, makes the website feel much more cramped, and lines squeezed together so closely might hamper readability more.

domenkozar: Could.you provide an example of such occourance?

samueldr: I don't know of specifics for padding and margins, almost nothing is set side-by-side in the site, so that part is not as much an issue.

Though, most of the text in the site inherits the absolute unit line-height. 

> image
> image

> image
> image

The reduced line spacing hampers readability, it should at least be increased accordingly. To be equivalent, it should be set to 23px on body.

Here's how it would look with a line-height of 23px on the body.

> image

domenkozar:
Sadly lost interested in this, maybe someone else will pick it up

Issue #248, opened 2018-10-26
Better landing page (nixcon followup)


samueldr:
Writing up some of the points we brought up in an informal group talk this morning at nixcon.

The couple needs I remember from this:

 * Landing page for `nixos.org/` should advertise Nix as well as NixOS
 * Landing page would need good "stories" to hook in users
 * Landing page would need an easy and well marked road to the nix installer (so users can try easily)
 * The manual**s** should be easier to find

* * *

Additional note for the manuals: when on a non-specific subsection of the website, it would be helpful to link to a Manual**s** page, which has the multiple (three) manuals linked, with an introductory paragraph explaining why one would read that manual.

* * *

So this probably means we would have four menu structures, one for non-specific pages (e.g. community, support, manuals), and then specific menus when on pages about one of the sub-projects. This is something that needs exploring: would the non-specific menu be shown anyways on the specific sub-websites? Not showing it might become confusing.

* * *

> *This is only a rough reminder of what we talked about rapidly with @garbas earlier today. I'm probably missing a bunch.*

davidak:
>Landing page would need good "stories" to hook in users

I also thought about this. NixOS and Nix are great for different groups of people, so we should address them directly.

A good discussion with ideas is here: https://discourse.nixos.org/t/how-would-you-pitch-nix-os-if-you-had-5-minutes/3270/2?u=davidak

I would like to have very little, but precise text on the landing page together with symbols or images to visualize it. A new user should not be overwhelmed by the complexity of the ecosystem, but lead to master it step by step.

That starts with the claim "The Purely Functional Linux Distribution" and ends with multiple huge manuals that are not really accessible.

Maybe even a 5 minute video that explains why one should use NixOS / Nix.

garbas:
This was now partially done with #425 and more updates are coming.

PR #277, opened 2019-05-06
Reviews the nixos.org main page

samueldr:
## Elevator pitch

> The nixos.org main page virtually traps the user into the Linux Distribution section of the site. This PR intends to make the different moving parts of NixOS more accessible at first glance.

* * *

## What does it look like?

> image

### New landing page

The new landing page shows both main components in big visible boxes. The order (left to right) defines their order in narrow mobile layout.

The description was lifted from their respective index pages.

The old nixos.org landing page contents is now the index page of the NixOS sub-section of the site.

### New "main" menu

When not inside a project-specific sub-section of the site, the menu will show quick access to the main components of the project. Maybe NixOps should be added to it, that's not my call, but yours.

### Misc. CSS fixes

These are minor, could be probably should have been handled separately. Misc things that were not obvious, but once seen are annoyances.

## What's left to do?

* [ ] Main menu items.

* [ ] Define the fifth box.

There are two ways to build that box. The first, shown in the current state, and in the screenshot, is to have a small sentence (to be written) heading a list of links to more parts. The second is to describe "The Organization", and link to a new page with more details about how all the parts fit, inspired by the way the [Nix Ecosystem](https://nixos.wiki/wiki/Nix_Ecosystem) page of the NixOS wiki is written.

> What's "The Organization"? Maybe I'm mistaken and it should be named "The Foundation", but from what I can gather, *The Foundation* enables *The Organization* to exist. *The Organization* could include projects not supported by *The Foundation*. I'm probably overthinking this. Additionally, this would help separate the identity of <i>NixOS<b>.org</b></i>, the GitHub organization and *NixOS the distribution*.

## Future work

This is related #248 but does not address it in full. To be addressed, a new Nix landing page would be needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,12 +13,15 @@
 /nixos/nix-pills/
 /nixos/nix-pills-raw
 /nixos/options.json.gz
+/nixos/options.json
 /nixos/screenshots/nixos-*-small.png
 
 /nixpkgs/manual/
 /nixpkgs/manual-raw
 /nixpkgs/packages-unstable.json.gz
 /nixpkgs/packages.json.gz
+/nixpkgs/packages-unstable.json
+/nixpkgs/packages.json
 
 /blogs.json
 /blogs.xml
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,8 @@ default: all
 HTML = index.html news.html \
   nix/index.html nix/about.html nix/download.html \
   nixpkgs/index.html nixpkgs/download.html \
-  nixos/about.html nixos/download.html nixos/support.html nixos/community.html nixos/packages.html nixos/options.html \
+  nixos/index.html nixos/about.html nixos/download.html nixos/support.html \
+  nixos/community.html nixos/packages.html nixos/options.html \
   nixos/security.html nixos/foundation.html \
   nixos/wiki.html \
   patchelf.html hydra/index.html \
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -283,12 +283,12 @@ div.search-details td.nix {
 }
 
 #search-results .description.docbook p > .docbook-important {
-	margin-top: 0.5rem;
+    margin-top: 0.5rem;
 }
 
 #search-results .description.docbook .docbook-important {
-	border-left: #F89406 solid 0.3rem;
-	padding-left: 0.5rem;
+    border-left: #F89406 solid 0.3rem;
+    padding-left: 0.5rem;
 }
 
 /* Manual. */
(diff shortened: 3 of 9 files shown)


edolstra:
Looks good. A few remarks:

* The link to the foundation could be replaced by a link to the Community page (which mentions the foundation).

* We can probably drop Hydra and maybe NixOps from the front page, since they're really only useful if you're already into the Nix/NixOS ecosystem.

* Maybe drop "NixOS Nixpkgs Nix" from the navbar, since they're kind of redundant with the content of the page? Or even drop the entire navbar?

samueldr:
Thanks for the review.

@edolstra If I understand right, the content would be:

<table width="100%">
<tr width="100%">
<td width=50%>Nix</td>
<td width=50%>NixOS                               
</td>
</tr>
</table>
<table width="100%">
<tr>
<td width=33%>Nixpkgs</td>
<td width=33%>??</td>
<td width=33%>The community</td>
</tr>
</table>

The three-then-two arrangement works really well visually, so it'd be a shame to lose it. If we drop NixOps, we're likely to have a hole in the bottom row.

Additionally, NixOps *already is* on the landing page for NixOS, so it might have some merit to stay there, at least until something else more important is found to replace it on the landing page.

As far as "The Community" section would go, I guess it would hold two links, one to the "NixOS Community" page, and the other to "NixOS support", though I am unsure of what to write as a quick one or two sentences blurb to put in front of the links. If anyone has suggestions. (I'll think about it too.)

* * *

Dropping the navbar entirely would trap a user knowing to access sub-sections of the site using the logo drop-down. Even though they're likely to know the full address for e.g. NixOps, they could likely be used to click the logo, then NixOps.

As far as dropping the contents of the navbar on the landing page. Yeah, that's fine, but it will look a bit empty, unless it ends up holding links to non project-specific pages like News or others. Though, losing the quick access links is not a big deal imo, and would ensure we can use the space for other links when we need to.

edolstra:
Yeah, I agree two-then-three (I assume you meant that) works better. So that's an argument for keeping NixOps on the front page :-)

Indeed I was talking about the navbar on the landing page.

grahamc:
I think keeping NixOps on the homepage is valuable for the "We're a
complete solution" story, too.

samueldr:
Updated the last new box:


> image

I'm not sure about the text: please provide something better if you want and can!

Other than that, this is ready.

davidak:
Very good work!

In the NixOS section, do we need "Nix" to be a link? I think we have it already on the left.

I think we should articulate "many other advantages" or delete the words. More informations are provided after clicking the more button.

NixOps section: "configura-tion"

Community section: Maybe first declare that "This is a community effort"

The texts can be improved more, but that is a topic for another PR.

edolstra:
Thanks @samueldr!

PR #281, opened 2019-05-26
Mac OS X -> macOS

sanmai-NL:


vcunat:
BTW, I've also heard of usage on (Free)BSD and some Solaris variant, but I don't expect these and others are significant enough ATM to be mentioned here (in terms of nix usage and support).

PR #286, opened 2019-06-27
Replace list of old releases with link to GitHub

davidak:
it was incomplete and this link is even broken

https://hydra.nixos.org/release/nixpkgs/nixpkgs-0.13/

(I just edited it on GitHub and havn't build it.)

### DIFF
--- a/nixpkgs/download.tt
+++ b/nixpkgs/download.tt
@@ -44,27 +44,7 @@ continuous build system:</p>
 
 <section><h2>Old releases</h2>
 
-<p>The following old releases of Nixpkgs are available:</p>
-
-<div class="row">
-<table class="table span3">
-  <thead>
-    <tr><th>Date</th><th>Version</th></tr>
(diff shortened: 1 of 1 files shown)


nixpkgs/download.tt
```diff
@@ -44,27 +44,7 @@ continuous build system:</p>
 
 <section><h2>Old releases</h2>
```
grahamc: Ok, all releases is good, but what about the old releases? =) (header update?)

davidak: I just wanted to fix this issue, but the complete page needs an overhaul :O

Like what about stable releases and different channels?

Can we just merge this for now and improve the page later. Just renaming the heading might lead people to think this is the right way to use it instead of channels.

edolstra:
IMHO the entire "Old releases" section" can be dropped, since they're all tagged in Git anyway.

PR #288, opened 2019-06-30
Add option to automatically rebuild with entr

egasimus:
Adding the [Event Notify Task Runner](http://eradman.com/entrproject/) to the build environment makes development faster by allowing the site to automatically rebuild in the background upon changing the source files.


egasimus:
This tiny addition is a stepping stone to further contributions planned. You people are doing great work and I'm looking forward to helping out!

flake.nix
```diff
@@ -107,6 +108,20 @@
...
+          export NIXPKGS_MANUAL_IN="${nixpkgsStable.htmlDocs.nixpkgsManual}"
+          export NIXPKGS_STABLE="${packages.x86_64-linux.stablePackagesList}"
+          export NIXPKGS_UNSTABLE="${packages.x86_64-linux.unstablePackagesList}"
+          export NIXOS_OPTIONS="${packages.x86_64-linux.nixosOptions}/share/doc/nixos/options.json"
+          export NIXOS_AMIS="${packages.x86_64-linux.nixosAmis}"
+          export NIXOS_GCE="${packages.x86_64-linux.nixosGCE}"
+          export NIXOS_AZURE_BLOBS="${packages.x86_64-linux.nixosAzureBlobs}"
+          export PACKAGES_EXPLORER="${packages.x86_64-linux.packagesExplorer}/bundle.js"
+          export NIX_PILLS_MANUAL_IN="${packages.x86_64-linux.nixPills}/share/doc/nix-pills"
+        '';
```
edolstra: This is a lot of code duplication, is that really necessary?

edolstra:
Do we actually still need this? It only takes a few seconds to build the homepage, so adding a dependency that we generally don't need *and* duplicating `makeFlags` in `shellHook` is a bit suboptimal.

PR #300, opened 2019-08-29
[WIP] Adds unfree software (hidden by default)

samueldr:
*Depends on #209 being merged.*

Adds back the unfree software listing.

Here's a reminder of my reasoning:

> ### About *unfree* software
> 
> The [project's view about unfree software is known](https://github.com/NixOS/nixos-homepage/pull/120#issuecomment-268485650). We shouldn't *promote* unfree software.
> 
> This rewrite of the packages-explorer was written in a way that is easy to extend, and any part can easily be removed or tweaked as desired, including the filter for unfree software (which defaults to hide them). The cost of adding such a filter is much reduced compared to the previous codebase.
> 
> Still, my personal opinion is that it is a usability issue *not* to list unfree software in some way. Though, I do agree that we shouldn't promote unfree software, which is why they are gated under a toggle, and phrasing is added to every unfree software entry. The tone of that note was [initially more neutral-to-negative](https://github.com/NixOS/nixos-homepage/pull/209/commits/5da6bf3d76779f2ec93494fc5a9cc396392c7603) by design.
> 
> Seeing that the foundation [depends on unfree software](https://search.nix.gsc.io/?q=allowUnfree&i=nope&files=&repos=NixOS-nixos-org-configurations), it seems weird to me to completely hide the presence of unfree software from the packages set, acting as if it didn't exist.
> 
> * * *
> 
> As a closing note, more details about my view. A part of Freedom, is the Freedom of choice, including the choice of using unfree software. We should better educate the user so they make an *informed decision* about their choice of using unfree software, and we can make it obvious what the position of the project is. Hiding the fact that the project has unfree software, while still packaging it is probably unwise, and does reduce usability for the users that have decided that the drawbacks of unfree software is worth it for their needs. Some users probably have skipped this distribution, thinking that they couldn't use some unfree software, possibly a hardware driver.


mmahut:
Looks like the unfree is visible by default now?

edolstra:
Looks like it. That seems accidental.

samueldr:
TLDR from IRC:

```
17:28 <samueldr> https://github.com/NixOS/nixpkgs/commit/ae16dd1a15dfbe29c6a1de61c4cc23e2cd6487f0 is the cause of the issue
17:29 <samueldr> https://github.com/NixOS/nixpkgs/commit/50148f06304ba1dce9130f9476a73fc2d955e10e being when the workaround(?) was introduced
17:51 <samueldr> hmmm, while the "cause" of the issue has been identifier, further changes seem to have changed the behaviours such as simply reverting the change is not enough
17:51 <samueldr> has been identified* 
```
— https://logs.nix.samueldr.com/nixos-dev/2019-11-28#1574962120-1574963468;

* * *

The situation is a bit more complex than intuition led me to believe.

First of all, this is verifiable on the commits before the inclusion of the new package manager.

[This is the nixos-homepage commit](https://github.com/NixOS/nixos-homepage/commit/118786ee22867254f3eaaf5d3e11295d69b27429) introducing the issue. It's the *benign* upgrade to 19.03.

**The issue is actually in Nixpkgs.**

Bisecting, I found the commit introducing the issue:

 * https://github.com/NixOS/nixpkgs/commit/ae16dd1a15dfbe29c6a1de61c4cc23e2cd6487f0

> This is what was used to validate:
> 
> ```
> env -i nix-env -f '<nixpkgs>' -I "nixpkgs=$PWD" -qa --arg config '{}' | grep '^google-chrome-[0-9]'
> ```
> 
>  * `env -i` to reduce environmental impurities like nix checking config.nix in the HOME., or `NIXPKGS_ALLOW_UNFREE`
>  * The `nix-env` invocation is basically a dumbed down invocation similar to the one in the Makefile.
>  * Grepping for a known pattern that is available since way back when.

Though, reverting this change (or re-creating it) does not solve the problem; it seems there are other commits in at least two inter-twined files that are relevant.

 * `pkgs/stdenv/generic/make-derivation.nix`
 * `pkgs/stdenv/generic/check-meta.nix`

I do not know if [this one can be an issue](https://github.com/NixOS/nixpkgs/commit/817c933878d1eea16af8a0efcb62fff59c698cdb). I haven't had the time to dedicate to better dissect the actual issue through the layers of history that have been building on top of this initial breaking change.

### Thoughts

(Though this is yet unsolved)

 * We may want to add a test, I don't know how, that validates that `nix-env` will not list unfree software.
 * Filtering the `nix-env` listing should probably not rely on the `assert` hack/workaround as it did.

edolstra:
@samueldr Ouch, we should definitely revert/fix that ASAP and backport to 19.03/19.09.

For a moment I was worried that this meant Hydra was building/distributing unfree software, but that appears not to be the case.

garbas:
@samueldr is this something we still need to finish? or we gave up?

samueldr:
> @samueldr is this something we still need to finish? or we gave up?

Two things here might be conflated together.

First, the soul of this PR. This PR was split from the main #209 PR. It adds the machinery required to index unfree packages, and the machinery needed to choose between displaying them or not. Simply fixing the conflicts is required here if we want to add this feature to the site. Though note that @edolstra had shown reluctance in indexing unfree packages. This is why it was split off.

Now, there is an added issue, where the way Nixpkgs filters unstable package is not broken. This is https://github.com/NixOS/nixpkgs/issues/74622. This is not something that can be trivially fixed in nixos-homepage, and probably shouldn't.

Though, with that said, if we merge this feature to the packages manager, we can regain control of showing unfree packages on the site.

garbas:
I'm closing this PR since we will rethink how to do this in https://github.com/NixOS/nixos-search

PR #303, opened 2019-09-09
Update Disnix version to 0.9

svanderburg:


disnix/download.tt
```diff
@@ -20,12 +20,12 @@ system of choice.
...
 
-<p>The latest version of Disnix is <strong>0.8</strong>,
-released 25 February 2018.</p>
+<p>The latest version of Disnix is <strong>0.9</strong>,
+released 9 September 2019.</p>
 
 <ul>
   <li><a
-  href="https://hydra.nixos.org/release/disnix/disnix-0.8"><strong>Download
+  href="https://hydra.nixos.org/release/disnix/disnix-0.9"><strong>Download
```
samueldr: [There is no `disnix-0.9` release](https://hydra.nixos.org/project/disnix#tabs-releases).

(There wasn't one for 0.8 either!)

Does someone need to do something for this to become a release on the Hydra instance?

edolstra: The whole Hydra "release" feature is deprecated. Better to just copy the tarballs to some stable location (like https://nixos.org/releases/).

svanderburg: Hmm, I don't know how to copy the tarballs to that location, and I believe I also don't have the rights to do this.

Alternatively, I can also refer to the release artifacts for the corresponding GitHub releases. This is what I also use to build from in Nixpkgs.

garbas:
closing due to https://github.com/NixOS/nixos-homepage/pull/350

PR #305, opened 2019-09-11
Enforce HTTPS and TLSv1.2 for curl | sh

yegortimoshenko:
This is similar to: https://github.com/rust-lang/rustup.rs/pull/1716 (https://rustup.rs)

The intent is to protect user against TLS downgrade attacks and enforce HTTPS on the client.


yegortimoshenko:
download (14)

nix/index.tt
```diff
@@ -26,7 +26,7 @@
     Nix</strong> is to run the following shell command (as a user
     other than <tt>root</tt>):</p>
 
-    <pre><span class="nix-command">curl https://nixos.org/nix/install | sh</span></pre>
+    <pre><span class="nix-command">curl --proto '=https' --tlsv1.2 <nobr>https://nixos.org/nix/install</nobr> | sh</span></pre>
```
lightdiscord: The `<nobr>` element seems [obsolete](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nobr).

PR #313, opened 2019-10-14
add manual link in nixos page

danielstaleiny:


vcunat:
Hmm, I'd personally expect that from this short page people click "many other advantages" to [a longer one](https://nixos.org/nixos/about.html) which could perhaps see some improvements including links to _all three_ relevant manuals (nix, nixos, nixpkgs).

danielstaleiny:
Perhaps it could be in the menu like it is under nix. My motivation is that is easier to search in google for nixos manual than it is to navigate it from home page of nixos. 

image

Menu bar should have Manual somewhere like nix page has. 

image

garbas:
there is "Learn" link in the toolbar

vcunat:
Right, that one was apparently added in the meantime.

PR #314, opened 2019-10-16
Add Manual link to menu bar. as layout

danielstaleiny:
alternative to #313


edolstra:
Hm, I think we discussed this in the past and the conclusion was that the navbar is already too overcrowded. Though maybe we can drop the "Security" link.

garbas:
there is "Learn" link in the toolbar

PR #318, opened 2019-11-03
Rename Support to Learn

asymmetric:
## Motivation

Whenever I'm looking for help or documentation, I find it difficult to find it under the word *Support*. 

I find *Learn* to be much clearer.

The idea is that if someone's at the NixOS homepage and looking for help, they're likely newbies (we all have the manuals in our browser histories already I presume).
Someone like that, in my opinion, wants to *learn* what Nix/NixOS is all about, rather than "receive support".

## Alternatives

*Documentation* would be another candidate, but we also point to Discourse/IRC rather than just the manual, meaning it's not a good fit.

*Help* is another good option, and I'm honestly not sure which is better between that and *learn*.

## Inspiration

This is inspired by Rust's excellent [home page](https://www.rust-lang.org/), and by discussions at the Documentation hackathon at NixCon 2019.

/cc @zimbatm @chreekat @layus @azazel75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ default: all
 HTML = index.html news.html \
   nix/index.html nix/about.html nix/download.html \
   nixpkgs/index.html nixpkgs/download.html \
-  nixos/index.html nixos/about.html nixos/download.html nixos/support.html \
+  nixos/index.html nixos/about.html nixos/download.html nixos/learn.html \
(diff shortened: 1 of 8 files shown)


chreekat:
Excellent!

azazel75:
well done!

chreekat:
I would like to see this PR merged. Who has the power to do so?

PR #319, opened 2019-11-18
Mention 'nix search' on search page

raboof:


edolstra:
`nix` is an experimental command (it even needs to be enabled explicitly using a flag in Nix 2.4) so we probably shouldn't advertise it on the website.

raboof:
oh, sorry, it was recommended to me on IRC and I really like it ;). What would be the non-experimental alternative?

edolstra:
`nix-env -qa`

raboof:
meh, that's way less convenient... nevermind I guess.

PR #324, opened 2019-12-28
Show how to get the interpreter from the gcc wrapper for patchelf

raboof:


grahamc:
Since this is documentation specific to Nixpkgs, maybe it should be documented in Nixpkgs? patchelf is used a lot outside of the Nix ecosystem, too.

zimbatm:
Actually if you read the whole page it makes sense to have that example there. The page start with:

>  And in the Nix Packages collection we build our own Glibc, which we obviously want to use. Building programs with a different interpreter is a matter of using the linker flag -dynamic-linker:
and then

> However, this doesn’t work with third-party binaries (such as Adobe Reader in Nixpkgs). This is where PatchELF comes in. You can simply rewrite the executable:

But there is no actual example of how we would do it for nixpkgs.

The wording could be a bit better but at least it ties up these references to nixpkgs.

edolstra:
I think we should remove this page and put any docs in the README.md of the patchelf repo.

zimbatm:
Probably. The patchelf project needs some more love overall.

PR #327, opened 2020-02-25
Add flake.nix

edolstra:
This adds a `flake.nix` to the repo, ensuring a reproducible build of the homepage. Previously the homepage used a Makefile that was downloading stuff from the Internet and calling nix-build, so we couldn't build the homepage inside a Nix build.

It also adds a NixOS container configuration, so you run 
```
$ nixos-container create homepage --flake nixos-homepage
$ nixos-container start homepage
```
to get a running version of the homepage.

Some dynamically fetched files (e.g. `blogs.xml`) are now part of the repository. We probably should add a script that updates these files and commits them.

For people who don't have a flake-enabled version of Nix, there is a `default.nix` and `shell.nix` that uses https://github.com/edolstra/flake-compat.

CC @garbas


nixos/amis.nix
```diff
@@ -0,0 +1,295 @@
+let self = {
```
garbas: couldn't we also get this file from `<nixpkgs/nixos/modules/virtualisation/ec2-amis.nix>` or `${pkgs.path}/nixos/modules/virtualisation/ec2-amis.nix`?

nixos-release.tt
```diff
@@ -0,0 +1 @@
+[%- latestNixOSSeries = "19.03"; latestNixOSRelease = "19.03.172883.1601f559e89" -%]
```
samueldr: Isn't it currently 19.09? And how does this get updated?

Makefile
```diff
@@ -221,46 +192,34 @@ endif
...
+
+NIXOS_OPTIONS = /no-such-path
+
+.PHONY: nixos/options.json
 
 # Cute hack, this allows future expansion if desired
 # Mainly, this allows tracking NIXOS_SERIES
 nixpkgs/packages-channels.json: Makefile
-	echo '["nixos-$(NIXOS_SERIES)", "nixos-unstable", "nixpkgs-unstable"]' > $@
+	echo '["nixos-$(NIXOS_SERIES)", "nixpkgs-unstable"]' > $@
```
samueldr: Lately, we have had consistently fresher `nixos-unstable`, compared to `nixpkgs-unstable`. Should we prefer the channel that updates more regularly?

samueldr:
# blogs.xml

This is something that we control. Since we need to make a static and reproducible page, we don't have much options.

 1. Update the data in the repo
 1. Remove the section from the site
 1. Augment the site at run-time

### Update the data in the repo

The first option is what is already proposed. Though I feel, and I guess others would tend to agree, that this is a dirty solution. The main advantage here is that it would continue providing the same level of service from the website, that is, the homepage shows the blogs listing even without JavaScript and AJAX.

### Remove the section from the site

Removing the whole section from the homepage *is* an option, though I feel it is a net negative. Replacing with a simple link to the planet's homepage reduces eyeballs on the contents of the community blogs.

### Augment the site at run-time

The last option is meeting halfway. The `blogs.xml` pre-baked rendering is removed. We leave a space with a proper JavaScript-less fallback inviting the user to visit the planet site. On load, the RSS feed from the planet is loaded via AJAX to fill in the contents.

This does mean that we lose the contents listing for users that load the page without JavaScript, but we also remove the need for the (albeit not that much) added complexity to keep `blogs.xml` updated in the repo. Since the homepage twitter feed *already* follows that convention, I feel this compromise is sensible.

Though, a real drawback here, this solution *does* require a bit of work to implement. Luckily, this is simple enough to implement, and *does not* require a complex setup like the packages listing, this can be done through trivial hand-written old-school enough JavaScript.

samueldr:
# nixpkgs-commit*.json

Those are a bit harder than `blogs.xml`. Two main reasons:

 * The information is more volatile.
 * We do not have a local data source to mine.

Really, the fact that we get so much more updates daily, even hourly, makes the option of updating the repo a less palatable one.

Off the top of my head, not much better solutions here. I looked at it quickly for the mobile-nixos site, but had no great solution. (In addition to it being less useful there.)

I think all solutions other than removing the section from the homepage revolve around the GitHub API, and JS augmentation.

* * *

### GitHub API

 * https://api.github.com/repos/NixOS/nixpkgs/commits?sha=master

<details>
<summary>Browser console toy</summary>

    fetch("https://api.github.com/repos/NixOS/nixpkgs/commits?sha=master").then((response) => response.json()).then(console.log)

</details>

This would allow us to get the commits for the master branch, without any infrastructure change. (Dropping `?sha=master` would show all new commits, regardless of branch.)

A hefty drawback here is that we're transmitting data to a third-party, and I don't know what kind of issues this could cause.

Another issue I can see, but have not researched, is that this may be eating in that user's unauthenticated requests to the GitHub API, and may fail if the user made unauthenticated queries!

### Mirroring the data

This could be solved by having the data mirrored at some location that is not a third-party. The mirror could be refreshed every minute, or even five or ten minutes. It already was refreshing on a timer!

Though this does mean some kind of infrastructure or setup has to handle this. The website-part for fetching and using the data is trivial and a non-issue. The other part is not. I figure a "nixos-homepage-data" repository with the data could be used, with the appropriate cached github raw data file, and some automation.

> Side-note: This solution could also be used to mirror the twitter feed, rather than embed a third-party widget.

nixos-release.tt
garbas: I'm split between manually updating this and getting this from nixpkgs.
Having this generated would of course mean less work for us, just rerunning the build every day. 

Having to manually update this value it would allow us organize releases in public better. 

Example of this would be: Friday master receives a bump of a version, we have some time to finish release announcements and publish them on (eg.) Monday. I can see in the future (I'm hoping for 20.9 we would already be doing this.

garbas:
Hi @samueldr,

I'll try to quickly answer few concerns.

1. blogs.xml

The easier will be to configure to run travis on nightly basis and pull new blogs each time. I will configure it in next PR which will be about pushing to netlify. I think if we would put all this changes in one PR it will be very hard to review.

2. nixpkgs-commit*.json

This section should be removed from front-page altogether. It doesn't bring any value. Those who know where to look for commits they will look on GitHub.

edolstra:
Couple of changes:

* I've removed nixpkgs-commit*.json.

* AMIs / Azure blobs are now obtained from Nixpkgs directly.

* I added an ./update.sh script that runs `nix flake update`, and updates `blogs.xml` and `nixos-release.tt`.

We can get rid of `nixos-release.tt` once 20.03 is released, because we can get that information directly from the Nixpkgs flake. (Nixpkgs 19.09 does not contain a `flake.nix` so we can't use it here yet. The 19.09 flake is not up to date and doesn't have the same commit hashes.)

For `blogs.xml`, I agree it would be nicest to fetch it at runtime in the browser.

garbas:
Ignore my previous comment. I wasn't aware of `nixFlakes`.

I'll slowly get up to speed. :)

PR #329, opened 2020-03-02
Update packages-explorer to use relative paths.

egasimus:
To save some seconds when doing the occasional lookup, I self-host a mirror of the NixOS `packages` and `options` pages. When serving from a subfolder, e.g. `http://localhost/nixos-homepage/nixos/packages.html`, the usage of absolute paths trips up the data fetching. Changing the paths to relative fixes that.


samueldr:
Thanks for the improvement!

PR #330, opened 2020-03-05
Deploy to Netlify

garbas:
This PR:
- uses Github Actions to build nixos-homepage
- deploys to Netlify (either as preview or to "production" if you commit to master branch)
- comments on PR or commit with a link to a preview the website
- 301 redirects we configured for things that should be redirected (look at netlify.toml file)

I already went ahead and configured secrets needed for this to work in NixOS/nixos-homepage repository. Secrets that I configures are:
- GH_TOKEN (to be able to create comments, only needs public:repo scopes)
- NETLIFY_AUTH_TOKEN (to authenticate to Netlify)
- NETLIFY_SITE_ID (to select the site to which to deploy to)

For now I generated github and netlify token from my own account. I wonder if we should create a separate accounts only for this (eg. nixos-homepage-bot on github and nixos-homepage-webmaster on netlify).

You can find preview links in the PR history.

/cc @samueldr @edolstra


.github/workflows/main.yml
```diff
@@ -0,0 +1,29 @@
...
+      run: nix-build && mkdir build && cp -RL ./result/* ./build/
+    - name: Deploy to Netlify
+      uses: nwtgck/actions-netlify@v1.0.5
+      env:
+        NETLIFY_AUTH_TOKEN: ${{ secrets.NETLIFY_AUTH_TOKEN }}
+        NETLIFY_SITE_ID: ${{ secrets.NETLIFY_SITE_ID }}
+      with:
+        publish-dir: './build'
+        production-branch: 'master'
+        github-token: ${{ secrets.GH_TOKEN }}
```
zimbatm: there is a default token you can use that is able to post comments on the pull request:
```suggestion
        github-token: ${{ secrets.GITHUB_TOKEN }}
```

garbas: I tried to use that and it didn't work ... it got stuck at fetching the token i think.

zimbatm: Was the branch pushed on a forked repo or the nixos-homepage repo?

PR #335, opened 2020-03-11
include nix install script in the website and remove the redirect (fo…

garbas:
…r now)

also:
- use GITHUB_TOKEN instead of custom GH_TOKEN
- don't create comment on cron jobs

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,9 +1,5 @@
-name: "Build & Deploy to Netlify"
+name: "Hourly Build & Deploy to Netlify"
 on:
-  pull_request:
-  push:
-    branches:
-      - master
   schedule:
   - cron: '0 * * * *'
 jobs:
@@ -26,6 +22,6 @@ jobs:
       with:
         publish-dir: './build'
         production-branch: 'master'
-        github-token: ${{ secrets.GITHUB_TOKEN }}
         deploy-message: 'Deploy from GitHub Actions'
       if: github.repository == 'NixOS/nixos-homepage'
+
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - master
-  schedule:
-  - cron: '0 * * * *'
 jobs:
   build-and-deploy:
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


samueldr:
For those out of the loop, and history spelunkers, can the reason for keeping the script as a website artifact, rather than a redirect, be added to either a comment to the Netlify config, or described in the PR body?

I'm saying this since I don't know, and I've been kinda following what's going on.

.github/workflows/cron.yml
```diff
@@ -0,0 +1,27 @@
+name: "Hourly Build & Deploy to Netlify"
```
edolstra: What the difference between this and `.github/workflows/main.yml` (which also has a `cron: '0 * * * *'` line)?

edolstra:
We probably also need to serve `https://nixos.org/releases/nix/nix-<version>/install` directly rather than as a redirect, since there are people who are installing it without `-L`. (We won't need to do that for new releases, just the existing ones.)

edolstra:
@samueldr The reason is that a lot of people are doing `curl https://nixos.org/nix/install` rather than `curl -L ...`, so they break if the file becomes a redirect.

edolstra:
BTW breaking `curl https://nixos.org/nix/install` is not really a big deal since people shouldn't be doing that anyway in scripts (because the version could change in incompatible ways at any time). But the version-specific scripts (`https://nixos.org/releases/nix/nix-<version>/install`) should continue to work.

.github/workflows/cron.yml
garbas: oh sorry, i was suppose to remove cron from `main.yml`. i'll fix it.

garbas:
@edolstra i will create the redirects. but going forward people should rely on https://releases.nixos.org/nix/nix-<version>/install. I will include this into announcement.

.github/workflows/cron.yml
edolstra: I mean what's the difference? They look very similar. Do we need both?

garbas:
oh actually ... above should also just follow the redirect.

.github/workflows/cron.yml
garbas: in cron.yml I also remove `github-token` to not add comments every hour. it makes listening to notifications on this github repo almost useless. this is the main reason for the split.

PR #344, opened 2020-03-15
Redirect hydra and nixops manuals to the latest hydra builds

samueldr:
With the move to netlify, the hydra and nixops manuals vanished, without a redirect. Considering search engines link to those manuals, I believe it's mandatory we keep [cool URIs](https://www.w3.org/Provider/Style/URI.html) working. 

Use a "200" redirect.

> 200: "OK". Redirects with this status code will change the server response without changing the URL in the browser address bar. This is used for rewrites and proxying.

 * https://docs.netlify.com/routing/redirects/redirect-options/#http-status-codes

This will keep the nixos.org/.../manual URL in the address bar, but serve the contents proxied from the hydra result.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -53,3 +53,9 @@
   to = "https://github.com/NixOS/nixops"
   status = 301
   force = true
+
+[[redirects]]
+  from  = "/nixops/manual"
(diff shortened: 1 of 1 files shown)


edolstra:
These should be regular redirects, to proxies, since these pages are not part of the NixOS website and don't have the same styling.

edolstra:
I've changed them to 302.

PR #345, opened 2020-03-15
netlify: Change 301 redirects to 302

adisbladis:
302s are potentially cached indefinitely and it makes it very hard to change anything in the future.

This also changes `/nixops/manual` to a 200 (netlify proxy).

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,19 +1,19 @@
 [[redirects]]
   from = "/channels/*"
   to = "https://channels.nixos.org/:splat"
-  status = 301
+  status = 302
   force = true
 
 [[redirects]]
   from = "/releases/channels/*"
   to = "https://channels.nixos.org/:splat"
-  status = 301
+  status = 302
   force = true
 
 [[redirects]]
   from = "/releases/nixos/channels/*"
   to = "https://channels.nixos.org/nixos/:splat"
-  status = 301
+  status = 302
   force = true
 
 [[redirects]]
@@ -25,7 +25,7 @@
 [[redirects]]
   from = "/releases/*"
   to = "https://releases.nixos.org/:splat"
-  status = 301
+  status = 302
   force = true
 
 [[redirects]]
@@ -33,7 +33,7 @@
   to = "https://releases.nixos.org/nix/nix-2.3.3/install"
   status = 200
   #to = "https://channels.nixos.org/nix-latest/install"
-  #status = 301
+  #status = 302
   force = true
 
 [[redirects]]
@@ -63,5 +63,5 @@
 [[redirects]]
   from  = "/disnix"
   to = "https://github.com/svanderburg/disnix"
-  status = 301
+  status = 302
   force = true


samueldr:
This may conflict with #344. Just saying for when this will be merged.

adisbladis:
@samueldr Should I cherry-pick your changes to this PR then?

samueldr:
As you see fit. Another option would be to remove the Nixops manual redirect from your PR I guess.

A last option is not to do anything and leave that to when it's being merged.

netlify.toml
```diff
@@ -25,25 +25,31 @@
...
 [[redirects]]
   from  = "/nixops"
   to = "https://github.com/NixOS/nixops"
-  status = 301
+  status = 302
+  force = true
+
+[[redirects]]
+  from  = "/nixops"
+  to = "https://github.com/NixOS/nixops/manual"
```
garbas: this is a non working link, you probably mean:

```suggestion
  to = "https://github.com/NixOS/nixops/tree/master/doc/manual"
```

garbas:
Would you like to publish manual in the same manner as nixos homepage (to netlify using github actions). And maybe even get a subdomain for it, eg. nixops.nixos.org or ops.nixos.org.

netlify.toml
```diff
@@ -25,25 +25,31 @@
...
+  status = 200
```
edolstra: We should not proxy /nixops (at least I don't see a reason to do so).

netlify.toml
```diff
@@ -25,25 +25,31 @@
...
 
 [[redirects]]
   from  = "/nixops"
   to = "https://github.com/NixOS/nixops"
-  status = 301
+  status = 302
+  force = true
+
+[[redirects]]
+  from  = "/nixops"
```
edolstra: This conflicts with the previous redirect for /nixops.

netlify.toml
edolstra: The rendered manual is at https://hydra.nixos.org/job/nixops/master/tarball/latest/download-by-type/doc/manual.

garbas:
I don't have strong opinion on this, but semantics say that when you move (permanently) you should use 301 and when this would be a temporary thing then use 302. I think we should continue using 301 when possible.

zimbatm:
Every second that we wait is another client caching the redirect permanently. It's one of the rare thing that has no TTL associated with it. It's better to use 302 by default and only use 301 when you are really really really sure that it's never going to move.

edolstra:
My understanding of the HTTP standard is that 301s are not in fact cacheable permanently, though they can be cached by default using the standard heuristics.

zimbatm:
I guarantee you that Chrome and Firefox will cache a 301 redirect with no expiry date.
Do what you want but I recommend merging this PR and have the long conversation later.

adisbladis:
Indeed browsers tend to treat 301 as a _permanent_ redirect.
Can we merge this PR before we have even more clients ending up with permanent caches?

garbas:
relevant why this needs to be merged. browsers cache 301 "forever" (until they ran out of cache or until cache is cleared)... https://stackoverflow.com/questions/9130422/how-long-do-browsers-cache-http-301s

edolstra:
Well, the whole point it to redirect permanently. Also, if clients have already cached it forever, it's too late to change it back.

samueldr:
Just a note, "permanent" here is not "this is permanently a redirect", but "we promise this new url is its new home forever". This annoying distinction is why 302 is more apt than 301. We can't control some of those URLs if, for any reason, the endpoint needs to be changed. Good luck getting GitHub to honor a redirect outside GitHub, or even inside GitHub depending.

PR #293, opened 2019-08-02
Duckduckgo

zaoqi:

PR #351, opened 2020-03-16
remove search from the toolbar

garbas:
nixos.org is already index in all the search engines and this is what most people use and search now days happens outside the website. This would of-course not be true if our search would be very tightly integrated with nixos.org.

Other websites I checked (kubernets.io, rust-lang.io) out also don't include this kind of search (full text over all sites) and just rely on google to index them.

I do think we "expose" more search over existing packages and nixos options but that would take some time to get there. first step would be to remove this search in the first place.

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -100,10 +100,6 @@
               <ul class="nav pull-right">
             [% END %]
 
-              <form class="navbar-search" method="get" action="https://www.google.com/search">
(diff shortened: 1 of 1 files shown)


garbas:
This also allows us to not have a discussion which search engine to use :) https://github.com/NixOS/nixos-homepage/pull/293

PR #352, opened 2020-03-16
Strip nix store paths from source nixpkgs in packages.json

### LINKED ISSUE #349: Paths for nix expressions have `/nix/store/l7dndanv2qakp6lbpbndgifk13rwf1ch-source/` prepended
Searches on https://nixos.org/nixos/packages.html?channel=nixos-19.09 are listing nix expressions with what appears to be an absolute path from the build process, breaking those links -- eg, it links to https://github.com/NixOS/nixpkgs/blob/54a27fb2fc5fbd8927f9f6dc77fe1de1faa05266//nix/store/l7dndanv2qakp6lbpbndgifk13rwf1ch-source/pkgs/applications/misc/cura/default.nix#L57 instead of https://github.com/NixOS/nixpkgs/blob/54a27fb2fc5fbd8927f9f6dc77fe1de1faa05266/pkgs/applications/misc/cura/default.nix#L57

samueldr:
This removes the accidental inclusion of /nix/store paths in the packages.json file.

This was part of the original expression:

```
| sed "s|$$nixpkgs/||g" | jq -c . > $@.tmp
```

It was kept as-is, while it's not a Makefile anymore.

Fixes #349.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,7 +24,7 @@
         nix-env -f '<nixpkgs>' -I nixpkgs=${src} -qa --json --arg config 'import ${./packages-config.nix}' >> tmp
         echo -n '}' >> tmp
         mkdir $out
-        < tmp sed "s|$$nixpkgs/||g" | jq -c . > $out/packages.json
+        < tmp sed "s|${src}/||g" | jq -c . > $out/packages.json
       '';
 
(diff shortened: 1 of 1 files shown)


samueldr:
The validation was added as a separate commit so removing it is trivial.

PR #357, opened 2020-03-19
Documentation overview page

garbas:
Two things were done in this patch:

1. There is only one Learn page which all learning material on there. It
   used to be the case that each project had (or didn't have) a Learn item
   in the navigation bar. In some cases the item in the navigation bar was
   called Manual which broke the consistency of the website and most
   importantly the expectations of the visitor.

2. An opinionated overview of the content of the manuals was created. I
   tried to avoid to much work and instead of changing the manuals I
   created a table of content which I believe newcomers will be more
   interested in. I'm not saying this is how table of content should
   look like, but it is a start we can build upon. Most importantly is
   that we keep in mind that target audience for this page are newcomers
   and visitors not that familiar with nix ecosystem.


rail:
I really like that I can find all the entry points easier than before. \o/

For some reason, when I want to get to the NixOS manual, I click on the "Download" button, then on the "Next Step: Manual". Not ideal :)

What do you think about making the listings shallow and limit to 1 level only? IMHO, it would reduce the noise and avoid the paradox of choice.

Also, not sure if it was discussed or not, I would expect the order of sections to be NixOS, Nix, NixPkgs.

Thank you for the effort!

davidak:
>For some reason, when I want to get to the NixOS manual, I click on the "Download" button, then on the "Next Step: Manual". Not ideal :)

Yes, that's also the easiest way i know. :grimacing:

learn.tt
```diff
@@ -0,0 +1,146 @@
...
+          ...
+      </li>
+      <li><a href="[%root%]/nixpkgs/manual/#chap-submitting-changes">Contributing to Nixpkgs</a></li>
+    </ul>
+    <a href="[%root%]/nixpkgs/manual" class="btn btn-block">More ...</a>
+  </div>
+
+  <div class="span4">
+    <h2>Learn NixOS</h2>
+    <p>How to install, use and extend NixOS, a Linux distribution based on Nix package manager.</p>
```
davidak: here should be "NixOS is ..." to be consistent with other sections

learn.tt
```diff
@@ -0,0 +1,146 @@
...
+      <li><a href="https://nix-cookbook.readthedocs.io">Nix cookbook</a> — Nix Community Cookbook presents topical, practical ways of using Nix package manager ecosystem.</li>
+    </ul>
+  </div>
+
+  <div class="span6">
+    <h2>Get help from the community</h2>
+    <ul>
+      <li><a href="https://discourse.nixos.org/">Discourse</a> — Discourse is a forum where we have discussions and help users out with their questions.</li>
+      <li><a href="irc://irc.freenode.net/#nixos">IRC</a> — You can drop by on the IRC channel <a href="irc://irc.freenode.net/#nixos"><strong><tt>#nixos</tt></strong></a> on <a href="https://freenode.net/"><tt>irc.freenode.net</tt></a> to chat with the developers or other users.</li>
+      <li><a href="https://stackoverflow.com/questions/tagged/nixos">Stackoverflow</a> — a place to ask questions.</li>
```
davidak: when stackoverflow is listed we might also want to list reddit?

https://www.reddit.com/r/NixOS/

garbas: :+1:

garbas:
> For some reason, when I want to get to the NixOS manual, I click on the "Download" button, then on the "Next Step: Manual". Not ideal :)
> 

Download page(s) will be done in some other PR. This PR is only about `/learn.html`. But you are right we need to fix download pages as well :smile: 

> What do you think about making the listings shallow and limit to 1 level only? IMHO, it would reduce the noise and avoid the paradox of choice.
> 

For now I think this is enough. I hope we can actually improve the table of content of all the manuals and organize them a bit better. I think topics that I exposed here might receive their own tutorial at some point and then this wont be an issue anymore.

> Also, not sure if it was discussed or not, I would expect the order of sections to be NixOS, Nix, NixPkgs.

We will have a discussion about this next meeting. What to put first, what second, what to emphasize,etc...

garbas:
In spirit that this already improved the current state I'm merging this. If you think we can even further improve this page, please send your PR.

PR #371, opened 2020-03-28
Handle null values in attrsets option examples

### LINKED ISSUE #370: options: clicking on environment.shellAliases does not show documentation
On this page: https://nixos.org/nixos/options.html#shellaliases

clicking on `programs.bash.shellAliases` works fine (also for other options) while clicking on `environment.shellAliases` does nothing.

The wrong behaviour trigger a `TypeError: can't convert null to object` in the development console. The stacktrace on chromium (same happens on firefox anyway):
```
Uncaught TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at needIndentation (options.html:256)
    at options.html:266
    at Array.some (<anonymous>)
    at ppNix (options.html:306)
    at HTMLTableRowElement.showOption (options.html:416)
    at HTMLTableRowElement.dispatch (jquery.min.js:2)
    at HTMLTableRowElement.u (jquery.min.js:2)
```

zarelit:
Fix #370

### DIFF
--- a/nixos/options.tt
+++ b/nixos/options.tt
@@ -168,6 +168,7 @@ function ppNix(indent, v) {
     return len <= i ? ' ' : '\n';
   }
   function needIndentation(v) {
+    if (v == null) return false;
     if (typeof v == "object") {
       if (Array.isArray(v)) {
         if (v.some(needIndentation))


zarelit:
@samueldr @garbas The issue arise only for options that have all of these together:

- non literal example
- example is an attrset
- the value of a key is null

the patch just adds this edge case to the `needIndentation` function. I've built and tested the patch offline

garbas:
@zarelit thx!

PR #377, opened 2020-04-02
redirect /nix-dev to new location

garbas:


netlify.toml
```diff
@@ -10,6 +10,12 @@
   status = 302
   force = true
 
+[[redirects]]
+  from = "/nix-dev/*"
+  to = "https://releases.nixos.org/nix-dev/:splat"
```
garbas, line 15: @edolstra In this PR I'm assuming that I will upload nix-dev to nix-releases bucket. Would it be better to create new bucket for this eg. old.nixos.org or archive.nixos.org?

samueldr: If it is in another bucket, I still strongly suggest suffixing with `/nix-dev/:splat` so we can archive more things as required.

edolstra: Yeah let's use releases.nixos.org. Otherwise we'll have to set up another bucket + Cloudfront distribution + DNS.

garbas: consider it done

PR #380, opened 2020-04-03
redirect old urls which are found by analytics

garbas:


Makefile
```diff
@@ -61,14 +61,22 @@ $(NIXPKGS_MANUAL_OUT): $(NIXPKGS_MANUAL_IN) bootstrapify-docbook.sh bootstrapify
...
 
-all: $(HTML) favicon.png $(subst .png,-small.png,$(filter-out %-small.png,$(wildcard nixos/screenshots/*))) \
+all: $(HTML) favicon.png favicon.ico robots.txt $(subst .png,-small.png,$(filter-out %-small.png,$(wildcard nixos/screenshots/*))) \
   nixos/packages-explorer.js \
   nixpkgs/packages-channels.json
 
 
+robots.txt: $(HTML)
+	echo "Users-agent: *" >> $@
+	echo "Disallow: /" >> $@
```
grahamc, line 71: This seems not optimal. Why?

garbas: Disallow everything first and then allow things we generate. How would you do it otherwise?

garbas: Another solution would be to just have `Users-agent: *`

grahamc: I guess I'm worried disallowing things that don't exist would just simply make 404's stop showing up from search engines, instead of showing up in a way we could fix them.

garbas: That is actually a very good point. I'll commend the `Disallow` and `Allow` lines for now and let search engines catch everything.

PR #381, opened 2020-04-04
migrate planet to nixos.org/planet

### LINKED ISSUE #336: planet.nixos.org
After moving to Netlify, we will shut down the EC2 instance that currently serves nixos.org but also planet.nixos.org. So we need to figure out what to do with it (assuming we want to keep it). Maybe the contents can be generated/committed as part of the update.sh script in nixos-homepage and served by Netlify?

garbas:
fixes #336

this is here just a proof of concept, questions now are:
-  should we have planet under subdomain (planet.nixos.org) but still keep the source in this repo. we would be pushing planet html to new netlify page.
- move code out into separate repository and deploy it to different netlify page.
- leave it like it is now and have code in nixos-homepage and also under nixos.org/planet.

/cc @edolstra


garbas:
i moved the planet code to NixOS/nixos-planet repository

PR #386, opened 2020-04-11
Download page refresh

garbas:

PR #387, opened 2020-04-11
Landing page

garbas:
this pr is based on top of #386


garbas:
This is not done.

layout.tt
```diff
@@ -40,76 +40,39 @@
...
+          [% IF !hideSearch %]
+          <form class="navbar-search pull-right">
+            <div class="input-append">
+              <input class="span2" id="appendedDropdownButton" type="text" />
+              <div class="btn-group">
+                <button class="btn dropdown-toggle" data-toggle="dropdown">
+                  Search
+                  <span class="caret"></span>
+                </button>
+                <ul class="dropdown-menu pull-right">
```
davidak, line 62: i would assume a search on the website there

nixos/options.tt
```diff
@@ -1,4 +1,4 @@
-[% WRAPPER layout.tt title='Search NixOS options' menu='nixos' %]
+[% WRAPPER layout.tt title='Search NixOS options' menu='nixos' hideSearch=1 %]
```
davidak, line 1: i think it makes sense to leave the search there for consistency

garbas: I agree, we could add to search on google by default (`https://www.google.com/search?q=site%3Anixos.org+nix&oq=site%3Anixos.org+<QUERY>`). and then also have an option to search for packages and option in the dropdown as it is now. 

I hope to change this soon once we remove current implementation of packages/options search.

davidak: I would consider leading visitors to Google controversial. Then we could directly use Google Analytics.

I personally use startpage.com for web search and it's quiet popular. It is just a proxy for Google and you should get the same results without personalization and location based ones. Search URL is `https://www.startpage.com/do/dsearch?query=site%3Anixos.org+nix`.

Screenshot from 2020-04-22 20-09-11

But i don't know if we can agree on that.

Another option is local search using JS like [Tipue](https://tipue.com/search/) or [lunrjs](https://lunrjs.com/). I use the first on my website which has 403 pages (number of pages, not HTTP code :D). It works great. The NixOS website has probably less pages after the redesign, or should it also include the manuals?

I think the search for packages/options should have a different place. A separate page was fine. I always type "n" in my browser and select the page for the packages or options search.

index.tt
```diff
@@ -6,152 +6,54 @@
...
-      <a href="/news-rss.xml">
-        <img src="/logo/rss.png"/>
-      </a>
-      News
-    </h2>
-    [% INSERT "latest-news.xhtml" %]
-    <p><a href="news.html" class="btn btn-mini">More »</a></p>
+    <h1>Reproducible builds and deployments.</h1>
+    <p class="lead">
+      NixOS is an open source set of tools that help you package and deliver
```
davidak, line 14: i don't know if i should already comment this, but here are my spontaneous thoughts

>NixOS is an open source set of tools

that's not a very specific description. you could say that about many projects that are totally unrelated. i think "Linux distribution" should be in there. we are already way ahead of what people know, give them something they can grasp

garbas: :+1: agree I'm so not happy with this description. But at least we have a titie :)

davidak: yes, that's something :sparkles:

index.tt
```diff
@@ -6,152 +6,349 @@
...
-          <i class="fa fa-chevron-right"></i> Learn more about Nix
-        </a>
+<div class="row-fluid">
+
+  <div class="span6">
+    <h1>Reproducible builds and deployments.</h1>
+    <p class="lead">
+      NixOS, an open source software, gives you tools to share
+      your development and build environments between machines with confidence.
+      Many know NixOS as the <strong>declarative</strong> Linux distribution
```
davidak: >Many know

could be understood as if it were not true. why not "NixOS is configured declaratively"? yes declarative would be nicer as bold word

you are in the right direction!

index.tt
```diff
@@ -6,152 +6,349 @@
...
-      <div class="learn-more">
-        <a class="btn btn-success" href="[%root%]nix/">
-          <i class="fa fa-chevron-right"></i> Learn more about Nix
-        </a>
+<div class="row-fluid">
+
+  <div class="span6">
+    <h1>Reproducible builds and deployments.</h1>
+    <p class="lead">
+      NixOS, an open source software, gives you tools to share
```
davidak, line 14: isn't that Nix? and NixOS builds upon this features?

index.tt
```diff
@@ -6,152 +6,349 @@
...
+      your development and build environments between machines with confidence.
+      Many know NixOS as the <strong>declarative</strong> Linux distribution
+      that is not afraid to be different (XXX: too bullshit argument).
```
davidak: "different" tells not much. do we have a flashy wallpaper or does the mouse move in opposite direction or what?

i understand you want to say "is not afraid to break with conventions (link to site containing FHS) to achive innovative features"

index.tt
```diff
@@ -13,7 +13,7 @@
     <p class="lead">
       NixOS, an open source software, gives you tools to share
       your development and build environments between machines with confidence.
-      NixOS GNU/Linux distribution is built for the cloud, that brings
+      NixOS GNU/Linux distribution is built for the cloud and brings
       declarative configuration to the next level.
```
davidak, line 17: i like that. differentiate us from other config. management

index.tt
```diff
@@ -13,7 +13,7 @@
     <p class="lead">
       NixOS, an open source software, gives you tools to share
       your development and build environments between machines with confidence.
-      NixOS GNU/Linux distribution is built for the cloud, that brings
+      NixOS GNU/Linux distribution is built for the cloud and brings
```
davidak, line 16: words words words

is "cloud" really something we want to use?

like "it's just someone elses computer". or do we need some buzzwords to please some people?

index.tt
garbas: closing since i changed it already

index.tt
garbas: Resolving since i changed it already

index.tt
garbas: Agree cloud is weird word, but couldn't come up with something better. We need to change this part.

index.tt
```diff
@@ -6,152 +6,349 @@
...
+            Nix</a>. To setup a more persistent environment you can also write a
+          simple <code>shell.niix</code> file:
+        </p>
+        <pre class="well" style="margin-top:1em">
+{ pkgs ? import &lt;nixpkgs&gt; {}
+}:
+pkgs.mkShell {
+  name = "dev-shell";
+  buildInputs = [
+    pkgs.python3
```
davidak, line 82: `with pkgs` would make it less cluttered and is what is used in practice

index.tt
```diff
@@ -6,152 +6,349 @@
...
+    pkgs.python3Packages.virtualenv
+    pkgs.nodejs
+    pkgs.yarn
+  ];
+}</pre>
+        <p>Then enter development environment with:</p>
+        <pre class="well" style="margin-top:1em">
+<span class="nix-command"><strong>nix-shell</strong></span>
+(nix-shell)<span class="nix-command">virtualenv --version</span>
```
davidak, line 91: before publishing this we should check if the tools like virtualenv actually work as expected

index.tt
```diff
@@ -6,152 +6,349 @@
...
-          <i class="fa fa-chevron-right"></i> Learn more about NixOS
-        </a>
+      <div class="span6">
+
+        <h3>Isolated development environments</h3>
+        <p>
+          After you get familiar with <code>nix-shell -p</code> you can take
+          the next step further and <a href="[%root%]learn.html">learn some
+            Nix</a>. To setup a more persistent environment you can also write a
+          simple <code>shell.niix</code> file:
```
davidak, line 74:
```suggestion
          simple <code>shell.nix</code> file:
```

index.tt
```diff
@@ -6,152 +6,349 @@
...
+        <p>Then enter development environment with:</p>
+        <pre class="well" style="margin-top:1em">
+<span class="nix-command"><strong>nix-shell</strong></span>
+(nix-shell)<span class="nix-command">virtualenv --version</span>
+16.7.9
+(nix-shell)<span class="nix-command">yarn --version</span>
+1.22.4</pre>
+        <p>
+          Commit the above <code>shell.nix</code> file and let you coworkers have
+          easier time setting their development environment.
```
davidak, line 97: you can "have easier time setting their development environment" with vagrant, docker, ...

be more specific and say e.g. "Commit the above <code>shell.nix</code> file and let your coworkers reproduce your environment."

is "Commit" enough that anyone know we are talking about version-control? the audience here is developers, so probably yes. but there are people developing software that don't use VCS, but they probably also don't would use Nix

coworkers is very work related. maybe something like "team mates"? it can also be a team of volunteers in some open source project like this

index.tt
```diff
@@ -6,152 +6,349 @@
...
-        More about NixOps
-        <i class="fa fa-angle-right"></i>
-      </a>
+  </section>
+  <section class="tab-pane" id="build">
+    <div class="row-fluid">
+      <div class="span6">
+        <h3>Minimal docker image</h3>
+        <p>
+          Declarative way to build minimal docker images. No build tools inside
```
davidak, line 108: >Declarative way to build minimal docker images.

there is something missing

index.tt
```diff
@@ -6,152 +6,349 @@
...
+          docker image with <strong>only</strong> Python 3 installed in it.
+        </p>
+        <pre class="well" style="margin-top:1em">
+{ pkgs ? import &lt;nixpkgs&gt; {}
+}:
+pkgs.dockerTools.buildLayeredImage {
+  name = "only-python";
+  contents = [ pkgs.python3 ];
+}</pre>
+        <p>To build and run the image you need to:</p>
```
davidak, line 123: it should probably explicitly be stated that it has to be run **in the same directory** where the default.nix was **created**.

i asked myself, why don't i have to add parameters to "nix-build"?

index.tt
```diff
@@ -6,152 +6,349 @@
...
+          The following Nix expression (<code>default.nix</code>) defines a
+          docker image with <strong>only</strong> Python 3 installed in it.
+        </p>
+        <pre class="well" style="margin-top:1em">
+{ pkgs ? import &lt;nixpkgs&gt; {}
+}:
+pkgs.dockerTools.buildLayeredImage {
+  name = "only-python";
+  contents = [ pkgs.python3 ];
+}</pre>
```
davidak, line 122: a better example would be something like a container you would actually build and deploy, like a webservice

so we could add one additional like like `config.Cmd = [ "/bin/hello" ];` and use curl in the example

index.tt
```diff
@@ -6,152 +6,349 @@
...
+        <p>To build and run the image you need to:</p>
+        <pre class="well" style="margin-top:1em">
+<span class="nix-command"><strong>nix-build</strong></span>
+...
+/nix/store/hb8xwmgd1pfrwgmc3indi2rrgcjr3wa3-docker-image-only-python.tar.gz
+<span class="nix-command">docker load -u ./result</span>
```
davidak, line 128:
```suggestion
<span class="nix-command">docker load -i ./result</span>
```

index.tt
```diff
@@ -6,152 +6,349 @@
...
+...
+Loaded image: hello:hb8xwmgd1pfrwgmc3indi2rrgcjr3wa3
```
davidak, line 130:
```suggestion
Loaded image: only-python:fgzj21lg3hjv8bxlaabcsfjkh4fg5ssk
```

index.tt
```diff
@@ -6,152 +6,349 @@
...
+        </p>
+        <p>
+          The following Nix expression (<code>default.nix</code>) defines a
+          docker image with <strong>only</strong> Python 3 installed in it.
+        </p>
+        <pre class="well" style="margin-top:1em">
+{ pkgs ? import &lt;nixpkgs&gt; {}
+}:
+pkgs.dockerTools.buildLayeredImage {
+  name = "only-python";
```
davidak, line 120: one additional line here makes the following examples shorter

`tag = "latest";`

index.tt
davidak, line 130:
```suggestion
Loaded image: only-python:latest
```

index.tt
```diff
@@ -6,152 +6,349 @@
...
+}</pre>
+        <p>To build and run the image you need to:</p>
+        <pre class="well" style="margin-top:1em">
+<span class="nix-command"><strong>nix-build</strong></span>
+...
+/nix/store/hb8xwmgd1pfrwgmc3indi2rrgcjr3wa3-docker-image-only-python.tar.gz
+<span class="nix-command">docker load -u ./result</span>
+...
+Loaded image: hello:hb8xwmgd1pfrwgmc3indi2rrgcjr3wa3
+<span class="nix-command">docker run hello:hb8xwmgd1pfrwgmc3indi2rrgcjr3wa3 python -c "print('hello world')"</span>
```
davidak, line 131:
```suggestion
<span class="nix-command">docker run only-python python -c "print('hello world')"</span>
```

index.tt
```diff
@@ -6,152 +6,349 @@
...
+        <p>
+          Learn more <a href="[%root%]nixpkgs/manual/#sec-pkgs-dockerTools">how
+            to build docker images</a>.
+        </p>
+
+      </div>
+      <div class="span6">
+
+        <h3>Declarative cloud images</h3>
+        <p>How hard would it be to build and configure <strong>Amazon AMI?</strong></p>
```
davidak, line 142:
```suggestion
        <p>How hard would it be to build and configure a <strong>Amazon AMI</strong>?</p>
```

index.tt
```diff
@@ -6,152 +6,349 @@
...
+        <p>
+          With the following <code>amazon.nix</code> we defined nginx which is
+          serving example <code>/var/www</code> folder, having a valid ssl
```
davidak, line 145: to make this more like a real world example: how would `/var/www` get content?

upload via ssh when it runs?

can we improve the experience with this example, so when someone actually does it, they should be rewarded with some website loading over https. maybe NixOS manual? or there is that nginx default page, use that as it's the same on ubuntu server after installing nginx with apt

index.tt
```diff
@@ -6,152 +6,349 @@
...
+    virtualHosts."example.com" = {
+      enableACME = true;
+      forceSSL = true;
+      locations."/".root = "/var/www";
+    };
+  };
+}</pre>
+        <p>Now we just need to build it and upload it.</p>
+        <pre class="well" style="margin-top:1em">
+<span class="nix-command"><strong>nix-build '&lt;nixpkgs/nixos/release.nix&gt;' -A amazonImage.x86_64-linux --arg configuration ./amazon.nix</strong></span>
```
davidak, line 169: that's quiet a long line...

index.tt
```diff
@@ -6,152 +6,349 @@
...
+...
+/nix/store/wmg6gcgnh6bk8d98syydfq73q6hjv682-nixos-amazon-image-20.09pre130979.gfedcba-x86_64-linux
```
davidak, line 171: promote stable channel instead? i think it should be used in production environments

index.tt
```diff
@@ -6,152 +6,349 @@
...
+<span class="nix-command">ls -lh ./result/</span>
+-r--r--r--     1 root root   1.4G Jan  1  1970 nixos-amazon-image-20.09pre130979.gfedcba-x86_64-linux.vhd
+dr-xr-xr-x     2 root root      4 Jan  1  1970 nix-support
+<span class="nix-command"><strong>nix-shell -p awscli</strong></span>
+(nix-shell)<span class="nix-command">export AWS_IMAGE="./result/nixos-amazon-image-20.09pre130979.gfedcba-x86_64-linux.vhd"</span>
+(nix-shell)<span class="nix-command">aws s3 cp --region $REGION $AWS_IMAGE "S3://${BUCKET}/${AWS_IMAGE#/}"</span>
```
davidak, line 177: explain where `$REGION` comes from

copying is fine, but they should also run it and see the website

index.tt
```diff
@@ -6,152 +6,349 @@
...
+    <div class="row-fluid">
+      <div class="span6">
+        <h3>TODO: Ideas</h3>
+        <ul>
+          <li>How to integrate with CI (travis, github actions, buildkite, ...) but that is notreally a deployment</li>
+        </ul>
+      </div>
+      <div class="span6">
+        <h3>NixOps</h3>
+        <p>TODO: can not come with a good example that is simple enough</p>
```
davidak, line 193: you could do the exact same as with AMI example, but that might be lame

check their manual. some example which setups multiple servers, like 2 or 3 with virtualbox. or better nixos containers as virtualbox might not be installed. explain you could deploy the same servers to X, Y, Z...

index.tt
```diff
@@ -6,152 +6,353 @@
...
+    <p class="lead">
+      NixOS package manager can be used independently on any POSIX compliant
+      operating system to reliably share your development and build
+      environments across machines.
+    </p>
+    <p class="lead">
+      NixOS, an independently developed GNU/Linux distribution, aims to improve
+      the state of the art in system configuration management. The entire
+      operating system, including the kernel, applications, system packages and
+      configuration files, are built from a single location.
```
davidak, line 22:
```suggestion
      configuration files, are built from a single configuration.
```

index.tt
```diff
@@ -6,152 +6,353 @@
...
+    </p>
+    <!-- <a class="btn btn-large" href="[% root %]download.html">Why NixOS</a> -->
+    <a class="btn btn-large btn-success" href="[% root %]learn.html">Get started</a>
+    <a class="btn btn-large" href="[% root %]download.html">Download</a>
+    <a class="" href="https://github.com">View on GitHub</a>
```
edolstra, line 27: Do we need "View on GitHub"? In any case it currently links to the GitHub front page.

index.tt
```diff
@@ -6,152 +6,353 @@
...
+    </p>
+    <p class="lead">
+      NixOS, an independently developed GNU/Linux distribution, aims to improve
+      the state of the art in system configuration management. The entire
+      operating system, including the kernel, applications, system packages and
+      configuration files, are built from a single location.
+    </p>
+    <!-- <a class="btn btn-large" href="[% root %]download.html">Why NixOS</a> -->
+    <a class="btn btn-large btn-success" href="[% root %]learn.html">Get started</a>
+    <a class="btn btn-large" href="[% root %]download.html">Download</a>
```
edolstra, line 26: Maybe the "Download" link should be first. The "Get started" page is really more about support / docs than getting started.

index.tt
```diff
@@ -6,152 +6,353 @@
...
-      <div class="learn-more">
-        <a class="btn btn-success" href="[%root%]nix/">
-          <i class="fa fa-chevron-right"></i> Learn more about Nix
-        </a>
+<div class="row-fluid">
+
+  <div class="span6">
+    <h1>Reproducible builds and deployments.</h1>
+    <p class="lead">
+      NixOS package manager can be used independently on any POSIX compliant
```
edolstra, line 14: This should read something like "Nix is a package manager for Linux and macOS that allows you to reliably share...". Although the word "share" is a bit vague...

index.tt
```diff
@@ -6,152 +6,353 @@
...
+      operating system to reliably share your development and build
+      environments across machines.
+    </p>
+    <p class="lead">
+      NixOS, an independently developed GNU/Linux distribution, aims to improve
```
edolstra, line 19: -> "NixOS is a Linux distribution that aims to...". "Independently" doesn't really mean anything (independent from what?).

index.tt
edolstra: "a single location" / "a single configuration" is not very clear. (What does "single" mean? It's thousands of files...) Maybe something like "a reproducible specification".

index.tt
davidak: independent from debian or redhat. it's not based on any other distro

index.tt
davidak: And make the words Nix and NixOS bold, like in ldleworks example. I also like the sentences that short, just visually.

index.tt
davidak: >What does "single" mean? It's thousands of files...

a single file that is used to build NixOS (`What does "single" mean? It's thousands of files...`). it can import content from other files like in other programming languages to organize big configurations and make it more modular, but that is something for advanced users of NixOS. the landingpage is for newcomers

i think "configuration" is the right word here. it's what admins work with. that's why it's called "configuration management". it's also `configuration.nix` and not "NixOS specification"

as a former admin, i don't really know how a "specification" would look like. but i have some experiences with configuration files and some expectations how they look

index.tt
davidak: it should be replaced with the getting started guide when it exists

index.tt
davidak: in my opinion no. it's the last place where you go, when you start contributing as a Nix/NixOS user

such a link belongs in the footer next to social media links

index.tt
garbas: To move on with this PR i will default to current text that we have on nixos.org. We can then fix it in a follow up PRs.

index.tt
garbas: I bolded "Nix" and "NixOS" but I kept the current text that is on nixos.org. We can/should update it in follow up PRs.

nixos/options.tt
garbas: For now I will leave it as it is. Currently we don't have a site search and we can add it in follow up PR is we decide to. But that means we will also have to select the search engine, etc.. which is another can of worms.

index.tt
garbas: I changed the example. I will review them again tomorrow.

index.tt
garbas: :+1:  i will fix this.

index.tt
garbas: I will hide which channel do we use and also all the hashes. this should also make it fit to the code block and not overflow.

index.tt
garbas: i made it multiline

index.tt
garbas: for now i'm skipping this examples. we can add them later.

index.tt
garbas: For now i will only write how to build it. we can later improve on this

index.tt
garbas: I reverted Nix and NixOS description the current description that is present in nixos.org. we can comeback to this and improve it in separate PRs.

index.tt
garbas: removed

index.tt
garbas: @edolstra I switch the buttons. Now Download is first (and green) and "get started" is gray and second.
@davidak exactly as you said. once we will have a get started guide we can link to is. I can already see this being shaped in #418

layout.tt
garbas: Me as well but for now we will stick to what is already present on current website. we can then discuss which search engine we want to default to.

garbas:
I'm closing this PR since it became to long to follow discussion. I'm opening a new one (from the same branch) with a summary what is still to be done. I hope to make it easier to review.

PR #393, opened 2020-04-13
indentation fix

garbas:


davidak:
Does it make sense to do identation like this:

```
<a>
  b</a>
```

instead of 

```
<a>
  b
</a>
```
like done here?

garbas:
I like to see that I'm closing the tags at the same level.

garbas:
looks like github action is stuck

garbas:
not anymore. I reran the job

PR #394, opened 2020-04-14
Remove Azure/GCE image tabs

edolstra:
It's very outdated.

Also update this flake to use NixOS 20.03.


grahamc:
One thing I miss about this new download page is that amazing, big green "next: installation manual" link. That was so helpful for me getting started.

davidak:
@grahamc i also said that https://github.com/NixOS/nixos-homepage/pull/386#discussion_r407035128

PR #403, opened 2020-04-20
Update flake.lock in cronjob

garbas:


garbas:
@samueldr Would this be a good strategy to go forward? To always regenerate flake.lock file on cron job and push changes to master. Should we also do this on master branch?

update.sh
```diff
@@ -2,4 +2,4 @@
 
 UPDATE=1 nix run nixpkgs#gnumake nixpkgs#curl -c make update --keep-going || true
 
-nix flake update --update-input released-nixpkgs || true
+nix flake update --recreate-lock-file || true
```
edolstra, line 5: Why this? Seems better to update the minimum necessary.

garbas: You are right. I should have just added nix-pills to be also updated.

worldofpeace:
@garbas I'm currently updating the docs with how to update nixos-homepage.
Is it possible to trigger this job manually?
This is important when you release nixos, because you have to be sure you get the exact revision of nixpkgs right.

edolstra:
@worldofpeace You can run `nix flake update --update-input released-nixpkgs --commit-lock-file`.

worldofpeace:
Thanks @edolstra, will embed that in https://github.com/NixOS/nixpkgs/pull/86368

Issue #406, opened 2020-04-21
https://nixos.org/nixos/support.html is missing yet still referenced in nixpkgs


operator-name:
I'm unsure if this should be a bug here or in nixpkgs, the offending nixpkgs source is [nixos/modules/misc/version.nix#L113](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/misc/version.nix#L113) which at least appears in Pantheon's about settings page.

![Screenshot from 2020-04-21 18 15 41](https://user-images.githubusercontent.com/53277700/79894394-4a049400-83fd-11ea-82f3-8acb9c9bf6f3.png)

[The last archive of the content](https://web.archive.org/web/20191027102323/https://nixos.org/nixos/support.html) seems similar to #402 so I assume things were moved around.

davidak:
The right URL is probably https://nixos.org/learn.html now, but we will discuss it.

Since it's related to the website, the issue is fine here.

Thank you for the report.

garbas:
I would say to change the following:
```
DOCUMENTATION_URL="https://nixos.org/learn.html"
SUPPORT_URL="https://nixos.org/community.html"
```

samueldr:
We should still provide a redirect to a page that makes sense.

davidak:
I created PRs for the suggestions.

PR #408, opened 2020-04-21
Add redirect for /nixos/support.html

davidak:
Came up in https://github.com/NixOS/nixos-homepage/issues/406

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -29,6 +29,12 @@
   force = true
 
 [[redirects]]
+  from = "/nixos/support.html"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -28,6 +28,12 @@
   status = 302
   force = true
 
+[[redirects]]
+  from = "/nixos/support.html"
+  to = "/learn.html"
```
garbas, line 33: this should redirect to `/community.html`.

davidak: OK

PR #412, opened 2020-04-24
Add Commercial support page

davidak:
includes removed content from previous community page

also removes content that is already on the community page and links to 
it instead

### DIFF
--- a/commercial-support.tt
+++ b/commercial-support.tt
@@ -1,0 +1,22 @@
+[% WRAPPER layout.tt hideTitle=1 title="Commercial support" menu='nixos' %]
+
+<h1>Commercial support</h1>
+
+<div class="lead">
+  A number of consulting companies are available for professional support.
+</div>
+
+<p>The list is sorted in alphabetical order.</p>
+
+<ul>
+  <li><a href="https://www.enlambda.com/">Enlambda</a></li>
+  <li><a href="https://nixos.mayflower.consulting">Mayflower</a></li>
+  <li><a href="https://nixcloud.io/">nixcloud</a></li>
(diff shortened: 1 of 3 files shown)


commercial-support.tt
```diff
@@ -0,0 +1,21 @@
...
+
+<h1>Commercial support</h1>
+
+<div class="lead">
+  A number of consulting companies are available for professional support.
+</div>
+
+<p>The list is sorted in alphabetical order.</p>
+
+<!-- Please submit a pull request to add your company -->
```
aepsil0n, line 11: Instead of this comment I would prefer a visible edit-link to GitHub. Admittedly this is a nice riddle for the exploration-minded visitor, but unfortunately not everyone bothers to read the source code or to find the repository. Especially in a professional context time is money, so we should make this as easy as we can.

davidak: good idea, but that is already an improvement. it's small, so i will add it, but i want get this finished and come to it later again

aepsil0n:
Thanks for including my comment so quickly. Otherwise the PR looks good to me. I'm not allowed to merge it though.

davidak:
@aepsil0n you can add a :heavy_check_mark: as a review.

@garbas will probably decide tomorrow if this should be merged

aepsil0n:
:heavy_check_mark:
(like this?)

davidak:
@aepsil0n no :D 
 Click on "Files changed" at the top and there on "Review changes".

aepsil0n:
Oh, thanks. I was not aware of that feature. Looks like I have not used GitHub enough lately. ;)

garbas:
@davidak :1st_place_medal:

PR #413, opened 2020-04-24
Make call more visible on NixCon Team page

davidak:
@andir @zimbatm @flokli what do you think about this change?

### DIFF
--- a/teams/nixcon.tt
+++ b/teams/nixcon.tt
@@ -7,6 +7,10 @@
   from any other OS/FOSS conference.
 </div>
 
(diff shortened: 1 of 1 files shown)


zimbatm:
Sounds good!

garbas:
Maybe - until our news section is done - we can have latest about nixcon also included on the website.

https://discourse.nixos.org/t/nixcon-2020-status-update/6914

davidak:
@garbas good idea. i would like to have a short summary and link to forum.

PR #415, opened 2020-04-25
Improve readability of text on site

aepsil0n:
This commit includes a few notable changes to the layout in order to improve its readability:

- Standard font size increased to 14pt
- Line height increased to 1.5em
- Width of single-column pages limited to 40em

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -1,6 +1,8 @@
 /* Account for the top navbar. */
 
 body {
+    font-size: 14pt;
+    line-height: 1.5em;
     padding-top: 50px;
 }
 
@@ -24,6 +26,10 @@ p {
     -webkit-hyphens: auto;
     -moz-hyphens: auto;
     hyphens: auto;
+}
+
+li {
+  line-height: normal;
 }
 
 .brand > a {
@@ -181,6 +187,10 @@ pre {
 .nix-shell-command:before {
     color: #999;
     content: " [nix-shell]$ ";
+}
+
+div.main.full-width {
+  max-width: 40em;
 }
 
 div.main li {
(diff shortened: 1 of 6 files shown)


css/nixos-site.css
```diff
@@ -183,6 +189,10 @@ pre {
     content: " [nix-shell]$ ";
 }
 
+div.main.full-width {
+  max-width: 40em;
+}
```
samueldr, line 194: The `full-width` name is counter-intuitive. I would have assumed, in a non-semantic way, that this means the element will use the full width of the parent container, rather than be constrained.

aepsil0n:
Hey there :)

Thank you for the review! I suspected this wasn't going to be that easy after having a look at the code. Anyway, I wanted to give a try.

I have seen your branch, so there likely is some duplicated effort. Before digging deeper: Do you think it's useful to do this? Are there better ways to help out with the redesign?

samueldr:
Since I did the work, I'm biased to say that it was useful to do this.

Though, it should be coming up in the next meeting, it would be "too much work to maintain" for the expected short time left until work for the actual redesign happens.

As far as helping with the redesign, the redesign will not start from the current implementation, and it's not going to be a step-wise approach. I figure we'll soon know more what's planned and how to help.

With that said, you can count on me to make sure integration of the new design implements what is necessary so it ends up fully scalable, making sure the text is always good and readable.

aepsil0n:
Oh, I was a bit vague there. I was questioning whether me continuing to work this PR is useful.

Anyway, it is probably easier to discuss face-to-face, so I'll hold up until the next meeting, before putting any more effort into this.

samueldr:
Technically speaking, it's going to be going against a core implementation detail of bootstrap 2, so way harder than it should be. As I haven't done the full work on top of bootstrap 2, I cannot say what other issues there could be. Though it does require all pages to be combed through carefully, including the manuals, which is not a trivial task.

aepsil0n:
After reading #409 and the last meeting, I believe there are more conversations to be had, before moving on with this. So I am closing this issue.

PR #418, opened 2020-04-30
Add quickstart guides

aepsil0n:
I have converted this quickstart guide into a PR now to make it easier to bikeshed.

pinging @domenkozar, as you had some thoughts on guides/tutorials to be added to the site, too?

PR #422, opened 2020-05-11
Update to use `nix shell` instead of `nix run`

aepsil0n:
This was renamed upstream, see https://github.com/NixOS/nix/pull/3551

For reference, I ran into this here: https://github.com/NixOS/nixos-homepage/pull/418/checks?check_run_id=663327024

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-UPDATE=1 nix run nixpkgs#gnumake nixpkgs#curl -c make update --keep-going
+UPDATE=1 nix shell nixpkgs#gnumake nixpkgs#curl -c make update --keep-going
 
 nix flake update \
   --update-input released-nixpkgs \


davidak:
We should wait for a new Nix release or even for it to land in NixOS stable? At least add which Nix version is needed for it to work.

How should this dependency hell be handled?

```
nix-shell -I nixpkgs=https://hydra.nixos.org/build/118764138/download/1/nixos-20.09pre225140.f83c5d3432d.tar.xz -p nix --run "nix shell nixpkgs#gnumake nixpkgs#curl -c make update --keep-going"
```

😬

edolstra:
> https://hydra.nixos.org/build/118764138

We shouldn't refer to Hydra because 1) Hydra is not performant enough as a distribution server; and 2) Hydra output could be garbage-collected (they're not at the moment, but they could be).

PR #427, opened 2020-05-17
Fix comma-separated list on /learn.html

lukegb:
The comma separated lists are really a display concern, and not a
markup concern - in particular, they're semantically lists.

In addition, laying them out like this means it's easy to forget
to put a trailing comma at the end of the line, like is present on
nixos.org/learn.html at the moment (which currently has:
"[Python], [Ruby] [Java], [Go], [Rust], " - note the missing comma
between "Ruby" and "Java").

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -436,3 +436,15 @@ div.docbook span.command {
     text-align: center;
   }
 }
+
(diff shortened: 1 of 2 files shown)


garbas:
Thank you.

PR #437, opened 2020-06-03
Fix bugs related to search on the index page

### LINKED ISSUE #436: Clicking the search option "... for options" leads to packages
Since tonight the website looks different. Options are not directly linked anymore. However I can search for them. I open the dropdown, click on "... for options" and I'm being lead to the package list.

The options are currently not reachable.

I'm on Firefox 76.0.1.

samueldr:
Options search simply doesn't work, it goes to the packages search.

Packages search doesn't work because it tries to use `nixos-undefined` as a channel. Look at the commit for more information why.

Fixes #436


garbas:
Tnx.

PR #438, opened 2020-06-04
homepage: Fix horizontal scroll on narrower widths

samueldr:
This is not trivially screenshotable, you have to compare by resizing the window.

I had to break out of bootstrap for the side-by-side "hero" unit, I can't see a non-hacky way full of custom rules compared to letting the flexbox layout deal with it.

Note that this also adds overflow-x to all the examples at the bottom, which otherwise would be wrapped weirdly. I find that neither options (wrapping weirdly or scrollbars) are satisfactory on wider widths, but on phone widths I know for sure scrolling horizontally those examples is better.

The asciinema box **will be weird**. I don't know if we can somehow make its text automagically rather than force an horizontal scroll (or worse, cutting off according to its defaults!).

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -507,12 +507,22 @@ ul.comma-separated > li:last-of-type:after {
   }
 }
 
-#examples {
-  margin: 0;
-  margin-top: 30px;
-  padding: 1em;
-  border: 1px solid #DDDDDD;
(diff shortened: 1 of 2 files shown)


garbas:
@samueldr :+1: , i will close this PR and push your branch to this repo and then open a PR from there. this will then create a preview.

PR #454, opened 2020-06-04
Integrate new search

garbas:
fixes https://github.com/NixOS/nixos-search/issues/50

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/454)


garbas:
@samueldr i think search.nixos.org is now ready to take place of packages-explorer. Am I missing something?

samueldr:
*This is not complete, I'm saving right now before I lose the contents*

## Navbar

The navbar's logo should probable link back to the nixos.org website. I would add a first option between the logo and "Packages" that says something like "Back to nixos.org" or similar, just to ensure people "are not trapped" on the packages website. (Yes, I know our target audience is tech-savvy and knows about the back button, but this is the kind of detailing work that can make or break changes in sites).

Additionally, it is pretty weird to see the navbar being a bit "wider" there.

* * *

## Error messages

> **Network Error!**
> Please check your network connection.

We may want to introduce a note about content blockers, something like the following (if we can distinguish third party from first party)

> **Network Error!**
> A request to a third party domain failed. This is either due to a content blocker or a networking issue.

* * *

## Permalink

This new search lacks the *attrname-based permalink* of the previous implementation:

 * https://nixos.org/nixos/packages.html?attr=linuxPackages_4_14.kernel

Note that this shows *only* the relevant result. This is especially good for third parties that want to link to the precise page of a package. I'm not sure, but I think the wiki has a template that uses this feature so an attribute name can be linked directly to.

Going 

* * *

## Navigation to a freshly opened row

Navigating to an open row is not good.

 * Resize your viewport or zoom so that [a search's results](https://search.nixos.org/packages?query=firefox&from=0&size=15&channel=unstable) **last entries fall under the fold**.
 * Scroll down
 * Click the last option
 * **Observe** (minor) the whole contents vanish for a brief moment (mostly cosmetic)
 * **Observe** (major) you are scrolled up to the top, with no clear indication that something really happened "down there".

Coincidentally, or maybe not, it looks like whatever handles the history API navigation does not handle resetting the scroll to where the user was. (TBF, the previous search didn't either, it is more of a clue.)

I think though, that the problem actually comes from the mostly cosmetic vanishing of contents. As the DOM is emptied, scroll goes naturally to the top.

samueldr:
(Unrelated to the search itself, but to the transition.)

We have to ensure that the redirects will transmit the query string and the anchor. I believe the anchor is fine, as this is done by the browser on redirect, the server has no way to influence it.

This is to ensure that an old link to `/nixos/packages.html?*whatever*` gets redirected to `search.nixos.org/packages?*whatever*`.

The options page is the one that requires a bit more finesse. `/nixos/options.html#something` needs to be redirected to `search.nixos.org/options?query=something`. This can either be done by having a dummy html page that handles rewriting to a query string at the location of the old options.html page, or by having the Elm app know about that specific. **There are many links online linking to that anchor-based permalink scheme**.

garbas:
@samueldr thank you thank you thank you :) I will go over the issues and try to address them.

samueldr:
And thank you again for your hard work. I personally know it's not easy re-implementing such a fundamental part of the website.

chrissound:
I think moving it at the bottom isn't ideal for discover-ability. The whole reason I even knew about the options search is because it was on the front page navigation (otherwise I assumed one just browses the nix source code to figure things out).

Possibly we could link the options search on the package search page? That could also help.

garbas:
Blockers were now fixed

06kellyjac:
Note for people stumbling over this PR from other issues:
This changed search is now located at <https://search.nixos.org/packages> (<https://search.nixos.org/> will redirect there too) and you should be auto redirected there from the old <https://nixos.org/nixos/packages.html>

PR #461, opened 2020-06-07
add missing <span class="nix-command">s

joachimschmidt557:


garbas:
@joachimschmidt557 thank you!

PR #464, opened 2020-06-08
Follow nix directly from the nix repository, following nix-latest branch

garbas:
also included in this PR:
* nixpkgs input updated
* remove the usage of `with` over whole outputs scope, just use it when listing `builtInputs`
* /nix/install now redirects to https://channels.nixos.org/nix-latest/install which points to latest nix release on https://releases.nixos.org

This makes it possible to update released Nix version on the website without touching `nixos-homepage` repository and making any kind of PRs to bump version.

Welcome to the **future**! :smile_cat:

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/464)


garbas:
@edolstra I'm now following the `2.3-maintenance` branch for nix releases and documentation. I think this is how I should have do it in the first place. Still learning how to use flakes :)

flake.nix
```diff
@@ -2,112 +2,123 @@
...
 
   # This is used to build the site.
-  inputs.nixpkgs.url = "nixpkgs/nixos-20.03";
+  inputs.nixpkgs = { url = "nixpkgs/nixos-20.03"; };
 
-  # These are used for the manuals, and release artifacts
-  inputs.released-nixpkgs.url = "nixpkgs/nixos-20.03";
+  # These inputs are used for the manuals and release artifacts
+  inputs.released-nixpkgs = { url = "nixpkgs/nixos-20.03"; };
+  inputs.released-nix = { url = "github:nixos/nix/2.3-maintenance"; flake = false; };
```
edolstra, line 9: A slight problem is that this branch also gets pushed to between releases (so it might point to an unreleased version). So it's not like the `nixos-*` branches which always correspond to a released version.

edolstra: So concretly, `NIX_VERSION` might end up pointing to a version that doesn't exist in releases.nixos.org yet.

garbas: Would it be possible to change the Nix release workflow and add some sort of `2.3-release` branch?

edolstra: Maybe a `latest` branch that gets force-pushed when we do a new release. I can add this to the release script.

garbas:
@edolstra I think this PR is now ready for next round of review and potentially merging it if things look good.

PR #465, opened 2020-06-08
Improve some wording in the demo

knedlsepp:
This fixes some grammar and improves wording in the demo.

I want to add that I'm not a native English speaker myself, so I'd appreciate a review from someone who is. :-)

### DIFF
--- a/demo.scenario
+++ b/demo.scenario
@@ -8,8 +8,8 @@
 # * everything else will be printed immediately
 
 $ # Hi!
-$ # Wonder how to start using Nix?
-$ # Here are few examples:
+$ # Wondering how to start using Nix?
+$ # Here are a few examples:
 
 --
 
@@ -31,11 +31,11 @@ $ bat default.nix
 ────────┬────────────────────────────────────────────────────────────────────
         │ File: default.nix
 ────────┼──────────────────────────────────────────────────────────────────── 
-    1   │ { pkgs ? import <nixpkgs> {} # here we import nixpkgs package set
+    1   │ { pkgs ? import <nixpkgs> {} # here we import the nixpkgs package set
     2   │ }:
-    3   │ pkgs.mkShell {               # mkShell is a helping function
+    3   │ pkgs.mkShell {               # mkShell is a helper function
     4   │   name="dev-environment";    # that requires a name
-    5   │   buildInputs = [            # list of packages
+    5   │   buildInputs = [            # and a list of packages
     6   │     pkgs.nodejs
     7   │   ];
     8   │   shellHook = ''             # bash to run when you enter the shell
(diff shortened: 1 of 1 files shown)


knedlsepp:
@garbas As suggested on discourse, I put my suggestions here as a PR. Please take them with a grain of salt since I'm not so sure about the changes myself.

demo.scenario
```diff
@@ -132,14 +132,14 @@ debian-redis         latest    8366943c77e8    3 days ago      136MB
 alpine-redis         latest    aae644cd3417    3 days ago      6.99MB
 nix-redis            latest    30486183a209    50 years ago    45.4MB
 nix-redis-minimal    latest    a21238890680    50 years ago    2.02MB
-$ # Did we just produced a docker image _smaller_ then Alpine? Interesting!
+$ # Did we just produce a docker image _smaller_ then Alpine? Interesting!
```
joachimschmidt557, line 135:
```suggestion
$ # Did we just produce a docker image _smaller_ than Alpine? Interesting!
```

garbas: hehe, I always get this wrong :) thank you!

demo.scenario
```diff
@@ -26,16 +26,16 @@ $ nix-shell -p nodejs
 --
 
 $ # Typing "nix-shell -p ..." each time can be tedious. We can do better.
-$ # We can write everything down in default.nix
+$ # We can write everything down in `default.nix`
```
garbas, line 29: This is not really a markdown document. But something that would get "typed" into a terminal in asciinema video. If I would be typing this in a terminal I wouldn't use back-ticks. Can you remove the back-ticks?
```suggestion
$ # We can write everything down in default.nix
```

demo.scenario
```diff
@@ -78,7 +78,7 @@ $ docker images | grep redis
 debian-redis  latest    8366943c77e8     3 days ago       136MB
 alpine-redis  latest    aae644cd3417     3 days ago       6.99MB
 nix-redis     latest    30486183a209     50 years ago     45.4MB
-$ # Size of a Nix based docker image is somewhere between a Debian and Alpine
+$ # The size of a Nix based docker image is somewhere between a Debian and an Alpine image
```
garbas, line 81: We need to make this git into 77 character in a line. I found 77 chars to produce the most readable video, while still having some space to write show things. Can you combine this with next line and make sure you don't go over 77 chars? even if we need to add additional line

demo.scenario
```diff
@@ -97,15 +97,15 @@ $ bat redis-minimal.nix
...
-$ # In redis-minimal.nix we override the default redis build with 3 changes:
-$ # 1.) remove redis systemd support
-$ # 2.) build to static binary with musl
-$ # 3.) only redis-server binary is needed
-$ # Now all we need to is build docker image with our new minimal redis
+$ # In redis-minimal.nix we override the default redis build with three changes:
+$ # 1.) Remove the redis systemd support
+$ # 2.) Build a statically linked binary with musl
+$ # 3.) Remove all binaries apart from redis-server
+$ # Now all we need to do is build the docker image with our newly created minimal redis
```
garbas, line 108: We need line to make fit 77 chars as well
```suggestion
$ # Now let's build the docker image with our newly created minimal redis
```

Issue #443, opened 2020-06-04
More documentations (tutorials/guides/…)


garbas:
Integrate nix.dev into nixos-homepage

garbas:
@edibopp You want to take this issue?

edibopp:
@garbas as discussed during the meeting, yes

edibopp:
So I thought about how to integrate the content on nix.dev into nixos.org more automatically. @garbas @domenkozar I am curious about your thoughts on this.

I believe we want to make users feel like they are still on an official NixOS site when reading guides.

In order to achieve that, I would suggest to

- maintain an updated link list to introduction guides on the learn page
- host the guides as part of the nixos.org domain
- adapt the styling of the guides to fit in with the NixOS homepage

### Updated link list

I don't think we need to specify implementation details up-front. Important here
is:

- listing all guides may be overwhelming, so I think it is fine to manually
  curate the relevant tutorials. this may overlap with nix.dev start page
  though
- requires nix.dev to provide stable links

### Build & deployment

I see two main options here:

- generate the guides using sphinx as part of the nixos-homepage build and
    depend on the nix.dev GitHub repository for this. we could pin a certain
    revision and do regular updates
- use a DNS entry to point guides.nixos.org → nix.dev (for example). this way we
    depend on nix.dev as a service

### Visual integration

This should be very simple, especially considering #449 I would not put a lot of
effort into this for now. But simple changes can go a long way already to create
a better integration. Mainly I would:

- Add the nixos.org top menu
- Adjust the fonts and colors to match nixos.org

This could be implemented as patches on top of nix.dev.

garbas:
@edibopp :100: 

### Updated link list

I had similar thoughts.

### Build & deployment

I would pin the repo using `flake.nix` since this is what we currently use to manage this. Since `nix.dev` not a flake, we can use `flake = false;`.

Currently `nix.dev` does not have any `default.nix` which would build the website with Nix, since they use Netlify to run the build. But maybe this would be something to send to `nix.dev` as a PR.

### Visual integration

Here I would do a bit differently.

I would copy out the content part of the specific guide and insert it into our guide template which would contain the website layout. I would also run few filters (eg. using tools such as `xmlstarlet`) to fix some things in html like: 
- make sure nix.dev links point to nixos.org/guides
- fix some minor visual quirks that require html change

Apart from that, as you suggested, we can add few minor styles fixes to make documentation look like the rest of the nixos.org.

PR #470, opened 2020-06-14
Copy first tutorials from nix.dev

edibopp:
The tutorials are sourced from https://nix.dev. The resulting HTML was adapted to remove Sphinx-specific references and styling.

This is meant to see how it fits into a page to get some first content onto the page.

----

As discussed during the last two marketing team meetings, this only the minimal step one to get some guide-like content onto the page following #418 and @domenkozar launching https://nix.dev. Feedback on the content would be better directed at https://github.com/domenkozar/nix.dev. Also first step of #443.

For next steps I would start automating this integration instead of doing it manually. I'll open a separate issue for that, also to discuss the strategy.


garbas:
Small step for human kind but HUGE for nixos.org :) /me still in affect from all the space news lately :rocket:

PR #472, opened 2020-06-15
move search from navigation bar to learn page

### LINKED ISSUE #457: Search "actions" in header can't be used on narrow viewports
The `... for packages` and `... for options` "drop down" items can't be activated using a mouse (unverified for keyboard) when the viewport is narrow. 

image

garbas:
fixes #457

### DIFF
--- a/guides/ad-hoc-developer-environments.tt
+++ b/guides/ad-hoc-developer-environments.tt
@@ -1,4 +1,4 @@
-[% WRAPPER layout.tt title="Ad hoc developer environments" hideTitle=1 menu='nixos' %]
+[% WRAPPER layout.tt title="Ad hoc developer environments" %]
 
 <h1>Ad hoc developer environments</h1>
 
--- a/guides/contributing.tt
+++ b/guides/contributing.tt
@@ -1,4 +1,4 @@
-[% WRAPPER layout.tt title="How to Contribute" hideTitle=1 menu='nixos' %]
+[% WRAPPER layout.tt title="How to Contribute" %]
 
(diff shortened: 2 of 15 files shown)


edolstra:
I'm not so sure about this. The only reason I visit nixos.org is for the options/package search...

davidak:
same

we have the links in the footer and package search on the start page. maybe add options search to a section about nixos (that don't exist yet) on the startpage?

so maybe just remove it from the menu? and put a site search there like people expect

edolstra:
I'd keep it in the navbar. You don't want to have to search for the search feature. If we need to save space in the navbar, it could be done like the Red Hat homepage, which has a search icon in the navbar.

garbas:
I'm going to leave a "search icon" link on the right side (pointing to packages search) just to keep it easier to find. Eventually (any day now) this will point to search.nixos.org where visitor will see that you can search also for nixos options.

garbas:
@davidak 

> we have the links in the footer and package search on the start page. maybe add options search to a section about nixos (that don't exist yet) on the startpage?

I tried and it doesn't look good since it complicated things. There is a link to options search.

> so maybe just remove it from the menu? and put a site search there like people expect

I don't think site search adds value since everybody is already using their own preferred search engine.

chrissound:
I think moving it at the bottom isn't ideal for discover-ability. The whole reason I even knew about the options search is because it was on the front page navigation (otherwise I assumed one just browses the nix source code to figure things out).

Possibly we could link the options search on the package search page? That could also help.

PR #474, opened 2020-06-16
I'm trying to reduce my scope and leave space for someone else

domenkozar:


grahamc:
It has been a pleasure, Domen =).

PR #480, opened 2020-06-19
Migrate the security teams GPG key download to keys.openpgp.org

mweinelt:
The old SKS system is flawed and shouldn't be used anymore. Hagrid (https://sequoia-pgp.org/blog/2019/06/14/20190614-hagrid/) is a relatively new key of replacement for the old SKS system. It strips the signatures from the keys, to prevent key poisoning, so the WoT is dead and gone.

Also key ids  can be cheaply forged, even long ones, see:
  - https://debian-administration.org/users/dkg/weblog/105
  - https://mailarchive.ietf.org/arch/msg/openpgp/Al8DzxTH2KT7vtFAgZ1q17Nub_g/

Therefore let's migrate to https://keys.openpgp.net and drop key ids entirely.

### DIFF
--- a/teams/security.tt
+++ b/teams/security.tt
@@ -43,13 +43,10 @@
       <a href="mailto:graham@grahamc.com">graham@grahamc.com</a>
       (gchristensen)
       <br />
-      <strong>GPG Key:</strong>
-      <tt>
-        <a href="https://pgp.mit.edu/pks/lookup?op=get&amp;search=0xFE918C3A98C1030F">0xFE918C3A98C1030F</a>
-      </tt>
-      <br />
       <strong>GPG Fingerprint:</strong>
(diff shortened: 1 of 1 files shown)


teams/security.tt
```diff
@@ -43,13 +43,10 @@
       <a href="mailto:graham@grahamc.com">graham@grahamc.com</a>
       (gchristensen)
       <br />
-      <strong>GPG Key:</strong>
+      <strong>GPG Fingerprint</strong>
```
cole-h, line 46: Nit:

```suggestion
      <strong>GPG Fingerprint:</strong>
```

mweinelt: Fixed.

edibopp:
lgtm, but @fpletz and @grahamc should probably confirm this.

edibopp:
Thanks for pushing on this, @flokli. I overlooked that the fingerprint was already in there, so no additional confirmation is required. So it's merged now.

Issue #481, opened 2020-06-20
make check references a unknown tool


nixinator:
in commit ae8536ffd52519987aa353b45d0f2a8f0426e1b6

Author: Eelco Dolstra <eelco.dolstra@logicblox.com>
Date:   Thu Nov 26 12:45:18 2009 +0000

added a 'make check' target to the Makefile.

This tool doesn't exist in nix-shell the environment built with the flake

The closest thing i see a linkchecker.. but not sure if it's off use any more.

you can test that with.

nix-shell
make

nix-env -i linkchecker
sed -i 's/checklink/linkchecker/g' Makefile

make check

------------------
I'm not sure if the linkchecker is of value.. or working well in this case... 

we can either put it back in, add linkchecker to the flake.

or remove the old cruft and the dependency from the make file.

for nostalgia , git checkout ae8536ffd52519987aa353b45d0f2a8f0426e1b6  and the web site from 2009!, and also the
hardpaths in the Makefile to eelco environment ;-).

nixinator:
seems linkchecker really wants to be pointed at a the webserver where you've just deployed the website... 

python -m http.server 8000

linkchecker http://127.0.0.1:8000

edibopp:
I feel like the tool would make even more sense to run against the deployed website. But a local run provides some value, too – if it works. By the way, linkchecker can take a local HTML file as input. A local run does not respect our Netlify redirects, of course, but we probably should not rely on them for internal links anyway.

garbas:
fixed in #487

PR #483, opened 2020-06-21
Enable local link checking and fix a few outdated links

edibopp:
This addresses #481.


garbas:
Awesome idea, but currently this check phase doesn't run. We also need to make sure that derivation doesn't build when links is missing.

Apart from this check I would also add a daily cron job to run `linkchecker` via github actions against https://nixos.org.

I would also suggest to push branches to this repo so that preview is created, so that a reviewer can actually "live test" the fixed links.

PR #485, opened 2020-06-21
added install and clean targets update flake.nix

nixinator:
This add two new targets (install, clean)

install , which scoops up the artefacts from the build process to build/

flake.nix modified to to make install and copy the build/ dir to $out.

the better way is to modify the make file entirely, but it's not a simple as it seems.

also this might break deployment to netify??????


garbas:
This approach will break the development cycle. Since it would not copy things to `build/` when items change. I would much rather see we explicitly move things to `build/`.

nixinator:
noted... 

alternatively  bring the copy changes into the deployment script?

    - name: Building nixos.org
      run: |
        nix build
        mkdir build
        cp -RL ./result/* ./build/
    - name: Deploy to Netlify

but the cleanest way to do it, is to completely rewrite the makefile.

garbas:
With the last proposed solution we would only catch problems once we deploy to Netlify and we wouldn't be able to catch them locally.

I still believe best approach would be to prepend `build/` to targets in Makefile.

PR #490, opened 2020-06-27
index: make (nix-shell) prompt label unselectable

jtojnar:
Screenshot:

image

### DIFF
--- a/css/nixos-site.css
+++ b/css/nixos-site.css
@@ -187,7 +187,7 @@ pre {
 
 .nix-shell-command:before {
     color: #999;
-    content: " [nix-shell]$ ";
+    content: "[nix-shell]$ ";
 }
 
 div.main li {
--- a/index.tt
+++ b/index.tt
@@ -95,18 +95,18 @@
 <span class="nix-command">python --version</span>
 python: command not found
 <span class="nix-command"><strong>nix-shell -p python3</strong></span>
-(nix-shell)<span class="nix-command">python --version</span>
+<span class="nix-shell-command">python --version</span>
 Python 3.7.7</pre>
 
         <h2>Multiple languages, one tool</h2>
         <p></p>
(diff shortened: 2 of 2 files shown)


jtojnar:
This works but we might want to leave the prompts in the code and just make them unselectable using CSS, see https://github.com/NixOS/nix/pull/3741

samueldr:
The accessbility implications (if any) would need to be checked for which is preferrable between a CSS-derived prefix, or disabling selection via CSS.

A quasi-accessibility issue that exists with CSS-derived content is that user agents without CSS may not get that content. E.g. some crawlers, or more importantly, text-based (console) browsers.

jtojnar:
Yeah, that was my concern as well. For regular browsers, `::before` should be seen be screen readers IIRC.

samueldr:
Yes, I wonder if there are *other* accessibility issues. Accessibility is hard!

Though I **know** that changing the contents via CSS is fraught with perils for some user agents. You're ending up losing some of the context. I'm thinking the most likely problematic user-agent (for the home page) would not be text browsers, but crawlers making "answer" based boxes. Though, in the manuals, definitely text-based browsers would end up losing context.

Additionally, changing the contents via CSS in cases like here pushes too much knowledge about the contents into the semantics, and then the presentation layer. It's not really wrong, but it's not exactly neat either.

I'm curious to hear opinions of others about that.

jtojnar:
Opened https://github.com/NixOS/nixos-homepage/pull/490, which is IMHO a better approach. The prompt is part of the screen just like the command or the output.

samueldr:
> Opened #490, which is IMHO a better approach. The prompt is part of the screen just like the command or the output.

Or #492? :)

jtojnar:
Yeah, that’s the one, sorry.

PR #492, opened 2020-06-27
Move shell prompts to the text

jtojnar:
They are part of the screen output just like the command or the printout.

Used `shell-prompt` class to avoid collision with Docbook’s prompt.

Screenshot:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ build it:
     $ cd nixos-homepage
     $ nix-shell
     [nix-shell]$ make
-    [nix-shell]$ python -m http.server 8000
+    [nix-shell]$ caddy
 
 then open http://127.0.0.1:8000/index.html
 
@@ -19,7 +19,7 @@ To automatically rebuild on every change:
 To test the complete result from a nix-build:
 
     $ nix-build
-    $ nix-shell --run 'python -m http.server 8000 --directory result'
+    $ nix-shell --run 'cd result/ && caddy'
 
 ## License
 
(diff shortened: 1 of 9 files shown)


jtojnar:
Thannks. However, the space should be part of the `prompt`, otherwise it gets copied and then the command will not be saved into shell history.

garbas:
ah right, i'll fix it asap.

jtojnar:
Looks like this was forgotten in the new design.

garbas:
@samueldr ping

Majsvaffla:
I installed nix today, following https://nixos.org/download.html#nix-install-windows, and noticed the prompt character was also selectable. I wanted to open an issue or MR myself but then I found this one. Since this one was merged almost three years ago, how come the changes havn't made it to the live website yet?

jtojnar:
Looks like the [stylesheet](https://github.com/NixOS/nixos-homepage/blob/7614bc91461a9239b2f2bb5faf848918613a1a0c/css/nixos-site.css#L181-L184) where the prompt is made unselectable is no longer used? Probably needs to be moved to https://github.com/NixOS/nixos-common-styles

Majsvaffla:
> Looks like the [stylesheet](https://github.com/NixOS/nixos-homepage/blob/7614bc91461a9239b2f2bb5faf848918613a1a0c/css/nixos-site.css#L181-L184) where the prompt is made unselectable is no longer used? Probably needs to be moved to https://github.com/NixOS/nixos-common-styles

Alright, I opened an MR over there 🙂

PR #500, opened 2020-07-27
Remove Governance from menu

davidak:
It is not that important that it should be in the menu. It's fine to be in the footer next to security.

People that are more interested in the project will find it there.

Removing it **now** will help the redesign process.

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -41,7 +41,6 @@
             <li><a href="[%root%]download.html">Download</a></li>
             <li><a href="[%root%]learn.html">Learn</a></li>
             <li><a href="[%root%]community.html">Community</a></li>
-            <li><a href="[%root%]governance.html">Governance</a></li>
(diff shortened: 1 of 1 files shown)


jwinnie:
Governance is very important for an operating system distribution; it's usually one of the first things I check when looking at a new Linux distribution. Good governance means high-quality packaging, maintainers who are easy to get along with, etc.; bad governance means chaos and instability, or worse, stagnation.

jonringer:
I would say that governance is important within the FOSS community in general, it would be unwise to make it less discoverable

davidak:
Sure governance is important, like security, but i think you don't visit these pages often. They don't get much new content to check out, unlike Learn, Community or Download.

While Features, Learn and Download are pages interesting to new visitors, that want to learn what Nix and NixOS is, governance, security and donate are more relevant to actual users.

But it's fine with me to leave governance in the menu. The page really shows that we are well organized and serious about the project.

PR #504, opened 2020-08-20
index.tt: Minor updates to wording.

toothbrush:
Here and there sentences didn't read like native English.  Also, the word "Docker" should be capitalised.


toothbrush:
I was thinking of trying out NixOS, and while reading the homepage thought i could probably improve on the wording a little bit.  Hope that's okay 🙌

edolstra:
Thanks!

PR #505, opened 2020-08-21
change manual location and add unstable version

### LINKED ISSUE #237: Keep versioned manuals
This is a multi-fold issue, I'm opening this quickly, this isn't as detailed as I'd like.

 * There are no `unstable` manuals available online.
 * Once a stable release updates, the manual links *may* end up being dead-ends.
 * There is no way to reference an older manual online.

I know hydra builds the manual, and I think they're available somehow online, but I don't really know how to access them.

I'd like opinions and suggestions about whether we should and how we should approach this.

(I'll write my owns later.)

garbas:
fixes #237

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/505)


fix-manual-headers.sh
```diff
@@ -0,0 +1,45 @@
+#! /usr/bin/env bash
```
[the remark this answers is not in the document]
garbas: I would leave this two scripts separate. We can revisit this once we work on next tooling for nixos-homework.

garbas:
> I guess the selector is WIP and unfinished?
> 
> image
> 
> Or is it a bug?

It looks like style is not applied. /me checks

edolstra:
In addition to `/manual/nixos/stable`, maybe we should also have `/manual/nixos/20.03` etc. like the Postgres manuals, which have https://www.postgresql.org/docs/12/index.html and https://www.postgresql.org/docs/current/index.html.

garbas:
> In addition to `/manual/nixos/stable`, maybe we should also have `/manual/nixos/20.03` etc. like the Postgres manuals, which have https://www.postgresql.org/docs/12/index.html and https://www.postgresql.org/docs/current/index.html.

Should we redirect 20.09 to unstable or unstable to 20.09 (I assume the second option)?

It doesn't make sense to me to have 2 urls showing the same url (like postgresql does it). I would expect "current" would redirect to "12".

I would still leave canonical url to be pointing at "stable" version of documentation and that will redirect to latest version.

Also should we start keeping older versions of documentation around (like postgresql)?

js/manual-version-switch.js
```diff
@@ -0,0 +1,26 @@
...
+    .append($("<dt/>").text("Versions"))
+    .append($("<dd/>").append(unstable))
+    .append($("<dd/>").append(stable));
+  let current = $("<div/>")
+    .click(function(e) {
+        e.preventDefault();
+        e.stopPropagation();
+        $(this).parent().toggleClass("opened");
+    })
+    .text("v: " + channel);
```
samueldr, line 20:
```
<i class="fa fa-caret-down" aria-hidden="true"></i>
```

Ideally with the proper rules to switch up "up" as needed

```
<i class="fa fa-caret-up" aria-hidden="true"></i>
```

Additionally, I don't know where the caret should be, right or left? A "proper" dropdown places it at the right, but this is not a "proper" dropdown!

I would also argue that + and - would be proper (too) since this is an expandable box

```
<i class="fa fa-plus" aria-hidden="true"></i>
<i class="fa fa-minus" aria-hidden="true"></i>
```

They also have a square variant if it makes sense here.

Tip: use the fontawesome 4 page to search for icons:

 * https://fontawesome.com/v4.7.0/icons/

garbas: I'll use + and - as you suggested then we can place it anywhere we want and see fit best.

garbas:
Another question that raised in my head since few days: Are we going to always keep documentation online, especially the versioned one?

What will happen is that with every release we will break what google has indexed. eg. All documentation under `/manual/nixos/20.03` will be removed once `20.09` is released.

For sake of moving forward with this I would suggest we only introduce channels (stable/unstable). 
It would be extremely nice to have all of the versions of documentation always there, but we might need to have another discussion regarding this especially once all the work around documentation format and tooling is decided.

/cc @edolstra

davidak:
> For sake of moving forward with this I would suggest we only introduce channels (stable/unstable).
> It would be extremely nice to have all of the versions of documentation always there, but we might need to have another discussion regarding this especially once all the work around documentation format and tooling is decided.

yes, we keep all the packages, so why would we not keep the manual. having stable and unstable is already a huge step :tada:

PR #506, opened 2020-08-27
[WIP] 2020 re-design for the website

samueldr:
**Note that the year is not an invitation to make this a yearly event!**

* * *

How we'll proceed:

 * Keep this PR as a draft until ready to be merged into master
 * Don't push to this branch directly
 * Open PRs against the `feature/2020-redesign` branch

This way we have this as the golden "next" website, it should be in a passable-to-ready state at any point, hopefully. At the moment this is opened, many pages are *not fine*, but depending on how we focus our energies either we can make them *work fine* first, or continue with the design of the components.

Note that criticism *about the design* in the sub-PRs will be broadly disregarded (but noted), to make the development go. Otherwise we'll stay stuck for long times in the implementation. Anyway most, if not all, of the suggestions can be bundled up at the end and fixed up at the end rather trivially; the implementation strives to stay flexible.

* * *

## Broad strokes TODO

I'll try to expand on those broad strokes later.

 * [x] Re-do and review JS interactive parts (still using "classic" jquery)
    * [x] Base setup
    * [x] Menu
 * [x] Finish implementing the new design components
 * [x] Have contributors (e.g. @garbas) figure out the contents

Misc tasks:

 * [ ] Find an appropriate way to minify the CSS (and find out if it even matters)
 * [ ] document browser compatibility (so that contributors/testers know where to test and where not to test)

* * *

## Specific TODO

 * [x] survey banner (lost while merging to reduce merge overhead)
 * [x] install nix button in the nav bar (what should it do exactly?)
 * [x] flavourful images for the "why nix" placeholder circles.
 * [x] Lorem ipsum for examples text


samueldr:
## Re-do and review the JS parts

Quite simply, take nothing for granted, as it was built with bootstrap's JS in mind, and not only that, but an older jQuery!

The plan is to

1. Update jQuery
1. Centralize *components-based* JavaScript behaviour

The latter part is weird, but basically I don't want to see any `$(".home-page .overlay").[...]`, but have this all built through graceful degradation, based on [*the component*](https://rscss.io/components.html) that is implemented, in a generic manner.

In addition, all behaviour should prefer working through CSS class name changes, rather than modifying DOM properties, where possible.

And, building from that last sentence, all behaviour has to gracefully degrade for JavaScript-less usage. I'm thinking about text browser uses, and content blockers. The main idea (to be done) is to rely on either `html.without-js` to override behaviour, or preferrably `html.with-js` to know that the CSS behaviour can be used. You'll see what I mean in the upcoming PR with JavaScript behaviour implementation.

hugolgst:
Hello, do you need help on this PR ?

hugolgst:
Hello @samueldr!

I have some comments about the current re-design of the web page:
- I think that the drop shadows are too much and not really present anymore on modern websites
- The green color looks a bit harsh and do not follow the "NixOS colors"
- The navigation bar takes a lot of space on the screen
- Is it me or the page looks really zoomed?
- We could add colors on the code snippets

Otherwise, I really like the re-design, especially the footer :)
If you need any kind of help do not hesitate to assign me or mention me.

samueldr:
> Hello @samueldr!

Hi!

> 
> I have some comments about the current re-design of the web page:
> 
> * I think that the drop shadows are too much and not really present anymore on modern websites

I'm implementing what was designed by an external (to this repo) designer, and selected in meetings with him and the team.

My opinion is that the death of the shadow in modern design is a waste. It makes things actually easier to comprehend, compared to way-too-flat design that only looks clean at glance, but actually isn't clear. Though here is not the time to gripe around bad popular modern design choices.

> * The green color looks a bit harsh and do not follow the "NixOS colors"

We don't have much of a palette except for the blues. Again, the designer has worked, has been working on the palette. If you have specific suggestions for colours, I guess we might be open 

> * The navigation bar takes a lot of space on the screen

It's been decided that it will not stay sticky with the website already, you're not alone thinking that.

> * Is it me or the page looks really zoomed?

It might be because you're used with the popular design trends, where fonts somehow are chosen to be too small. Many designs these days design around 14px-equivalent font sizes, while we're working off of 16px-equivalent as a base size. Then, everything is derived from the base font size.

One thing that might *also* be putting someone off is that this scales up with the width of the browser with extremely large windows. This might be turned off depending on what the community thinks, but this is a choice in the balance of designing for extremely large viewports. It's preventing another common complaint from bubbling up, which I'm sure will once we change that setting.

> * We could add colors on the code snippets

Yes. This part is currently not done.

> Otherwise, I really like the re-design, especially the footer :)
> If you need any kind of help do not hesitate to assign me or mention me.

Thanks, for the current moment the help I'm looking for is for the rest of the team to finish working on updating the *content*, so I'm pretty sure *you* can't since that implies knowledge you don't have of previous discussions :).

hugolgst:
About the font size, I find it really too much, I am on a 13" screen and it isn't really convenient.
The drop shadow I think that if you choose the right colors you don't need it and even if you want one it might be better if you reduce it?

garbas:
I merged the master and resolve the conflict on `index.tt` file.

edolstra:
Under "Examples" on the front page, the first row of examples is taller than the second row. Is that intended?

garbas:
> Under "Examples" on the front page, the first row of examples is taller than the second row. Is that intended?

I made the examples all the same size. The idea - in the future - is to have 9 examples, where 3 are more "important" for the visitor. More important meaning that they are more suitable for the newcomers, our main audience.

domenkozar:
cc @anamarijas

0x4A6F:
Nice work!

- asciinema-demo-cover doesn't wait between topics, so last output of a section is hard to grasp
- typo in asciinema-demo-example_6 "pacakge"

davidak:
The orange on light blue is harder do read than dark blue on light blue.

signal-2020-09-17-212304

i like how the asciicinema is now opened inline.

from my view this is ready to deploy :rocket:

davidak (approved):

samueldr:
> Under "Examples" on the front page, the first row of examples is taller than the second row. Is that intended?

To add to what @garbas said, the design included a short description (two lines) for the examples, and the smaller version was a 4:3 version leaning left to the example text.

<details>
<summary>Mockup for reference:</summary>

image

</details>

samueldr:
Hi, the "new strategy" for the "big angles" section *could* have a performance impact which has not been validated or tested yet.

I'm not positive on how to *properly* test it for performance, but I guess that if scroll is fine on an "underpowered" mobile device it should be fine.

Infinisil:
On my iPhone 5 (iOS 10.3.4), there's just NaN for the countdown. This is a very outdated device and OS though, I can't even update anymore :) 
image1

garbas:
@Infinisil the above commit should "fix" the counter. well I hide it in the case `new Date` returns `NaN`

samueldr:
@Infinisil while the device's OS is outdated and you should probably figure out an update for your own good, this is going to be quite useful to know if the site is *completely broken* or if features degrades nicely on a less-compliant browser.

If you can have a look around and compare with a narrow / zoomed-in browser window and gauge whether things are okay or totally wrong, it would be helpful.

Infinisil:
Two problems with text wrapping I noticed:
On the features page, the text doesn't wrap, requiring horizontal scrolling to read the full lines (expand details for picture):
<details>

image3
</details>

And on the governance page, only the RFC Process text doesn't fill the whole width of the page:
<details>

image2
</details>

(same on both firefox and safari, they probably use the same web engine after all)

Other than that, I'm not noticing anything problematic

samueldr:
Thanks for testing! And good catches, here's a :cookie:... But wait, there's more, those are not "old iOS" specific issues!

Good catch on the governance page: a bug with the way columns spanning more than one column-width were handled. Basically they were still limited in width even if it was not needed at that width.

As for the features page, the issue was the fresh `.columns` implementation, which didn't set a max width on the "flexed" elements. In such a layout, they're free to take as much place as needed for their child, which ended up being whatever was needed to fit the `<pre>` elements.

The existing columnar layout implementation handled that case. Though it did not handle the desire to serve the columns in a linear fashion at `-sm` sizes. So the latest change adds a variant to the non-semantic `.columnar-layout` with `.-longform`.

Infinisil:
Awesome, can confirm that fixed it!

garbas (approved):

Colecf:
This appears to have broken the css on https://search.nixos.org/options. It looks like this for me now: https://i.imgur.com/bxYUtDD.png

PR #510, opened 2020-09-02
Re-work the presentation of the main page

samueldr:
Subject of this PR: only the *presentation* of the index page. Any other page should see no changes, though if a regression compared to the main PR's state is found, it's a bug.

Note that the *text* of the main page is not subject of this PR either. I'm working on the implementation only.

* * *

## TODO

 * [x] svg-based code examples (as discussed with @garbas)
 * [x] js solution for "full screen" code examples
 * [x] make the examples section match the mockups

## Contents TODO

 * [ ] Automatically build svgs for demos
 * [ ] A sixth demo for the examples


samueldr:
@garbas the SVG file template is:

 * https://github.com/NixOS/nixos-homepage/blob/4613f04a4a7dd7817074489e3a0c8aac24c28404/demos/cover.svg?short_path=fe43942

It is *totally expected* to see a black file.

They end up being embedded as `<svg />` elements in the HTML document. This helps ensure the monospace font is made available to the svg document, but less importantly to most, we control its color scheme in tandem with the asciinema screencast's color scheme!

Hopefully you can extrapolate how it works from that document, though here's some notes:

 * Spaces are preserved, so be careful in generating the SVG document
 * *newlines* are not preserved. New lines are `<tspan>` with `dy=` set accordingly.
 * Colours are `<tspan>` with the proper class, where the number matches the escape sequence number
 * There is no support for background colours as SVG doesn't support them as text properties. Implementing this would be extremely difficult without also embedding the text outlines rather than text.

Be careful when continuing to the next line; `<tspan>` have to be valid xml, so a colour continuing to the next line has to be closed and re-opened! I think that's the main gotcha. Note that **bold** text is handled as with asciinema, with the brighter colours being bolded.

samueldr:
This is not entirely complete, there's still action points to do, but they are *content* action points, which are not part of the work on this PR.

Issue #512, opened 2020-09-02
NixCon Banner


davidak:
The NixCon organizers have launched a survey.

We should add a banner to NixOS.org to promote NixCon and the survey.

https://nixcon.limequery.org/

https://discourse.nixos.org/t/nixcon-2020-pre-event-survey/8851

PR #517, opened 2020-09-04
Add pre-nixcon survey banner

davidak:
Closes https://github.com/NixOS/nixos-homepage/issues/512

The design is not supposed to look good, just be acceptable.
We will have a new design soon.

The content should be good.
Feel free to suggest better wording!

### DIFF
--- a/index.tt
+++ b/index.tt
@@ -6,6 +6,11 @@
 
 <link rel="alternate" type="application/rss+xml" title="RSS" href="/news-rss.xml" />
 
+<div class="alert alert-info">
+  This year's <a href="https://2020.nixcon.org/">NixCon</a> will take place <b>online</b> on <b>October 16th-18th 2020</b>. It is a community-oriented conference for contributors and users of Nix and NixOS.<br />
+  Please help the organizers to plan the event by filling out <a href="https://nixcon.limequery.org/">this survey</a>!
+</div>
+
 <div class="side-by-side">
   <div class="hero">
     <h1>Reproducible builds and deployments.</h1>


addict3d:
Thank you @davidak!

index.tt
```diff
@@ -6,6 +6,11 @@
 
 <link rel="alternate" type="application/rss+xml" title="RSS" href="/news-rss.xml" />
 
+<div class="alert alert-info">
+  This years <a href="https://2020.nixcon.org/">NixCon</a> will take place <b>online</b> on <b>October 16th-18th 2020</b>. It is a community-oriented conference for contributors and users of Nix and NixOS.<br />
```
hugolgst, line 10: There is a typo on `years` :)

hugolgst (changes requested):
It would be great if you could just correct the typo.
Otherwise, great idea! :)

index.tt
davidak: you mean it should be `year's`?

hugolgst: Only 'year' because it is singular

hugolgst: Maybe you could try something like grammarly, even it isn't really privacy-friendly :/

davidak: i want to say "the conference of this year" not "in this year, the conference"
maybe you misunderstand? but my english is not great and it is totally possible that i'm wrong

hugolgst: I think that in both cases it should be "year" without a "s"

davidak: i think that changes the meaning

this dictionaries also say it's "year's"
https://www.dict.cc/?s=diesj%C3%A4hrig&failed_kw=diesj%C3%A4hrige
https://dict.leo.org/german-english/diesj%C3%A4hrig
https://en.pons.com/translate/german-english/diesj%C3%A4hrig

hugolgst: Honestly I think that you should write "This year NixCon will take place..." otherwise it's nonsense

hugolgst: It feels wrong for be because if there is "years" that's plural but it's singular in the sentence and if it is "year's" then it's "This year is NixCon will take place" with two verbs so I think that's wrong too

davidak: do you understand german?

i want to say
"Die diesjährige NixCon findet online statt"
instead of (your suggestion)
"Dieses Jahr findet die NixCon online statt"
i think that's a bit more elegant, at least in german

but let's wait for a third opinion

hugolgst: Ok, I understand. Sorry for that "year's" is correct then

garbas:
:heart:

PR #520, opened 2020-09-09
re-design: MVP contents for the last footer section

samueldr:
Better than the "TODO" previously found there.

The mockups have this widget thing (right column):

image

I, frankly, don't know how we can do anything useful with the suggestion of a square radio button (ew) and common buttons. It adds machinery that we're probably better off not having to deal with for this to work as expected.

As always, this is open to further improvements. Right now, we at least have the same spiritual use case (link to tool-specific sections).


davidak (approved):
i agree that this is an improvement over the TODO

i'm not sure if we need such marketing/sales tricks in general, but let's try

PR #521, opened 2020-09-09
re-design: Implement navbar active state

samueldr:
## Navbar active state

It was missing from its previous implementation.

This adds a bit of machinery in the layout where we are templating the links for the header.

## Navbar search

Simply done for now.


davidak (approved):
it works to show the active page

Screenshot from 2020-09-10 21-58-20

search also works

only thing not working in the manu is the _Install Nix_ button (not part of this pr, just want to mention it)

PR #522, opened 2020-09-09
re-design: MVP all (most) pages

samueldr:
Subject to review: Any page where the equivalent `.tt` source file has been edited. Any of the pages untouched, or discussed in the next section are not to be tested or reviewed yet.

* * *

We're left with:

 * learn
 * download
 * manuals
 * ~~guides~~

The learn page needs a bit more care than the basic MVP, but as stated in the last meeting, this'll be done with minimal implementation, not the full blown re-design

The download page will need even more care than the basic MVP, compared to learn. But still, not the full blown re-design yet.

The manuals should be almost ready, but I know there were some things to review.

~~(new from merge with master) guides will need a bit more work...~~

* * *

Open questions

 * Tabs styling... see the donate page. This is extremely an MVP.

### DIFF
--- a/commercial-support.tt
+++ b/commercial-support.tt
@@ -1,11 +1,14 @@
-[% WRAPPER layout.tt title="Commercial support" %]
+[% WRAPPER layout.tt title="Commercial support" handlesLayout=1 %]
 
-<h1>Commercial support</h1>
-
-<div class="lead">
-  A number of consulting companies are available for professional support.
+<div class="page-title">
+  <h1>Commercial support</h1>
 </div>
 
+<section class="generic-layout lead">
+  A number of consulting companies are available for professional support.
+</section>
+
+<section class="generic-layout">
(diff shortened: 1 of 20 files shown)


davidak:
the box don't really fit the rest of the design. it's ok for now, but we should ask the designer for a proper design of it

Screenshot from 2020-09-10 22-12-03

https://raw.githubusercontent.com/NixOS/rfcs/master/rfcs/0036-review-process.png needs a preview effect on the gov. page. it opens as new page now

i wouldn't expect the whole banner to be clickable, only the underlined text. but might be ok

Screenshot from 2020-09-10 22-19-11

samueldr:
I was actually more concerned about making the tabs work as expected (e.g. reload or anchor link opens the right tab) than the looks.

Depending on what is done for the downloads page, which will be closer to the actual redesign, the tabs looks may change.

PR #526, opened 2020-09-13
switch to asciinema-scenario

garbas:
also adds svg preview

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/526)


garbas:
I'm using here cutting edge unstable (not even in channel yet) and i'd wait until then before merging it.

PR #528, opened 2020-09-14
Remove pre-nixcon survey link. Keep nixcon banner

addict3d:
We're closing the pre-event survey.

I figure that it is good to keep the NixCon 2020 banner itself until the event is completed.

CC @davidak @garbas


davidak (approved):

davidak:
Thanks for the PR!

PR #532, opened 2020-09-16
redesign: misc. fixes for asciinema panes

samueldr:
We discussed about:

 * [x] Pausing on close
 * [x] Playing on open
 * [ ] Smooth scrolling

The first two are implemented. See the new custom event for that.

Smooth scrolling using `el.scrollIntoView({ behavior: "smooth" });` does not smooth scroll, AFAICT.

So we would be left to choose an additional JavaScript library to handle the smooth scroll, only for the pane open/close sequence. We apparently can't rely on the browser to smooth scroll the site.

My personal experience with sites that scroll smoothly like that is that it's not ever needed, and most of the time feels wrong. For something I'm not entirely sure is needed.

I'll be leaving the decision for smooth scrolling to you, and I'll implement it if desired. Though it's back at the end of the queue for priorities for the time being. It would take a bunch of time to decide on the better library to use, time I prefer to use to implement the rest of the pages as MVPs.

### DIFF
--- a/js/nixos-site.js
+++ b/js/nixos-site.js
@@ -27,6 +27,18 @@ $(function () {
     $("body").toggleClass("-debug");
   });
 
+  $(".pane-asciinemaplayer").each(function () {
+    var player = $(this).find("asciinema-player")[0];
+
+    this.addEventListener("paneOpen", function (e) {
+      player.play();
+    });
+
+    this.addEventListener("paneClose", function (e) {
+      player.pause();
+    });
+  });
+
   $("[data-fullscreen-pane]").each(function () {
     var $source = $(this);
     var $pane = $($source.attr("href"));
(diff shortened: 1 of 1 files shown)


davidak (approved):
It works to open the demo,
it starts playing and pauses when open again.
when starting another one and close it,
 the first is still at the same time paused

davidak:
I don't like that the background is just grey. It feels like it opens in a new tab and i'm not on the same page anymore.
The common and expected behavior would be to open it over the content and make a transparent layer over the content to darken it.

https://www.w3schools.com/bootstrap4/bootstrap_modal.asp

garbas (approved):
Scrolling is still missing but the rest works as expected. I'll merge it - to be able to show it at the meeting - and we can improve the asciinema player in future PRs

PR #534, opened 2020-09-16
Feature/2020 redesign improve landing page content

garbas:
This PR fixes:
- Why choose Nix or NixOS? title removed
- In "why Nix" section we should hide those gray circles until they are designed (at the end)
- remove "Install Nix” button from navbar
- one more example added (eg. NixOS test)
- replaced the lorem ipsum in other examples 
- fixed items in the footer


garbas:
against wrong branch :smile:

PR #535, opened 2020-09-16
improve content on the landing page

garbas:
This PR fixes:
- Why choose Nix or NixOS? title removed
- In "why Nix" section we should hide those gray circles until they are designed (at the end)
- remove "Install Nix” button from navbar
- one more example added (eg. NixOS test)
- replaced the lorem ipsum in other examples 
- fixed items in the footer

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/535)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,8 @@ DEMOS = \
   demos/example_2.svg \
   demos/example_3.svg \
   demos/example_4.svg \
-  demos/example_5.svg
+  demos/example_5.svg \
+  demos/example_6.svg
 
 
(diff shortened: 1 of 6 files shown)


davidak:
I think the "Why choose Nix or NixOS? title" is important to give context to the section.

Screenshot from 2020-09-16 17-54-29

It don't look that bad without the icons.

The rest is good.

garbas:
> I think the "Why choose Nix or NixOS? title" is important to give context to the section.

I think it is clear what this sections is for. Even currently we don't have a title.
Also the content of this 3 items will change once we come to that point.

PR #536, opened 2020-09-17
redesign: Add events banner (with NixCon banner)

samueldr:
I'm not positive on the wording, but the long sentence didn't feel right.

Compare to mockups:

image

Implementing the numbers on top of days for only days would be quite cumbersome. Furthermore, it's not like it's really important to have a second-by-second countdown for now.

I don't think it would be *too* time consuming to also add more than days, and to make it update, but I'm not really sure it's worth it for the time being.

Let's see your thoughts all and see how this evolves?

### DIFF
--- a/index.tt
+++ b/index.tt
@@ -150,7 +150,7 @@
     [% PROCESS demoPane demo="example_2" %]
     [% PROCESS demoPane demo="example_3" %]
 
-    <ul class="examples-list">
+    <ul class="examples-list -spotlight">
       [% WRAPPER exampleBlock demo="example_4" title="Minimal docker images" %]
       [% END %]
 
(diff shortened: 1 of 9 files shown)


davidak:
Looks OK. Maybe make a short sentence out of it.

"is online" sounds like some content release. better use something like "takes place online"

the countdown will probably use JS and you probably can find some code on stackoverflow that is fine for now. but what about people that don't have JS enabled? having the day at least update daily would be nice

domenkozar:
cc @anamarijas for design review: https://5f6371fc4f81460f301e5f8c--nixos-homepage.netlify.app

garbas:
@domenkozar better to use #506 for the review.

samueldr:
Thank you @garbas for taking the time to improve what I literally threw together in the quickest manner as a last task :).

PR #538, opened 2020-09-17
Re-do "improve links"

samueldr:
I wasn't trying to be flippant in a66b8e1d96621d7cc2257bc73452201e04c35115, it causes misc. issues around the site as this adds borders to links *wholesale*.

Some links are "structural" in the site; they are `display: block` to fill their parent element to make a bigger touchable target.

The best example of an issue this change causes is in the tabs navigation:

image

This is not the only issue this causes, but this is the most obvious.

* * *

What is the desired outcome of replacing the font-engine-native underline with a border? I don't see how this is improved.

The font-engine-native underline does things a border cannot do, like not striking through descenders.

image

While a border can do things the font-engine-native underline cannot do, but here it's not being done. Things like using a dotted or dashed underline, or changing the colour or size independently.

* * *

In a discussion with @garbas, I learned the reasoning behind this change.

It seems that user-experience research at mozilla found out that font-engine-native underlines make things hard to read, compared to the more spaced-out lines the borders allow.

I am personally somewhat doubting this until I can see research, but still can see how it can be a statistically-proven truth.

Thus, we probably can do this, but there are issues this causes. The main problem this causes is that at any place in the site a link is used as structural element, it needs to also gain the proper rules to prevent its border to be used. It also causes issues in that a "structural" element link can't gain a "proper" underline.

This sounds like something that shouldn't be pushed down onto the web developers to fix, but it's probably a much more complicated issue because of the incredible amount of baggage web design work is built on.

### DIFF
--- a/site-styles/components/footer.less
+++ b/site-styles/components/footer.less
@@ -33,8 +33,10 @@ body > footer {
 
 	a {
 		text-decoration: none;
+		border-bottom: 0;
 		&:hover, &:focus, &:active {
 			text-decoration: none;
+		  border-bottom: 0;
 		}
 	}
 
--- a/site-styles/components/navbar.less
+++ b/site-styles/components/navbar.less
@@ -176,6 +176,7 @@ body > header {
 	a {
 		white-space: nowrap;
 		text-decoration: none;
+		border-bottom: 0;
 		padding-left:  15*@unit;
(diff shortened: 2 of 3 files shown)


samueldr:
:/ this wasn't ready to merge, since there were the bugs to fix still. I'll do that on the main PR I guess.

PR #547, opened 2020-09-21
Don't break '60 000'

raboof:
It looks really weird when the line is broken in the middle of '60 000'


garbas:
@raboof Thank you! PR is even better then an open issue :smile:

PR #549, opened 2020-09-21
on mobile page have less margin-bottom between page-title and content

garbas:
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/549)


samueldr:
Literally only changed the redundant "times one", but that wouldn't have been the end of the world if it shipped with it.

Issue #446, opened 2020-06-04
Blog / News section


garbas:
- aggregate categories of teams on discourse
- integrate planet.nixos.org

Profpatsch:
I’m quoting your thoughts from https://github.com/NixOS/nixos-homepage/pull/552#issuecomment-697167063 @garbas:

> Here is a mix of my thoughts on the news section in general:
> 
>     * I personally find it completely OK if news from 10 years ago has broken links. I use [Wayback Machine](https://addons.mozilla.org/en-US/firefox/addon/wayback-machine_new/) add-on which fixed this for me most of the time.
> 
>     * News section was taken from the landing page because existing users are not primary audience of the landing page. But there is a section on the landing page were we can show one important announcement, like we already have for NixCon or like we will have when new Nix/NixOS version is released. So in reality is not really gone from the landing page just reduced to one item.
> 
>     * There is a plan that each release version would get a dedicated page, linking to only, well mostly, to release.nixos.org, which is our immutable storage - out archive.org.
> 
>     * It is important to define the purpose of the news section of the site. I would refer here to the audience and fuel the purpose from there. Main audience of the website are new nix users and they "story" could be something like this:
>       
>       1. Just heard of Nix, went to nixos.org and got excited because of the examples and benefits it brings - Landing and Features page
>       2. Downloaded it for my system / cloud (Download page)
>       3. Need to learn more about Nix, I hope there are some tutorials about Nix for X language - Learn page
>       4. Everything is working except one little thing is missing how do I contribute back, oh is just a PR? nice and simple - Community page -> pointing to How to contribute tutorial
>       5. I'm sold on Nix I want to get involved into the community and help as much as I can (News/Blog)
>       
>       As you can see each of above points happen in some sort of order and each of above points correlates to some page. You can also see that we try this order of pages in the navigation bar (Community, Governance, Donate pages are going to be merged soon, once we have other pages to take their place).
> 
> 
> The purpose of the news/blog section is then to show the "internals" of the community, trying to answer things such are "What are others working on?", "What is happening in marketing team?", "What is coming next in Nix?", "Which RFCs are being reviewed/discussed?", ...
> 
> This works is being tracked under #446
> 
> Regardless of my comments I think this PR improves on the current state, until we actually implement #446 (expecting this to happen this year).

Profpatsch:
[...]

Issue #397, opened 2020-04-18
Consolidate NixCon pages


davidak:
# Original state

We had 4 official NixCons and their pages are on 4 different domains:

- conf.nixos.org
- nixcon2017.org
- nixcon2018.org
- 2019.nixcon.org

Can we get access to nixcon.org?

It would be great to have:

- https://nixcon.org/ (announcement of and link to next planned conference, links to previous)
- https://nixcon.org/2020 (with simple announcement "a conference is planned. help \<link to discourse thread\>")
- https://nixcon.org/2019
- https://nixcon.org/2018
- https://nixcon.org/2017
- https://nixcon.org/2015

See https://fosdem.org/2020/ for example.

Maintaining the domain and webspace is probably a task for the infrastructure team. The conference organizers have to upload the pages (maybe in a git repo?)

garbas:
pinging current nixcon team members /cc @andir @zimbatm @flokli

flokli:
Yeah, the current situation is a bit chaotic - so far, we repurposed the old website for the next years nixcon, by forking and editing it ;-)

Moving everything below specific subpaths might introduce too much churn in updating references - so I'd prefer a subdomain for each year. This would also allow to change ways how we host things (GitHub Pages vs. another service vs. hosting by ourselves).

zimbatm:
Curently we have one sub-domain per year. The advantage with that approach is that each year is completely independent and can use a different static generate and template.

eg: http://2019.nixcon.org/

You can find all the sites at https://github.com/nixcon/

davidak:
>so far, we repurposed the old website for the next years nixcon, by forking and editing it ;-)

that's totally fine. FOSDEM seems to do that too and they are a lot bigger and more popular.

>Moving everything below specific subpaths might introduce too much churn in updating references - so I'd prefer a subdomain for each year.

theoretically it should just be 1 sed command, right? i think we should invest time here when it improves visitors experience. i think a subpath is better symantically. NixCon 2020 has the URL https://nixcon.org/2020 (you can read it from left to right. and search engines probably count it as one site and get a better rank, not that we need more participants :D)

>This would also allow to change ways how we host things (GitHub Pages vs. another service vs. hosting by ourselves).

that is a point, but is that really a problem we need to solve? we can easily host all static sites in one webspace at one hoster or migrate them to a new one. actually any webhoster should work here

>The advantage with that approach is that each year is completely independent and can use a different static generate and template.

[...]

PR #552, opened 2020-09-21
news.xml: fix links as far as possible

Profpatsch:
In order to honor the history of Nix(OS), this is an effort to restore
the announcement page to former glory. It is really quite fascinating,
how basically every link before 2010 is broken by now.

Some links are unfortunately beyond rescue, including University Delft
websites that were never archived and now point to a Github 404 …

Things left to do:
- Change all pdfs to archive links. Eelco’s papers still work, but
  they shouldn’t change and the links will break eventually. Better
  safe than sorry.
- Fix the release pages. From around 2010, a grave sin was commited
  and most links point directly to build outputs of hydra.nixos.org,
  which are obviously all broken by now. I found that most should be
  easy to recover by pointing them to `releases.nixos.org`. Others
  might be lost.
- Think about how to link to historical release documents that use
  relative links right now. The sections might be restructured in the
  future, breaking the anchors.
- Archive videos of talks we point to. At least one is on Youtube.

Most external historical links should be back to “working”, for
various definitions of working.

### DIFF
--- a/news.xml
+++ b/news.xml
@@ -291,7 +291,7 @@
       We're happy to announce that <strong>NixCon 2017</strong>, the
       second Nix Conference, will take place <strong>October 28–31 2017 in Munich</strong>
       For more information, see the
-      <a href="http://nixcon2017.org/">NixCon 2017 website</a>.
+      <a href="https://web.archive.org/web/20200816233409/http://nixcon2017.org/">NixCon 2017 website</a>.
       And please consider
       <a href="https://schedule.nixcon2017.org/en/nixcon2017/cfp/">submitting a talk</a>!
     </description>
@@ -337,7 +337,7 @@
     <description>
       <a href="/releases/nixops/nixops-1.4/nixops-1.4.tar.bz2">NixOps
         1.4</a> has been released. This release contains contains many
-      nice new features. See the <a href="https://nixos.org/nixops/manual#ssec-relnotes-1.4">manual</a>
+      nice new features. See the <a href="https://web.archive.org/web/20200613150138/https://hydra.nixos.org/build/115931128/download/1/manual/manual.html#ssec-relnotes-1.4">manual</a>
       for details.
     </description>
   </item>
@@ -404,12 +404,12 @@
       NixCon 2015
     </title>
     <description>
-      <a href="http://conf.nixos.org"><img class="inline" style="width: 10em;"
-          src="https://d2z6c3c3r6k4bx.cloudfront.net/uploads/event/logo/1005856/banner.png" alt="NixCon logo" /></a>
+      <a href="https://web.archive.org/web/20200210001459/http://conf.nixos.org/"><img class="inline" style="width: 10em;"
+          src="https://web.archive.org/web/20180921172104if_/https://d2z6c3c3r6k4bx.cloudfront.net/uploads/event/logo/1005856/banner.png" alt="NixCon logo" /></a>
       We're happy to announce that <strong>NixCon 2015</strong>, the
       first Nix Conference, will take place on <strong>November
         14—15th 2015 in Berlin</strong>. For more information, see the
-      <a href="http://conf.nixos.org">NixCon website</a>. And please
+      <a href="https://web.archive.org/web/20200210001459/http://conf.nixos.org/">NixCon website</a>. And please
       consider <a href="http://conf.nixos.org/submit-a-talk.html">submitting a
         talk</a>!
     </description>
@@ -478,7 +478,7 @@
       We’re having a NixOS sprint at the <a href="https://www.kiberpipa.org/en/">Kiberpipa hackerspace</a>
       in Ljubljana, Slovenia, on <strong>August
         23—27</strong>. Joining is free! For more information and to
-      register, please go to the <a href="http://www.kiberpipa.org/nixos-sprint-ljubljana-2014/">sprint
+      register, please go to the <a href="https://web.archive.org/web/20150623082420/http://www.kiberpipa.org/nixos-sprint-ljubljana-2014/">sprint
         page</a>.
     </description>
   </item>
@@ -508,7 +508,7 @@
     <description>
       <a href="https://hydra.nixos.org/release/nixops/nixops-1.2">NixOps
         1.2</a> has been released. This release contains contains many nice new features. See the <a
-        href="https://nixos.org/nixops/manual#ssec-relnotes-1.2">manual</a>
+        href="https://web.archive.org/web/20200613150138/https://hydra.nixos.org/build/115931128/download/1/manual/manual.html#ssec-relnotes-1.2">manual</a>
       for details.
     </description>
   </item>
@@ -520,7 +520,7 @@
     </title>
     <description>
       <a href="https://hydra.nixos.org/release/nix/nix-1.7">Nix 1.7</a>
-      has been released. See the <a href="https://nixos.org/releases/nix/nix-1.7/manual/#ssec-relnotes-1.7">release
+      has been released. See the <a href="https://web.archive.org/web/20170723084043/https://releases.nixos.org/nix/nix-1.7/manual/#ssec-relnotes-1.7">release
         notes</a> for a list of new features.
     </description>
   </item>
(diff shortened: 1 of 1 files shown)


news.xml
```diff
@@ -291,7 +291,7 @@
       We're happy to announce that <strong>NixCon 2017</strong>, the
       second Nix Conference, will take place <strong>October 28–31 2017 in Munich</strong>
       For more information, see the
-      <a href="http://nixcon2017.org/">NixCon 2017 website</a>.
```
davidak, line 294: http://nixcon2017.org/ still works for me

news.xml
```diff
@@ -404,12 +404,12 @@
...
     </title>
     <description>
-      <a href="http://conf.nixos.org"><img class="inline" style="width: 10em;"
-          src="https://d2z6c3c3r6k4bx.cloudfront.net/uploads/event/logo/1005856/banner.png" alt="NixCon logo" /></a>
+      <a href="https://web.archive.org/web/20200210001459/http://conf.nixos.org/"><img class="inline" style="width: 10em;"
+          src="https://web.archive.org/web/20180921172104if_/https://d2z6c3c3r6k4bx.cloudfront.net/uploads/event/logo/1005856/banner.png" alt="NixCon logo" /></a>
       We're happy to announce that <strong>NixCon 2015</strong>, the
       first Nix Conference, will take place on <strong>November
         14—15th 2015 in Berlin</strong>. For more information, see the
-      <a href="http://conf.nixos.org">NixCon website</a>. And please
```
davidak, line 412: http://conf.nixos.org/ still works

news.xml
```diff
@@ -404,12 +404,12 @@
       NixCon 2015
     </title>
     <description>
-      <a href="http://conf.nixos.org"><img class="inline" style="width: 10em;"
```
davidak, line 407: http://conf.nixos.org/ still works

news.xml
```diff
@@ -556,10 +556,10 @@ $ nix-store -qR /run/current-system | grep openssl
     </title>
     <description>
       <a href="https://www.domenkozar.com/">Domen Kožar</a> gave <a
-        href="https://fosdem.org/2014/schedule/event/nixos_declarative_configuration_linux_distribution/">a
```
davidak, line 559: use official archive https://archive.fosdem.org/2014/schedule/event/nixos_declarative_configuration_linux_distribution/

news.xml
```diff
@@ -556,10 +556,10 @@ $ nix-store -qR /run/current-system | grep openssl
...
+        href="https://web.archive.org/web/20200424105824/https://archive.fosdem.org/2014/schedule/event/nixos_declarative_configuration_linux_distribution/">a
         talk at FOSDEM about NixOS</a> (<a
         href="https://video.fosdem.org/2014/H1309_Van_Rijn/Saturday/NixOS_declarative_configuration_Linux_distribution.webm">video</a>).
-      Also, Ludovic Courtès gave <a href="https://fosdem.org/2014/schedule/event/gnuguix/">a talk on
```
davidak, line 562: https://archive.fosdem.org/2014/schedule/event/gnuguix/

news.xml
```diff
@@ -590,7 +590,7 @@ $ nix-store -qR /run/current-system | grep openssl
       potentially destabilising changes that sometimes occur on the
       unstable branch. You can get NixOS 13.10 ISOs and VirtualBox
       appliances from the <a href="/download.html">download
-        page</a>. See the <a href="https://nixos.org/nix-dev/2013-October/011941.html">announcement</a>
```
davidak, line 593: it redirects to https://releases.nixos.org/nix-dev/2013-October/011941.html

not sure if that is a stable url

news.xml
```diff
@@ -607,7 +607,7 @@ $ nix-store -qR /run/current-system | grep openssl
       <a href="https://hydra.nixos.org/release/nix/nix-1.6.1">Nix
         1.6.1</a> has been released. This is primarily a bug fix
       release but has some minor new features. See the <a
-        href="https://nixos.org/manual/nix/stable/#ssec-relnotes-1.6.1">release
```
davidak, line 610: the link still works

why have you replaced it?

news.xml
```diff
@@ -674,7 +674,7 @@ $ nix-store -qR /run/current-system | grep openssl
     </title>
     <description>
       A sprint focused on NixOS and <a href="http://kotti.pylonsproject.org/">Kotti</a> will be held <a
-        href="http://www.coactivate.org/projects/zidanca-sprint-2013/project-home">22-26
```
davidak, line 677: the link still works

davidak (changes requested):
many links don't need to get changed because they still work

Profpatsch:
> many links don't need to get changed because they still work

I applied some reason to which links to archive. In general, pages that are static websites which won’t change in the future (e.g. the nixcon 2017 site) we can archive right now, since the only change to it in the future will be the link breaking.

news.xml
Profpatsch: static site that won’t change in the future.

news.xml
Profpatsch: static site that won’t change

news.xml
Profpatsch: static site that won’t change

news.xml
Profpatsch: static site that won’t change

news.xml
Profpatsch: static site that won’t change

news.xml
Profpatsch: static site that won’t change

davidak:
i'm not sure if that makes sense. i wouldn't do that on my personal page

i don't think the fosdem page will break and we should make sure the nixcon pages also don't break for historic reasons. i would like to have them on one domain to make it more consistent https://github.com/NixOS/nixos-homepage/issues/397

Profpatsch:
> i don't think the fosdem page will break and we should make sure the nixcon pages also don't break for historic reasons. i would like to have them on one domain to make it more consistent #397

Once we get them all to stable subdomains, we can switch the links back, yes.

Profpatsch:
The news page is really a snapshot of the state at the point of writing the news, so I’d argue to snapshot the pages at the state they were in when a news item is released. The links are going to rot, the pages are going to change (or even change into advertising for sunglasses, as happened with one of the ACM Sigplan domains …).

davidak:
that makes sense

garbas:
Here is a mix of my thoughts on the news section in general:

- I personally find it completely OK if news from 10 years ago has broken links. I use [Wayback Machine](https://addons.mozilla.org/en-US/firefox/addon/wayback-machine_new/) add-on which fixed this for me most of the time.

- News section was taken from the landing page because existing users are not primary audience of the landing page. But there is a section on the landing page were we can show one important announcement, like we already have for NixCon or like we will have when new Nix/NixOS version is released. So in reality is not really gone from the landing page just reduced to one item.

- There is a plan that each release version would get a dedicated page, linking to only, well mostly, to release.nixos.org, which is our immutable storage - out archive.org.

- It is important to define the purpose of the news section of the site. I would refer here to the audience and fuel the purpose from there. Main audience of the website are new nix users and they "story" could be something like this:

  1. Just heard of Nix, went to nixos.org and got excited because of the examples and benefits it brings - Landing and Features page
  2. Downloaded it for my system / cloud (Download page)
  3. Need to learn more about Nix, I hope there are some tutorials about Nix for X language - Learn page
  4. Everything is working except one little thing is missing how do I contribute back, oh is just a PR? nice and simple - Community page -> pointing to How to contribute tutorial
  5. I'm sold on Nix I want to get involved into the community and help as much as I can (News/Blog)

  As you can see each of above points happen in some sort of order and each of above points correlates to some page. You can also see that we try this order of pages in the navigation bar (Community, Governance, Donate pages are going to be merged soon, once we have other pages to take their place).

 The purpose of the news/blog section is then to show the "internals" of the community, trying to answer things such are "What are others working on?", "What is happening in marketing team?", "What is coming next in Nix?", "Which RFCs are being reviewed/discussed?", ...

  This works is being tracked under https://github.com/NixOS/nixos-homepage/issues/446

Regardless of my comments I think this PR improves on the current state, until we actually implement #446 (expecting this to happen this year).

garbas (approved):
:+1:

PR #566, opened 2020-09-22
Replace default.nix by shell.nix in cover screencast.

maralorn:
Sorry missed that one before.


garbas:
@maralorn thank you! step by step and we'll get there.

PR #570, opened 2020-09-24
Learn: replace 'More ...' with 'Manual'

raboof:
I would actually prefer 'Nix Manual', 'NixOS Manual' and 'Nixpkgs Manual',
but those documents actually have different titles.

Taking one step at a time, I think changing 'More ...' to 'Manual' is a
good first action :)

### DIFF
--- a/learn.tt
+++ b/learn.tt
@@ -68,7 +68,7 @@
           </ul>
         </li>
       </ul>
-      <a href="[%root%]manual/nix/stable" class="button">More ...</a>
+      <a href="[%root%]manual/nix/stable" class="button">Manual</a>
     </li>
 
     <li>
@@ -143,7 +143,7 @@
         </li>
         <li><a href="[%root%]manual/nixpkgs/stable/#chap-submitting-changes">Contributing to Nixpkgs</a></li>
       </ul>
-      <a href="[%root%]manual/nixpkgs/stable" class="button">More ...</a>
+      <a href="[%root%]manual/nixpkgs/stable" class="button">Manual</a>
     </li>
 
     <li>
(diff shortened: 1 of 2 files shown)


maralorn:
I think something along this lines would be nice.

Maybe even "The whole Nix Manual", ... ?

My point would be: When I look for documentation for a project I really want to understand what documents there are is it one wiki? one userguide and one reference? A loose collection of blog posts? Not knowing the overarching simple structure decreases discoverability and faith in finding information a lot.

So in this case we should really communicate "There are exactly three manuals and those useful links we have collected for you all point to one of them.“

The links _are_ great, but I would really love something to make clear that there are this three documents.

Maybe a small subtitle under the headding:

“Learn Nix 
from the Nix manual"

where "Nix manual" is a link?

raboof:
> Maybe a small subtitle under the heading:

Personally, I mostly think of the manuals as reference manuals: great to look things up when you already know your way around the system, but a bit daunting if you're just getting started.

From that perspective I like having the links to more 'tutorial-like' resources first, and the manual as a (prominent) button at the bottom of each column.

garbas:
@raboof @maralorn Thank you for looking into this. I think you are right that there should be the manual in the name, but I would change the title - eg. Learn Nix -> Nix manual - and only change the bottom button from "More..." to "Read more...".

raboof:
> I would change the title - eg. Learn Nix -> Nix manual - and only change the bottom button from "More..." to "Read more..."

I suspect that would be confusing: that gives the impression that that section *is* the manual, rather than just linking to it...

garbas:
@raboof I added the changes to this PR, I think it makes it makes it even more obvious now.

maralorn:
Great!

I hope that at the end of the whole process we have a solution for the very unfortunate situation that two of the links in the former "Learn nix" section actually point to the nixpkgs manual.

PR #575, opened 2020-09-24
Use search icon instead of search input box in the navbar

### LINKED ISSUE #545: No search at the top on smaller widths
Even just splitting my browser to half of my 16:9 1080p monitor is small enough that the search bar disappears. And, of course, it's not there on mobile. As a new user of NixOS, I am constantly going to the website for the manuals and to search for packages/options. I really dislike having to scroll or resize my window in order to search. I suggest something like the old website on smaller screens - a clickable search icon.

garbas:
fixes #545

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/575)

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -50,11 +50,6 @@ BLOCK navigationLink %]
   <header>
     <div>
       <h1><a href="[% root %]">NixOS</a></h1>
-      <form class="header-search" method="get" action="https://search.nixos.org/packages">
-        <div class="form-input">
-          <input type="text" placeholder="Search for Nix packages" name="query" />
-        </div>
-      </form>
       [%# ???! go read `navbar.less` and `nixos-site.js`, it's all fine I swear. %]
       <nav style="display: none;">
         <ul>
@@ -66,6 +61,7 @@ BLOCK navigationLink %]
           [% INCLUDE navigationLink href="donate.html" title="Donate" %]
         </ul>
       </nav>
+      <a class="header-search-link" href="https://search.nixos.org/packages"></a>
     </div>
   </header>
 
--- a/layout.tt
+++ b/layout.tt
@@ -59,8 +59,8 @@ BLOCK navigationLink %]
           [% INCLUDE navigationLink href="community.html" title="Community" %]
           [% INCLUDE navigationLink href="governance.html" title="Governance" %]
           [% INCLUDE navigationLink href="donate.html" title="Donate" %]
-          [% WRAPPER navigationLink class="search" title="Donate" %]
-            <a href="https://search.nixos.org/packages">Search</a>
+          [% WRAPPER navigationLink class="search" title="Search" %]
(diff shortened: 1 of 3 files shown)


garbas:
@samueldr I'm not so sure I did a good job with the flex box, but I got it working on the end. I think you should look at this to make sure I'm not complicating things unnecessary.

samueldr:
*Totally not reviewing the important bits first.*

When working with the DOM, we are using `RSCSS` to help keep class names coherent.

 * https://github.com/NixOS/nixos-homepage/blob/master/site-styles/README.md#how-this-is-laid-out

With this one notable addendum:

> With one addendum that for qualified layout elements like `body > header` the component will not be named (e.g. `page-header`); the qualified name for the innate structure of the document is enough.

It's a bit hard sometimes to distinguish what is a *component* in our setup.

 * https://rscss.io/components.html

I would argue that, here, the component is the `header nav` (because of that addendum). So we're now dealing with *elements*.

 * https://rscss.io/elements.html

Thus the element would be `search` or `searchitem` or `searchlink`.

In the next, and upcoming comment: actually reviewing the meat of the change.

samueldr:
Some misc. tidbits:

We should remove unused styles, to prevent rot. If unused, but the intent is to re-use it soon, adding a comment stating so would be helpful. (Please call me out on it if I do it!)

No *actionable* elements should have an empty body! Right now this link is just empty if there is a lack of styles, or a lack of sight!

* * *

It seems there was an issue with image replacement stemming form the way LESS applies `:extend`. I fixed it.

This allows us to define the image replacement strictly for the proper media queries.

But why is the `#svg` definition applied outside the media query? For the same reason! Though here we **definitely** don't want to make `#svg` mixins that copy the whole SVG for each elements. This is not an issue because the `:after` pseudo-element won't show up in narrow viewports.

* * *

Finally, it'd be better to include this as part of the main navigation. It's weird to have an option that's not available for *narrow viewports*.

layout.tt
```diff
@@ -64,6 +59,9 @@ BLOCK navigationLink %]
           [% INCLUDE navigationLink href="community.html" title="Community" %]
           [% INCLUDE navigationLink href="governance.html" title="Governance" %]
           [% INCLUDE navigationLink href="donate.html" title="Donate" %]
+          [% WRAPPER navigationLink class="search" title="Donate" %]
```
davidak, line 62: `class="search" title="Donate"` that seem wrong

davidak (changes requested):
but it works

layout.tt
garbas: nice catch, fixed

garbas:
@samueldr thank you for all the explanation. Agree with your decisions on every point.

layout.tt
samueldr: oops!

PR #584, opened 2020-09-27
Add instructions to build in Docker for Mac

### LINKED ISSUE #515: error building homepage on macos
Following the instructions on https://github.com/NixOS/nixos-homepage to build the homepage as a novice Nix user I get the following error:
```
$ git clone https://github.com/bxrt/nixos-homepage.git
Cloning into 'nixos-homepage'...
remote: Enumerating objects: 425, done.
remote: Counting objects: 100% (425/425), done.
remote: Compressing objects: 100% (247/247), done.
remote: Total 5450 (delta 242), reused 340 (delta 176), pack-reused 5025
Receiving objects: 100% (5450/5450), 17.09 MiB | 8.01 MiB/s, done.
Resolving deltas: 100% (3329/3329), done.
$ cd nixos-homepage/
$ nix-shell
unpacking 'https://github.com/edolstra/flake-compat/archive/master.tar.gz'...
error: nix-shell requires a single derivation
Try 'nix-shell --help' for more information.
```
This is on macOS.

bxrt:
Resolves #515

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ To build it on Linux/NixOS:
     [nix-shell]$ make
     [nix-shell]$ python -m http.server
 
-then open http://127.0.0.1:8000/index.html
+then open http://127.0.0.1:8000
 
 
 To automatically rebuild on every change:
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To test the complete result from a nix-build:
     $ nix-build
     $ nix-shell --run 'cd result/ && python -m http.server'
 
-### Docker Desktop for Mac
+### MacOS (via Docker)
 
 Unfortunately it will not build on macOS natively. 
 But it can be build in a [Docker container on macOS](https://www.docker.com/products/docker-desktop). 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Add the `--bind 0.0.0.0` argument to bind the web server with the container's ex
 
     [nix-shell]$ python -m http.server --bind 0.0.0.0
 
-Then open http://127.0.0.1:8000/index.html in your browser on macOS.
+Then open http://127.0.0.1:8000 in your browser on macOS.
 
 A single terminal screen is inconvenient for serious development, so you may want to set up openssh on the container to access the sources with the IDE on your Mac.
 Alternatively you can set up something like [tmux](https://github.com/tmux/tmux/) to allow for multiple terminals into the container. Replace
--- a/README.md
+++ b/README.md
@@ -1,16 +1,20 @@
 NixOS.org
 =========
 
-This repository contains the sources of the `nixos.org` website.  To
-build it:
+This repository contains the sources of the `nixos.org` website.
 
-    $ git clone git@github.com:NixOS/nixos-homepage.git
+### Linux/NixOS
+
+To build it on Linux/NixOS:
+
+    $ git clone https://github.com/NixOS/nixos-homepage.git
     $ cd nixos-homepage
     $ nix-shell
     [nix-shell]$ make
     [nix-shell]$ python -m http.server
 
 then open http://127.0.0.1:8000/index.html
+
 
 To automatically rebuild on every change:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,17 +1,21 @@
...
+
+To build it on Linux/NixOS:
+
+    $ git clone https://github.com/NixOS/nixos-homepage.git
     $ cd nixos-homepage
     $ nix-shell
     [nix-shell]$ make
     [nix-shell]$ python -m http.server
 
 then open http://127.0.0.1:8000/index.html
```
garbas, line 16:
```suggestion
then open http://127.0.0.1:8000
```

README.md
```diff
@@ -21,6 +25,40 @@ To test the complete result from a nix-build:
     $ nix-build
     $ nix-shell --run 'cd result/ && python -m http.server'
 
+### Docker Desktop for Mac
```
garbas, line 28:
```suggestion
### MacOS (via Docker)
```

README.md
```diff
@@ -21,6 +25,40 @@ To test the complete result from a nix-build:
...
+
+Unfortunately it will not build on macOS natively. 
+But it can be build in a [Docker container on macOS](https://www.docker.com/products/docker-desktop). 
+(You may need to increase the maximum "Disk image size" in Docker's Resources Preferences.)
+
+In your favourite macOS terminal:
+
+    $ docker run -p 8000:8000 -it --name nixos-homepage nixos/nix
```
garbas, line 36:
```suggestion
    $ git clone https://github.com/NixOS/nixos-homepage.git
    $ cd nixos-homepage
    $ docker run -v $PWD:/work -w /work -p 8000:8000 -it --name nixos-homepage nixos/nix nix-shell
```

It would be better if git is used from the operating system. I think we can expect this to be installed on macos. I realize that then we have to fix `default.nix` and `shell.nix` a bit (which I already did).

This would avoid running nix-shell in nix-shell, since `fd` and `entr` are already present in development environment of `nixos-homepage`

README.md
```diff
@@ -21,6 +25,40 @@ To test the complete result from a nix-build:
...
+    [nix-shell]$ git clone https://github.com/NixOS/nixos-homepage.git
+    [nix-shell]$ cd nixos-homepage
+    [nix-shell]$ nix-shell
+    [nix-shell]$ make
+
+Add the `--bind 0.0.0.0` argument to bind the web server with the container's external interface too:
+
+    [nix-shell]$ python -m http.server --bind 0.0.0.0
+
+Then open http://127.0.0.1:8000/index.html in your browser on macOS.
```
garbas, line 53:
```suggestion
Then open http://127.0.0.1:8000 in your browser on macOS.
```

README.md
```diff
@@ -21,6 +25,40 @@ To test the complete result from a nix-build:
...
+Following with the same steps as on Linux/NixOS:
+
+    [nix-shell]$ git clone https://github.com/NixOS/nixos-homepage.git
+    [nix-shell]$ cd nixos-homepage
+    [nix-shell]$ nix-shell
+    [nix-shell]$ make
+
+Add the `--bind 0.0.0.0` argument to bind the web server with the container's external interface too:
+
+    [nix-shell]$ python -m http.server --bind 0.0.0.0
```
garbas, line 51: I think we could also use some other way to serve from the host environment (not from within docker) to serve a folder. Maybe simply running `nix-shell -p python3 --run "python -m http.server"` outside the container is better.

README.md
```diff
@@ -21,6 +25,40 @@ To test the complete result from a nix-build:
...
+
+Then open http://127.0.0.1:8000/index.html in your browser on macOS.
+
+A single terminal screen is inconvenient for serious development, so you may want to set up openssh on the container to access the sources with the IDE on your Mac.
```
garbas, line 55: I would skip this section and try to provide instructions to only run one instance of container with running `nix-shell -p` outside the container.

garbas (changes requested):
@bxrt :heart: 

I added few things which I hope will make the instructions even easier.

README.md
bxrt: @garbas, thanks. My change puts everything in the container. I agree that it is nicer to have the sources on the macos-side and mount those as a volume in the container as you suggest. I tried this and the `nix-shell` command is successful. So that's great. However, `make` then fails in the container:

```
[...]
nix-build -A packages.x86_64-linux.siteStyles --out-link styles
error: creating symlink from '/work/styles.tmp-16815-189614047' to '/nix/store/d9f5ppqpx1p6cswpg3wlzy5dwy7sdwi5-nixos-homepage-styles': Operation not permitted
make: *** [Makefile:199: styles] Error 1
```
I don't know how to fix this.

README.md
bxrt: @garbas, this is a great idea if we get the `make` issue fixed, as long as there is content dynamically generated in the backend. I experimented with your suggestion and I believe the interfaces ip address must be specified within the nix-shell, so: `nix-shell -p python3 --run "python -m http.server --bind 127.0.0.1"`(and the `-p 8000:8000` argument must be removed from the docker invocation).

README.md
bxrt: Actually, I do have ideas on how to fix this, but that involves changing how the build is done and that goes beyond the scope of documenting it in Docker on Mac.

garbas: Is this maybe because the link already existed because you already tried building it outside docker?

bxrt: No, it is because symlinks between container and volumes mounted on a host are unsupported by Docker.

lockejan:
Hi, what requested changes are missing here? I think this would make contributing a lot easier for new mac users.
I'd be happy to help resolving this.

garbas:
@lockejan (or anybody else using macOS) could you test the instructions in this PR and submit new PR?

PR #586, opened 2020-09-28
fixing broken preview video link

garbas:
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/586)


garbas:
@samueldr This fixed the problem but I'm not sure this is the correct way. But since this was quite a big problem I merged it ASAP. We can fix the solution in followup PR.

PR #587, opened 2020-09-28
index: Rename "isolated" to "declarative" example

andir:
As there is no isolation and all of these shells are still able to wipe
all of your data we should probably not call them "isolated". They are
probably better described as `declarative` as that is what is happening
here.

### DIFF
--- a/index.tt
+++ b/index.tt
@@ -141,7 +141,7 @@
       [% WRAPPER exampleBlock demo="example_2" title="Multiple languages, one tool" %]
       [% END %]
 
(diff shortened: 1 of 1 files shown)


garbas:
I think this is an improvement but it still can be improved (in later PRs).

Nobody uses Nix because it is declarative, but because there is an important benefit of _"declarativness"_. To newcomers this benefit might not be obvious and might be even misunderstood as "unneeded complexity". I think it would be much better to just talk in terms of benefits of features and avoid being misunderstood.

But right now I don't have a better idea. @andir maybe you have some.

PR #593, opened 2020-09-29
Learn page redesign

garbas:
## Design

- [Learn_deskop_2.pdf](https://github.com/NixOS/nixos-artwork/blob/master/www/2020_redesign/Learn/Learn_deskop_2.pdf)
- [Learn_mobile.pdf](https://github.com/NixOS/nixos-artwork/blob/master/www/2020_redesign/Learn/Learn_mobile.pdf)

## TODO

- [x] design get started section (@garbas)
- [x] extract icons for get started section (@samueldr)
- [x] use search boxes (instead of examples) for second section, for now (@garbas)
- [x] design guides section, list of tutorials with number of estimated minutes to read (@garbas)
- [x] design manuals section (@garbas)
- [x] extract manual icon (@samueldr)
- [x] design other resources section (@garbas)
- [x] check and create more todos for actual tutorials and manuals (@garbas)

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/593)


andir:
This move yet again further away from NixOS and only focuses on Nix. Is that really the goal?
Maybe we should have a different section there where the users can decide what direction they want to go?

garbas:
@andir Yes, currently we only have tutorials/guides around Nix, but I hope to get much more regarding NixOS. But we do intend to focus guides and tutorials more around devops (building different deployment images for services) then around linux desktop environment.

samueldr:
@andir I've opened #633 so we can centralize discussion about the *distribution* content.

And yes, for the main "property" online, it was the goal to focus on the Nix tool rather than the distribution.

Not from previous discussions with the "marketing" team, but from thinking it afresh, I realise now that it was problematic how Nixpkgs and Nix were stuffed deep into the website of the distribution. I've been into situations where a proper Nix, the ecosystem and the Language website would have helped "sell" Nix.

Inverting this is probably a good idea, but we'll have to figure out a proper pattern to still have a proper "display case" for the distribution. Yes, this is moving without having a plan for that really, but the existing "focus" is so unfocused it's not really losing much the way it's moving.

davidak:
What is the purpose of the rocket icon there?

Screenshot from 2020-10-26 23-08-22

It would help to guide the viewer if it was different icons.

---

I can't read the text in the text box

Screenshot from 2020-10-26 23-10-09


---

where do you want to guide the viewer to first? right now both are equally attention seeking. having the same icon don't help

Screenshot from 2020-10-26 23-11-23

the buttons don't have links

---

except that, it looks good. well done

davidak (approved):
looks good now :sparkles:

samueldr (approved):

andir:
Looking at the preview deployments it looks nicer than what I did grasp on the rendered PDFs linked in the PR description. I actually like it. :)

One thing (not sure if that will come after this): When a user follows the "Get Started" link they'll not have any information on further steps from there. Might be worth adding the "Develop with Nix" link in some form as "next steps" to the bottom of the install page?

garbas:
@andir You are reading my mind. I was just about to add the "breadcrumbs" like we have when you visit a team (under governance).

davidak (approved):

PR #594, opened 2020-09-29
Download page redesign

garbas:
## Design

- [Download_deskop_2.pdf](https://github.com/NixOS/nixos-artwork/blob/master/www/2020_redesign/Download/Download_deskop_2.pdf)
- [Download_mobile_2.pdf](https://github.com/NixOS/nixos-artwork/blob/master/www/2020_redesign/Download/Download_mobile_2.pdf)

## TODO

- [x] design headers for both sections (@garbas)
- [x] design collapsible that scales down to mobile (@garbas)
- [ ] design dropdown to show older versions of Nix/NixOS from the section headers (@samueldr), we can also design this after this PR, but we need to open a ticket so we don't forget.
- [x] text above navigation is not designed properly, should be above the navigation (@garbas)
- [x] is More options even needed (@garbas)
- [x] go over each navigation content pane and implement design (@garbas)

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/594)

### DIFF
--- a/download.tt
+++ b/download.tt
@@ -31,59 +31,57 @@
 
 <section class="download-nix">
   <div>
-    <div class="tabs-navigation">
+    <p>
+      Below you will find installation instructions for Nix, the package manager.
+    </p>
+    <article>
+      <h2><span>1.</span>Quick install</h2>
       <p>
-        Below you will find installation instructions for Nix, the package manager.
+        The quickest way to install Nix is to open a terminal and run the following
+        command (as a user other than <code>root</code> with <code>sudo</code> permission):
       </p>
-      <article>
-        <h2>Quick install</h2>
-        <p>
-          The quickest way to install Nix is to open a terminal and run the following
-          command (as a user other than <code>root</code> with <code>sudo</code> permission):
-        </p>
-        <pre class="terminal-console">
+      <pre class="terminal-console">
 <span class="shell-prompt">$ </span>curl -L https://nixos.org/nix/install | sh</pre>
-        <p>Make sure to follow the instructions output by the script.</p>
-        <p>The installation script requires that you have <code>sudo</code> access to <code>root</code>.</p>
-      </article>
-      <article>
-        <h2>Verify installation (GPG)</h2>
-        <p>You may want to verify the integrity of the installation script using GPG:</p>
-        <pre class="terminal-console">
+      <p>Make sure to follow the instructions output by the script.</p>
+      <p>The installation script requires that you have <code>sudo</code> access to <code>root</code>.</p>
+    </article>
+    <article>
+      <h2><span>2.</span>Verify installation</h2>
+      <p>You may want to verify the integrity of the installation script using GPG:</p>
+      <pre class="terminal-console">
 <span cls="shell-prompt">$ </span>curl -o install-nix-[%latestNixVersion%] https://releases.nixos.org/nix/nix-[%latestNixVersion%]/install
 <span cls="shell-prompt">$ </span>curl -o install-nix-[%latestNixVersion%].asc https://releases.nixos.org/nix/nix-[%latestNixVersion%]/install.asc
 <span cls="shell-prompt">$ </span>gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE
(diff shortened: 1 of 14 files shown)


site-styles/mixins.less
```diff
@@ -28,7 +28,7 @@
 // By default, xs and under use only one column.
 // "under" xs cannot be forced to anything else than 1.
 // sm and md can be configured if needed, but will default to @amount.
-#columns(@amount, @xs: 1, @sm: @amount, @md: @amount, @spacing: @gutter) {
+#columns(@amount, @xxs: 1, @xs: 1, @sm: @amount, @md: @amount, @spacing: @gutter) {
```
samueldr, line 31:
```suggestion
#columns(@amount, @xxs: 1, @xs: @xxs, @sm: @amount, @md: @amount, @spacing: @gutter) {
```

This way setting to e.g. `#columns(3, @xxs: 2)` sets it to 2 for `xs` too, as expected, rather than `[2, 1, 3, 3, 3]`.

samueldr (commented):
(notes of things to do)

download.tt
```diff
@@ -6,60 +6,55 @@
...
-<span class="shell-prompt">$ </span>curl -o install-nix-[%latestNixVersion%] https://releases.nixos.org/nix/nix-[%latestNixVersion%]/install
-<span class="shell-prompt">$ </span>curl -o install-nix-[%latestNixVersion%].asc https://releases.nixos.org/nix/nix-[%latestNixVersion%]/install.asc
-<span class="shell-prompt">$ </span>gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE
-<span class="shell-prompt">$ </span>gpg2 --verify ./install-nix-[%latestNixVersion%].asc
-<span class="shell-prompt">$ </span>sh ./install-nix-[%latestNixVersion%]</pre>
+<span cls="shell-prompt">$ </span>curl -o install-nix-[%latestNixVersion%] https://releases.nixos.org/nix/nix-[%latestNixVersion%]/install
+<span cls="shell-prompt">$ </span>curl -o install-nix-[%latestNixVersion%].asc https://releases.nixos.org/nix/nix-[%latestNixVersion%]/install.asc
+<span cls="shell-prompt">$ </span>gpg2 --recv-keys B541D55301270E0BCF15CA5D8170B4726D7198DE
+<span cls="shell-prompt">$ </span>gpg2 --verify ./install-nix-[%latestNixVersion%].asc
+<span cls="shell-prompt">$ </span>sh ./install-nix-[%latestNixVersion%]</pre>
```
samueldr, line 41: `s/cls/class/` ??

davidak:
I just took a superficial look, since it is not ready. 

The 1. 2. 3. implies you have to do all these steps to sucees. That makes no sense.

Screenshot from 2020-10-06 22-51-55

We have discussed the white background before.

>You may want to verify the integrity of the installation script

so the menu item should be called "verify installation script"

>You can uninstall Nix simply by running rm -rf /nix.

manual says:
>The install script will modify the first writable file from amongst .bash_profile, .bash_login and .profile to source ~/.nix-profile/etc/profile.d/nix.sh. 

does that also have to be removed to have a clean system again?

maybe that should be better documented in the manual und is not a topic for the download page

the $ in the command in selectable

please notice that light orange on dark orange has a bad contrast and is not very readable!

Screenshot from 2020-10-06 23-04-16

download.tt
samueldr: Why? This would break a rule I follow when writing: content should be in the file, and not in the CSS. This is part of the content here.

If it's about being able to select the content, we have the `#no-select();` LESS mixin which could be used here. We might want to apply `#no-select();` to something like `pre .shell-prompt { #no-select(); }`

samueldr:
I added an additional commit where we deviate from the palette by making the orange darker. That commit can be dropped from the change set if desired, though since this style is used in *body text*, I figure following the recommendations would be better.

garbas:
Looks good.

Regarding the colors, I created the issue to double check that our color ratio follows accessibility, https://github.com/NixOS/nixos-homepage/issues/609.

Issue #508, opened 2020-08-28
Font and typography: give it another look once redesign and content is in place


davidak:
This is a meta ticket to collect all the fixes needed regarding fonts and typography.

We need to give another look at the fonts and typography once redesign is done. The proposed design did not force us to use certain font (this was requested) but left it to us to decide. Few proposals were given:

- **Zilla Slab**
- **Krub**
- **Sora**
- **Open Sans**

### Collected feedback that we should consider:

- The typography on mobile is too splashy: too big and lack of security margins. Consequentially, it is harder to sync and parse the content (#600)
- Self host the font (#601, #551)

<details><summary>Original message of the issue</summary>
I'm not happy with the font choices for the redesign. They feel arbitrary and not specifically a good fit for our project.

- **Zilla Slab** looks like a typewriter, so not modern.
- **Krub** has many curves which don't fit the rest of the design with a lot of sharp edges
- **Sora** feels messy and unfocused with all the small details
- **Open Sans** is actually not bad. it's perfectly readable and for good reason the second most popular font on google fonts

Once we select fonts we should serve them from the nixos.org.

The demo uses **Sora** for the headings and **Krub** for the paragraphs in the current state.

![Screenshot from 2020-08-28 21-46-00](https://user-images.githubusercontent.com/91113/91609614-08ba5d00-e978-11ea-8d45-2a5e2f187412.png)

I find the fonts too fancy and detailed. The round w and x distract me from the content. I have autism and dyslexia, so i'm very easy distracted and it's hard for me to read. So it is an accessibility issue.

Here are my alternative suggestions:

![Screenshot from 2020-08-28 17-57-55](https://user-images.githubusercontent.com/91113/91610064-d5c49900-e978-11ea-897e-7e3b59f6226e.png)

**Source Sans Pro + Open Sans**

Perfectly readable, heavy heading.

![Screenshot from 2020-08-28 18-01-57](https://user-images.githubusercontent.com/91113/91610234-220fd900-e979-11ea-8f16-d5f8a9d3aa1d.png)

**Overpass + Open Sans**

(https://reproducible-builds.org/ uses Overpass for all text 😬)

i find the first suggestion more readable, the second looks sharper and might fit the nix snowflake and lambda better

We can ask the designer if these fonts fit together and to the rest of the new design. Source Sans Pro + Open Sans is a popular pairing according to google fonts.
</details>

blaggacao:
[...]

PR #601, opened 2020-10-01
Remove 'variants'

raboof:
In the sources there are 2 'variants' defined for fonts, and 3 for brightness
(light/dark/darker).

Do we plan to expose those at some point, or would it be a simpler to
get rid of them?

(the background is I'd like to look into self-hosting the fonts, but right now
it's not clear whether it makes sense to port all fonts or just the ones that
are used in 'variant A')

### DIFF
--- a/site-styles/components/navbar.less
+++ b/site-styles/components/navbar.less
@@ -11,11 +11,8 @@ body > header {
   #fill-width();
   background: #theme.navbar[background];
 
-  // The light variant of the navbar needs a shadow.
-  & when (#theme.navbar[variant] = light) {
-    border-bottom: 1*@unit solid rgba(0, 0, 0, 0.2);
-    box-shadow: #theme.common[box-shadow];
-  }
+  border-bottom: 1*@unit solid rgba(0, 0, 0, 0.2);
+  box-shadow: #theme.common[box-shadow];
 
   // Ensure *everything* in the navbar fills it vertically.
   * {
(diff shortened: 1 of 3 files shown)


samueldr:
See #508 for fonts; the choice hadn't been made yet.

For the time being, if the goal is to self-host the fonts (thanks) just prepare the files for the variant in use, and document any non-straightforward thing you had to do to get self-hosting done. **If** we end up going over fonts choice (as we should have beforehand) hopefully we (me) can just pick the appropriate files.

samueldr:
For any other kind of variants: don't remove any.

(Though I see it's only been done for the navbar.)

They are not present in the compiled CSS (if they are it should be fixed).

This is something you couldn't have known as it came from internal talks about implementing this, but we might end up using variants under some circumstances.

An example might be the Mobile NixOS website would use the same foundation for the site by importing the nixos homepage repository, but configure the darker menu rather than the lighter menu.

Same for the search site, it's not totally decided what is going to happen exactly, but it might be using variants that are not in use on the main site.

raboof:
> See #508 for fonts; the choice hadn't been made yet.

Yeah, saw that (I'm partial to the 'Source Sans Pro + Open Sans' option myself) - still we don't really need variants for that, we can just change it when we make the decision (possibly prototyping it in a branch). 

> This is something you couldn't have known as it came from internal talks about implementing this

Yeah, I figured as much - I didn't know where to bring it up and a PR seemed like a reasonable place to do it ;)

> it's free except in maintainer cognitive load (and then it's basically not an issue).

I think maintainer/contributor cognitive load is important :) - but if we see a potential use for them in the near(ish) future I'll close this PR :+1:

samueldr:
And for completeness' sake, this was done that way *during development* to allow other contributors to trivially test the two proposed font stacks.

PR #615, opened 2020-10-11
made the complete dependencies paragraph in features.tt a lot clearer for new users

makoConstruct:
When trying to learn how nix works through the features page, I had a lot of trouble understanding this section. We discussed it here https://discourse.nixos.org/t/site-intro-to-nixs-repeatable-builds-point-has-some-clarity-issues/9154/4 , and eventually things were explained to me. Accordingly, I have changed the section so that it will be more clear for new users.

In short, saying "compilers don’t search in per-packages
        directories such as
        <tt>/nix/store/5lbfaxb722zp…-openssl-0.9.8d/include</tt>,"
Is very confusing for a new user, because they are trying to understand how compilers actually do find their way into nix's store when they're allowed to, and this comes across as if it's saying that they don't. It is in fact a regular occurrence for compilers to "search" those directories, and they need to see this acknowledged, so it's very hard for them to find the intended meaning.

"Runtime dependencies are found by scanning binaries" is also somewhat confusing, as the result files of a build might not always be binaries. A new user would be led to ask "why just the binaries?"

### DIFF
--- a/features.tt
+++ b/features.tt
@@ -46,16 +46,15 @@
 
       <h2>Complete dependencies</h2>
 
-      <p>Nix helps you make sure that package dependency specifications are
+      <p>When you’re making a package for a package management system, like RPM,
+        you are supposed to declare its dependencies, but you can't easily
+        guarantee that your dependency declaration is complete. If you forget
+        a dependency, that you have separately installed on your machine, then
+        the component may build and work correctly on <em>your</em> machine,
+        but failing on the end user's machine.</p>
+
+      <p>Nix ensures that that package dependency specifications are
         complete.</p>
-      
-      <p>When you’re making a package for a package
-        management system like RPM, you are supposed to declare its dependencies,
-        but you can't easily guarantee that your dependency declaration is complete.
-        If you forget a dependency, then the
-        component may build and work correctly on <em>your</em> machine, if
-        you already have the dependency installed, while failing on the end user's
-        machine, if they don't.</p>
 
       <p>Under Nix, a build process will only find resources that have been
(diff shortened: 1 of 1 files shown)


features.tt
```diff
@@ -47,26 +47,25 @@
...
+        machine, if they don't.</p>
+
+      <p>Under Nix, a build process will only find resources that have been
+        declared explicitly as dependencies. There's no way it can build until
+        everything it needs has been correctly declared. If it builds, you will know
+        you've provided a complete declaration.</p>
+
+      <p>Once a build is complete, ongoing runtime dependencies are detected
+        automatically by scanning the package's build result files for the hash parts
+        of Nix store paths (such as <tt>r8vvq9kq…</tt>). This may sound risky,
```
7c6f434c, line 67: Maybe mention that it is enough to find at least one copy, not all of them? This adds confidence, and judging from some other discussions, it is a point worth an emphasis to give people a clearer picture.

garbas (approved):
@makoConstruct I think you are pointing a valid concern. Thank you.

I have added few changes, but overall it is an improvement over previous version.

Major rework on content of this page is coming together with the redesign sometimes next week.

features.tt
garbas: Features page shouldn't have to explain how scanning works just that runtime dependencies are detected.

makoConstruct: I don't see how that would be clarifying. Which other discussions? Perhaps there will turn out to be a better way of saying whatever it was they found.

Implying/letting on that it might not find all of them doesn't seem like it would instill confidence, to me.

PR #617, opened 2020-10-17
manual: do not use normal whitespace in code

jtojnar:
It breaks prompts among other things:
https://nixos.org/manual/nixpkgs/stable/#sec-patches

### DIFF
--- a/site-styles/pages/manual.less
+++ b/site-styles/pages/manual.less
@@ -61,7 +61,6 @@ div.docbook {
     background-color: inherit;
     color: inherit;
     font-size: 100%;
-    white-space: normal;
     -webkit-hyphens: none;
     -moz-hyphens: none;
     hyphens: none;


samueldr:
(Note to self) fix `pre code`

samueldr:
Yep, [it was part of the previous code](https://github.com/NixOS/nixos-homepage/blob/461c7622cc6cfa6a199c1c552c703e907544b841/css/nixos-site.css#L381-L391).

jtojnar:
Might be caused by a change in the syntax highlighter then? Because it seems to wrap the prompts in `<code>`.

samueldr:
Ah, I was updating the comment, sorry :)

The produced DOM is equivalent, so I don't think so.

The reason is that bootstrap added:

```
code {
   white-space: nowrap;
}
```

By default.

PR #632, opened 2020-10-26
learn: add link to the release notes

andir:
It *should* be frequently accessed content for everyone upgrading a
NixOS machine. This make it a bit more discoverable from the learn page.


samueldr (approved):

garbas:
@andir :+1: but I'll manually apply change to #593 or if you add the PR against `redesign-learn` before I do the change.

PR #634, opened 2020-10-27
Fix asciinema player breaking site width

garbas:
Fixes #597 

@samueldr I wonder if we should move this fix to `#flex-flexible-spacing` to apply it across all the site?

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/634)

Issue #597, opened 2020-10-01
Ascinema on mobile 2 (overflow)


blaggacao:
The videos overflow the screen.

I doubt it's soecial to my device (android one - motorola), is it?

_For some inexplicable readon I'm technically barred from pasting from the mobile._

garbas:
The video should not overflow. I'm attaching the screenshot from my phone

![photo6046177560712295166](https://user-images.githubusercontent.com/20208/95308479-40c39280-088a-11eb-9efd-1b52e03b1426.jpg)

Can you provide a screenshot from your phone? Also which browser are you using?

blaggacao:
![Screenshot_20201007-085938](https://user-images.githubusercontent.com/7548295/95341082-8593fd00-087b-11eb-8fa3-846b3b8f85a7.png)

I agree on the landscape mode, here is how it looks on my phone: slightly off.

garbas:
@blaggacao which browser is that?

blaggacao:
Ok sorry, I didn't figure it might be relevant.

- Motorola Android One
- DuckDuckGo Browser v5.66.1

turboMaCk:
I think this might be caused by right padding on `<li>` element as visible in the screenshot below.

![image](https://user-images.githubusercontent.com/2130305/97312351-adb4c180-1865-11eb-85ec-ed4d9c9c2c79.png)

At least its size looks somewhat similar to the offset I see on screenshots there. In any case I believe this padding should not be there on mobile version as it generally breaks the center alignment of that content.

samueldr:
That padding will not fix the overflow.

The overflow comes from a badly tested rule set for handling vertical/horizontal in [`site-styles/components/asciinema.less`](https://github.com/NixOS/nixos-homepage/blob/ce5d0225b7498dacd7ec1581cd9b75f0491e3f2f/site-styles/components/asciinema.less#L19-L23).

To reproduce:

 * Have an horizontal window, rather than vertical window with `xs` or smaller.

Pro-tip: Docking the inspector at the bottom helps here.

turboMaCk:
I can reproduce it now. It's impossible to reproduce using mobile size simulation but in classic desktop mod with both low height and width and with player open this easily reproducible (both firefox and chrome/brave)

PR #636, opened 2020-11-04
misc. alignments and columns fixes

samueldr:
This PR closes #634.

* * *

With this, we also gain a new debugging superpower: the site width is always shown with yellow rulers. In addition to this, columns are pink, and the flexible spacing helpers are orange and green.

(What does this mean? Scroll down to the NixOS logo in the footer and double click the lambdaflake.)

* * *

These are all issues I've had to fixup following through actually fixing what #634 fixed (which was not fixing the issue it tried to, this is for another PR.)

Mostly things that were found out to accidentally align invisibly with an additional stray "gutter" width to the right. Comparing the site with the live one you should see *some* changes at many places columns are used. For some it might have changed how the text can flow.

This does address the stray padding @TurboMack saw

 * https://github.com/NixOS/nixos-homepage/issues/597#issuecomment-717267271

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -80,63 +80,65 @@ BLOCK navigationLink %]
   </main>
 
   <footer>
-    <div class="upper">
-      <section>
-        <h4>The project</h4>
-        <ul>
-          <li><a href="https://status.nixos.org/">Channel Status</a></li>
-          <li><a href="https://search.nixos.org/packages">Packages search</a></li>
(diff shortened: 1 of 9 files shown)


garbas (approved):
Looks good

turboMaCk:
For the record this looks great :+1: I also like the idea introducing the debugging grid.

PR #637, opened 2020-11-04
site-styles: Fix asciinema breaking width by removing aspect ratio code

### LINKED ISSUE #597: Ascinema on mobile 2 (overflow)

garbas:
fixes #597 

> The reason there was that aspect ratio thing there was to, if I remember
> right, make the whole player fit in the height of a side-ways mobile
> device.
> 
> Though I think I must have forgotten to finish implementing this, as
> instead it just resolved to using the default font sizes.
> 
> Rather, let's do what works fine for every other sizes: a ratio of the
> device width.

— @samueldr

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/637)

### DIFF
--- a/site-styles/components/asciinema.less
+++ b/site-styles/components/asciinema.less
@@ -18,7 +18,7 @@
 
       // vertical (portrait) and square aspect ratio
       // In that case we set it as a fraction of the width.
-      @media screen and (max-aspect-ratio: ~"1/1") and (max-width: (@screen-sm-max)) {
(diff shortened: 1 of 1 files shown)


garbas:
I forgot to rebase on top of master, before opening this PR.

PR #642, opened 2020-11-12
Revert "NixOS Download: recommend GNOME"

grahamc:
Reverts NixOS/nixos-homepage#641 Let's go back to Plasma until @worldofpeace / @jonringer take a looksee.

### DIFF
--- a/download.tt
+++ b/download.tt
@@ -110,13 +110,13 @@
 
         <ul class="download-buttons">
           <li>
-            <a href="[%prefix%]/latest-nixos-gnome-x86_64-linux.iso">Download (GNOME, 64bit)</a>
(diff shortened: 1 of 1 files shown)


grahamc:
From that PR:

> The GNOME ISO correctly detects and configures HiDPI scaling.
> More, it supports changing the scale. The Plasma Desktop ISO
> doesn't detect and configure the scale automatically, and doesn't
> allow you to change it without restarting. This makes the GNOME
> ISO significantly more accessible for users on laptops with HiDPI
> displays.

garbas:
Thank you for reverting this. Lets bring this up for discussion on discourse and involve release manager(s).

PR #643, opened 2020-11-12
Revert "Revert "NixOS Download: recommend GNOME""

grahamc:
Reverts NixOS/nixos-homepage#642

From that PR:

> The GNOME ISO correctly detects and configures HiDPI scaling.
> More, it supports changing the scale. The Plasma Desktop ISO
> doesn't detect and configure the scale automatically, and doesn't
> allow you to change it without restarting. This makes the GNOME
> ISO significantly more accessible for users on laptops with HiDPI
> displays.

Bringing this back to the table. After discussing it with @worldofpeace, they +1'd. Maybe let's hear from @jonringer too.

### DIFF
--- a/download.tt
+++ b/download.tt
@@ -110,13 +110,13 @@
 
         <ul class="download-buttons">
           <li>
-            <a href="[%prefix%]/latest-nixos-plasma5-x86_64-linux.iso">Download (Plasma Desktop, 64bit)</a>
-            <a href="[%prefix%]/latest-nixos-plasma5-x86_64-linux.iso.sha256">(SHA-256)</a>
+            <a href="[%prefix%]/latest-nixos-gnome-x86_64-linux.iso">Download (GNOME, 64bit)</a>
+            <a href="[%prefix%]/latest-nixos-gnome-x86_64-linux.iso.sha256">(SHA-256)</a>
             <span class="label -success">Recommended for most users</span>
           </li>
           <li>
-            <a href="[%prefix%]/latest-nixos-gnome-x86_64-linux.iso">Download (GNOME, 64bit)</a>
-            <a href="[%prefix%]/latest-nixos-gnome-x86_64-linux.iso.sha256">(SHA-256)</a>
+            <a href="[%prefix%]/latest-nixos-plasma5-x86_64-linux.iso">Download (Plasma Desktop, 64bit)</a>
+            <a href="[%prefix%]/latest-nixos-plasma5-x86_64-linux.iso.sha256">(SHA-256)</a>
           </li>
         </ul>
 


edolstra:
Recommending Gnome over Plasma on the basis of HiDPI support doesn't seem very compelling. I mean, these desktop environments have thousands of features, why should HiDPI support be the deciding one?

grahamc:
The main issue with the Plasma ISO is the font size is very small and unreadable, and unfixably so from within the desktop environment. I'm not sure the features of a live CD matter when the user isn't able to read it. If "can read the screen" isn't a compelling feature, maybe it isn't useful to recommend either.

edolstra:
KDE/Plasma has been NixOS's default desktop environment since pretty much the very beginning. Changing it should be a major decision, and not based exclusively on "the font is too small on some screens". (And certainly such a change shouldn't be merged [in 3 minutes](https://github.com/NixOS/nixos-homepage/pull/641)...)

grahamc:
Yeah, I merged it a bit quick :wink:, sorry about that. I'm not sure it is actually such a major decision, though.

worldofpeace:
> @worldofpeace:  so, overall the GNOME3 desktop is in a much more working state in nixpkgs and it actually has wayland support. Updates seem to have a lot less fallout, we don't have the difficult situation like qt has as much in the glib/gtk world I think

> I think discussing on discourse is not really going to be a good idea
> this is a marketing, release managers, and the desktop maintainers discussion. It will likely get hijacked to not be productive. 

> @grahamc: what do you think about recommending the gnome iso by default?

> @worldofpeace: I would
> like, the PR for wayland support for plasma has been open for really long time now without progress. Upstream wayland session support came from gnome focused nix devs, overall if you look at the release notes there's a lot of enduser consideration there as well. It would be a good choice and I'd be happy to invest more fixes into stable

Quoting my discussion I had quickly on #nixos-dev. 

 In my observation there is actually traffic to this DE in nixpkgs, we had the most contributors working together to land 3.38 https://github.com/NixOS/nixpkgs/projects/30 recently and I hope we can sustain more people. Plus GNOME releases two times a year so it could be easy to harmonize the latest if we tweak our release date a little bit.

@edolstra Yeah, I think @grahamc point there was just a side effect of all the other things I mentioned. I think we should just recommend what is likely to be fresh and a good experience in nixpkgs. Not because x and x features, Plasma 5 wouldn't have some of these issues if we had the wayland session. What Graham noticed is because the **default** is wayland, and I don't think Plasma5 has parity with the X session yet. So the real compelling feature is how quick GNOME was to adopt wayland and that we have it working veritably well. But even then that isn't my main motivation at least. In the end it's just putting one image in prominence over the other, and in my, probably slightly biased opinion, would be a good choice because: more than one human actively working on it, and clearly evident that it continues to improve in nixpkgs. A read from from some of the past few releases should show this pretty clearly. :rofl: Even default color emoji was a decision from GNOME nix devs. I hope this makes sense

I would also ping @jtojnar into this.

worldofpeace:
> Yeah, I merged it a bit quick wink, sorry about that. I'm not sure it is actually such a major decision, though: it isn't actually the _default_ environment, just what the ISO contains. The ISO doesn't influence what the configuration generator does, and the configuration generator doesn't enable a desktop environment.

Right, I think with what information I've provided  it's clear it isn't a major decision, just that we give prominence to that needs a bit of due consideration. NixOS can't make someone **have a default DE**, but we can recommend a *default*.

jtojnar:
> The ISO doesn't influence what the configuration generator does, and the configuration generator doesn't enable a desktop environment.

I believe it does now: https://github.com/NixOS/nixpkgs/pull/100199

Otherwise, I agree with what has been said. GNOME is currently significantly more maintained than Plasma – we have a [team](https://github.com/NixOS/nixpkgs/blob/25f45a5226704046861455131adce3d6b4a4c19a/maintainers/team-list.nix#L59-L67) and lot of other people helping, whereas Plasma is just Thomas AFAICT.

worldofpeace:
Oh yeah I did do that https://github.com/NixOS/nixpkgs/pull/100199. I can change that so the line is commented if that's desired.

jonringer:
Personally, I don't use a DM, so I don't have a huge preference. However, having tried to fix plasma issues, I would prefer having gnome as our recommended graphical iso.

Reasons being:
- The plasma ecosystem has a lot of behavior that's hard to determine until runtime (qml + qt), this makes iterating on issues really difficult
- KDE ecosystem is huge and always out of sync (qt + kde applications + plasma) (there was 15+ issues on gh about broken kdeapplication in 20.09)
- Plasma has a 4 month release cycle, while gnome has a 6 month release cycle, matching our own cadence.
- Support: gnome has many people packaging it for nix, while plasma is largely just lead by @ttuegel (which he does great work, but qt + kde + plasma packaging needs many more contributors)
- Since we cut from master, we can't really use LTS versions of plasma, as they are unlikely to support the latest systemd and other "fundamental packages" https://github.com/NixOS/nixpkgs/issues/98141

In short, supporting plasma is a massive burden, and we don't have a proportional number of maintainers to support it as a tier 1 Desktop manager.

worldofpeace (approved):
I'm all for this. @jonringer expressed his opinion and it's supporting. I think basically all reasoning has been laid out to do this. I'll be merging soon if there's no objections. (mostly want to hear back from @edolstra ).

Another thing we should do is express somewhere less static that we're now recommending gnome.

ttuegel:
> The main issue with the Plasma ISO is the font size is very small and unreadable, and unfixably so from within the desktop environment.

For what it's worth, I believe this is described in #105943.

> Support: gnome has many people packaging it for nix, while plasma is largely just lead by @ttuegel (which he does great work, but qt + kde + plasma packaging needs many more contributors)

Ultimately, this is my fault for being difficult to work with and failing to recruit other maintainers. :disappointed:

This time around, we were especially thwarted by this:

> KDE ecosystem is huge and always out of sync (qt + kde applications + plasma) (there was 15+ issues on gh about broken kdeapplication in 20.09)

Plasma and KDE Applications couldn't agree on a version of Qt with long-term support. This sort of problem will go away in the future because Qt is not making long-term support releases anymore.

> The plasma ecosystem has a lot of behavior that's hard to determine until runtime (qml + qt), this makes iterating on issues really difficult

Increasingly, I worry it may not be possible to sanely support Qt and Plasma on NixOS for this reason. I try to look to GTK and Gnome for inspiration whenever possible, but often I find they just don't have the problems of Qt in the first place. :worried:

grahamc:
Thomas, I just want to say I don’t think this reflects on you at all. You’ve done great work and continue to. Maybe we can grow the team — it’d surely be nice to. Thank you for all the very difficult and routine work you do.

PR #648, opened 2020-11-24
Change landing page description

garbas:
and add 3 icons for the "why nix" section.

---
This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/648)


garbas:
Oh I was hoping to discuss this first. Maybe somebody has a better version, but maybe this will spark the discussion and we can change it afterwards as well. Thank you for the all the merging.

samueldr:
The lambdaflakes have the wrong colours here.

This probably means the dark blue in use in the borders is not "proper", but it's entirely possible we want it darker than the dark blue.

Issue #104, opened 2016-10-08
separate source files and build artifacts


cko:
Atm the build outputs are stored in the same directory as the source files e.g. `nixos-homepage/nixos/options.tt` (source) and `nixos-homepage/nixos/options.html` (build output). Then everything seems to be deployed to the live website e.g.  http://nixos.org/nixos/options.html and http://nixos.org/nixos/options.tt
It would be nicer to have a directory like `build` where all build artifacts are stored. This would also simplify the `.gitignore` file and prevent accidental commits of generated files.

nixinator:
I've looked it this make, it's not as simple as it seems, Make is been used as it was intended, it's tracking the resultant .html files, if there are no change, and the .html files exist, nothing is built!!!..... I presume this was something to do with the neat

[nix-shell]$ fd | entr make

where as any modification to file in the directory, would trigger a rebuild,  but only what needs to be rebuilt. 

 if you try repath the output from tpage, then it works, but you end up with make file that constantly builds the page as it can't track the relocated html files. 

If this really is an issue, then maybe make should copy of build file to some output directory, and that can be served by the web server???

garbas:
@nixinator the idea is to have all website related files (html/css/js/...) being created in some other folder (eg. `./build`) and when deploying I will only sync this folder to netlify. Then all the other things will just continue to work. Makes sense?

nixinator:
Sure,  I'll send you a PR.

garbas:
@nixinator also maybe work on top of #454 since removes some resources and we can already avoid conflicts this way.

nixinator:
Ok, i'll take a look at that. I can modify the make file to do a build in place as it works now, and then mv all the artefacts to build/ directory. However, the way 'make' works, it's not great at tracking when files have been moved with mv. If make cant't track these, it can't do conditional rebuilds based on files changing. 

The make file builds the website in place as I suspect tools like xsltproc can play havock with relative and absolute paths when processing, so it builds it in the root. Maybe the best course of action is that everything that is created in a .tmp , gets copied to a build/. dir. I'm also loathed to do that too, as it replicates stuff, and make file more complex :-(.

That means changing the  bootstrapify-docbook.sh and the make file :-( with a cp -r build/ after each step.

garbas:
I think you can simple track files in in build folder.

for example the following line
```
HTML = index.html download.html news.html learn.html community.html \
```

would become

```
[...]

PR #652, opened 2020-11-27
Smaller improvements to the README

garbas:
and some smaller improvements to the README

This change is [Reviewable](https://reviewable.io/reviews/nixos/nixos-homepage/652)


README.md
```diff
@@ -1,25 +1,47 @@
-NixOS.org
-=========
+# Source for nixos.org
 
-This repository contains the sources of the `nixos.org` website.  To
-build it:
+[![GitHub Workflow Status of master branch](https://img.shields.io/github/workflow/status/nixos/nixos-homepage/Build%20&%20Deploy%20to%20Netlify?style=flat)](https://github.com/NixOS/nixos-homepage/actions?query=workflow%3A%22Build+%26+Deploy+to+Netlify%22) [![Number of open GitHub issues](https://img.shields.io/github/issues/nixos/nixos-homepage?style=flat&color=red)](https://github.com/nixos/nixos-homepage/issues) [![Number of open GitHub pull requests](https://img.shields.io/github/issues-pr/nixos/nixos-homepage?style=flat&color=blue)](https://github.com/nixos/nixos-homepage/pulls) [![Month of last commit](https://img.shields.io/github/last-commit/NixOS/nixos-homepage?style=flat)](https://github.com/NixOS/nixos-homepage/commits/master) [![Number of all contributors](https://img.shields.io/badge/all_contributors-10-orange.svg?style=flat)](https://github.com/nixos/nixos-homepage#how-to-help)
+
+Code and content for [nixos.org](https://nixos.org) website.
```
samueldr, line 5:
```suggestion
Code and content for the [nixos.org](https://nixos.org) website.
```

run.py
```diff
@@ -0,0 +1,15 @@
+from livereload import Server, shell
```
samueldr, line 1: I know there's already a lot of clutter in the root folder. It's a bit late to ask that, but it would have been nice to make all the *tooling* parts live in a subfolder, e.g. `build-support`, so that the root folder really only holds the data for the site. Do you think it's the time we start doing that for the newly added tooling?

(This is somewhat attached to the `serve` alias, where it wouldn't matter where this script lives.)

README.md
```diff
@@ -1,25 +1,57 @@
...
+
+# Help us!
+
+There are many ways how you can help:
+
+- if you are familiar with CSS look at the [issues tagged with `design` tag](https://github.com/NixOS/nixos-homepage/issues?q=is%3Aissue+is%3Aopen+label%3Adesign).
+- if you are an native English speaker or just a person that is very good with words, please look at the [issues tagged with `content` tag](https://github.com/NixOS/nixos-homepage/issues?q=is%3Aissue+is%3Aopen+label%3Acontent)
+- if you are developer and just eager to fix stuff please look at the [issues tagged with `bug` tag](https://github.com/NixOS/nixos-homepage/issues?q=is%3Aissue+is%3Aopen+label%3Abug)
+
+If you feel lost where and how to contribute, ask [marketing team](https://nixos.org/teams/marketing.html) on [`#nixos-marketing` channel on IRC](irc://irc.freenode.net/#nixos-marketing).
```
samueldr, line 16:
```suggestion
If you feel lost where and how to contribute, ask the [marketing team](https://nixos.org/teams/marketing.html) on the [`#nixos-marketing` channel on IRC](irc://irc.freenode.net/#nixos-marketing).
```

README.md
```diff
@@ -1,25 +1,57 @@
...
+      and open browser on:
+          https://127.0.0.1:8000
+
+      It will rebuild the website on each change.
+
+    [nix-shell]$ python run.py
+
+Open your browser at: http://127.0.0.1:8000/index.html
+
+In order for browser to automatically refresh install [Livereload extension](http://livereload.com/extensions/) for you browser.
```
samueldr, line 41:
```suggestion
In order for the browser to automatically refresh, install the [Livereload extension](http://livereload.com/extensions/) for your browser.
```

README.md
```diff
@@ -1,25 +1,57 @@
...
+
+Before creating a pull request make sure that `nix-build` runs successfully.
+
+
+## Binary cache
 
-To automatically rebuild on every change:
+It can take some time to enter the development environment. To speed up and avoid building from source, you can use a binary cache that is also use to speed up our GitHub Actions.
```
samueldr, line 48:
```suggestion
It can take some time to enter the development environment. To speed up and avoid building from source, you can use a binary cache. The same cache is used to speed up our GitHub Actions.
```

README.md
```diff
@@ -1,25 +1,57 @@
...
 
-    [nix-shell]$ fd | entr make
+Add the following to the `/etc/nix/nix.conf` or `~/.config/nix/nix.conf`:
```
samueldr, line 50: Hm, editing `/etc/nix/nix.conf` on NixOS is not really going to work. How can we rephrase this in a way that handles NixOS, and Non-NixOS alike?

Quite out of scope, but it would be best if there was a centralized page that deals with how to edit the Nix configuration, so we could link to it, with which configuration option to edit.

Or maybe even a documentation section about adding *binary caches* outright?

README.md
```diff
@@ -1,25 +1,57 @@
...
+          python run.py
+
+      and open browser on:
+          https://127.0.0.1:8000
+
+      It will rebuild the website on each change.
+
+    [nix-shell]$ python run.py
+
+Open your browser at: http://127.0.0.1:8000/index.html
```
samueldr, line 39:
```suggestion
Open your browser at: http://127.0.0.1:8000/
```

Since the command doesn't use `index.html` in its output.

flake.nix
```diff
@@ -147,6 +145,16 @@
             # SITE_STYLES skipped by design.
             export NIX_PILLS_MANUAL_IN="${nixPills}/share/doc/nix-pills"
             export NIX_DEV_MANUAL_IN="${nix-dev.defaultPackage.x86_64-linux}/html"
+
```
samueldr, line 148: Since we're using the shell hook, could we provide a `serve` alias that wraps the method by which the site is served, so that whenever it is changed in the future, users don't have to care about the implementation detail?

"Whenever?" you ask? We don't know what tooling is going to be used for the website once it's migrated to a better integrated markdown tooling following that RFC. That tooling might provide its own method of live reloading.

flake.nix
```diff
@@ -147,6 +145,16 @@
...
+            echo ""
+            echo "  To start developing run:"
+            echo "      python run.py"
+            echo ""
+            echo "  and open browser on:"
+            echo "      https://127.0.0.1:8000"
```
samueldr, line 154:
```suggestion
            echo "  and go to the following URL in your browser:"
            echo "      https://127.0.0.1:8000/"
```

The  `/` at the end is me being a purist about URLs, we can drop it. If we do, also drop it in that other comment about `index.html`. (Or use `/index.html` here too.)

README.md
```diff
@@ -1,25 +1,57 @@
...
     $ nix-shell
-    [nix-shell]$ make
-    [nix-shell]$ python -m http.server
 
-then open http://127.0.0.1:8000/index.html
+      To start developing run:
+          python run.py
+
+      and open browser on:
+          https://127.0.0.1:8000
```
samueldr, line 33: Remember to update this with the final actual output of the shell hook if you edit it.

README.md
```diff
@@ -1,25 +1,57 @@
...
 
-To automatically rebuild on every change:
+It can take some time to enter the development environment. To speed up and avoid building from source, you can use a binary cache that is also use to speed up our GitHub Actions.
 
-    [nix-shell]$ fd | entr make
+Add the following to the `/etc/nix/nix.conf` or `~/.config/nix/nix.conf`:
 
-To test the complete result from a nix-build:
+substituters = ... https://nixos-homepage.cachix.org
+trusted-public-keys = ... nixos-homepage.cachix.org-1:NHKBt7NjLcWfgkX4OR72q7LVldKJe/JOsfIWFDAn/tE=
```
samueldr, line 53:
````suggestion
```
substituters = ... https://nixos-homepage.cachix.org
trusted-public-keys = ... nixos-homepage.cachix.org-1:NHKBt7NjLcWfgkX4OR72q7LVldKJe/JOsfIWFDAn/tE=
```
````

samueldr (changes requested):
Some nit picks, some weird english formulation. And a request that I don't think will be handled.

Otherwise good idea.

samueldr:
In addition to the review, I will formally add that I don't like embracing yet another third-party service as part of the official presentation and documentation of the project. I'm talking about `gitpod`.

While yes, they allow self-hosting, adding it that way to the README and the project adds tacit approval from the organization to that *service*.

run.py
garbas: I also dislike this big root folder, but I woulnd't like to compilcate this PR.

Lets plan this for the new tooling or maybe even to do it together with #104.

flake.nix
garbas: :+1: awesome idea.

garbas:
I've address all the comments and suggestions.

> In addition to the review, I will formally add that I don't like embracing a third-party service as part of the official presentation and documentation of the project. I'm talking about `gitpod`.
> 
> While yes, they allow self-hosting, adding it that way to the README and the project adds tacit approval from the organization to that _service_.

Personally I don't think having a GitPod link in readme adds any approval from the organization. I would like more contributions and if that means we will add more configuration for services that help contributors of `nixos-homepage`, then I don't see why not.

But there is a big difference when adding this links and promoting this tools and services on the nixos.org or other subdomains.

We should use all the help internet can provide in order to attract more contributions to `nixos-homepage`.

edolstra:
I'm not in favor of adding a dependency on yet another proprietary service, especially when it should be unnecessary. Building the homepage should require nothing more than running `nix build`.

> Personally I don't think having a GitPod link in readme adds any approval from the organization.

I think that's a bit naive. From the Gitpod homepage: "Describe your dev environments as code". That's exactly what Nix purports to do, so the fact that we're relying on Gitpod undercuts the argument for using Nix.

garbas:
> I'm not in favor of adding a dependency on yet another proprietary service, especially when it should be unnecessary. Building the homepage should require nothing more than running `nix build`.
> 

If this would be the case then I wouldn't be trying to convince you to add this button.

> > Personally I don't think having a GitPod link in readme adds any approval from the organization.
> 
> I think that's a bit naive. From the Gitpod homepage: "Describe your dev environments as code". That's exactly what Nix purports to do, so the fact that we're relying on Gitpod undercuts the argument for using Nix.

I don't think relying on GitPod undercuts the argument for using Nix. Then we could use the same argument for Docker or even AWS. Nix works alongside Docker and AWS and improves the experience. This is the same for Gitpod.

Just to move further I will remove this link and configuration, but I hope to discuss this at the next marketing meeting. Eg. why do we provide links to AWS and not to other services. Just to define what are the guidelines and what are the policies.

PR #656, opened 2020-12-17
extract common styles that can be used for other websites

### LINKED ISSUE #606: Document site-styles/
I suggest use some of the _living style generators_ to extract documentation from comments.

I think this one is popular at the moment https://github.com/mpetrovich/stylemark but I don't have any preference.

garbas:
This is initial work to extract common styles into https://github.com/NixOS/nixos-common-styles.

fixes #606 as well


samueldr (approved):

blaggacao:
Wow! This is such well invested effort. And thanks for providing a flake example for those kind of use cases.

PR #658, opened 2020-12-20
add short description for Nix and NixOS

bobvanderlinden:
I noticed that the website prominently shows the download button as one of the first things on the main page. It feels like I should download first before hitting the tutorial behind the "Get started" button.

Additionally, NixOS section is sometimes a bit hidden because it is shown below the Nix section. Example:

image

I can imagine people getting confused when hearing about an OS and needing to run a shell command in order to install it.

For these reasons I thought it would be good to shortly explain the difference between Nix and NixOS. In this PR I have added a short description to each section header.

Alternatively (or additionally) it might be a good idea to show a short block of text explaining which choice to make with a link to the right section.


asymmetric:
@bobvanderlinden could you provide a screenshot of the changes?

garbas (approved):

PR #669, opened 2021-01-17
Blog sections of the website

### LINKED ISSUE #446: Blog / News section

garbas:
This PR introduces new section to the website: Blog

We used to show latest announcements on the landing page, but that clearly is not what landing pages are about. For some time now the news.html was still there and we only show the banner on the front page for announcements. But all along the plan was to rework this section and this PR is is the first step towards that.

At the same time as I added the Blog section I also joined the Community and Governance section.

Icons clearly need some work, but I need some help here, since my GIMP/Photoshop level is pretty basic.

The content of this page is far from perfect, but it is an important first step.

The future plan for this sections of the website is to:
- include NixOS Weekly and show it at `/blog/index.html`
- include more sections from discourse, but this needs to be coordinated with the discourse team and some moderation help might be needed.
- retire planet.nixos.org

fixes #446


edolstra:
Does this replace planet.nixos.org?

garbas:
@edolstra Not yet. That would be a separate PR, I will add this into the after steps. It will complicate a bit the build process, so I want to tackle it separately.

scripts/update_blog_categories.py
```diff
@@ -0,0 +1,121 @@
...
+@click.option('--output', required=True, type=str)
+def main(limit, number_of_feeds, output):
+    loop = asyncio.get_event_loop()
+    result = loop.run_until_complete(fetch_categories(limit, loop, CATEGORIES))
+
+    click.echo(f"      Writing to `{output}`")
+
+    categories = TEMPLATE.format(
+        categories="".join([
+            CATEGORY_TEMPLATE.format(
```
hugolgst, line 99: Just a quick comment for maintainability/readability:
Shouldn't we split this part into different variables?

What do you think?

samueldr:
(Not a review yet)

 - [ ] archive.html redirection is missing

PR #677, opened 2021-02-14
Tweak Explore page graphic

DeeUnderscore:
There is something odd going on with the legs on the lion graphic on the Explore page—the lion's leg is in front of the table leg:

Lion graphic from live website

I've manually relayered the relevant groups to put the lion's leg behind the table leg, so after rebuilding the website locally, the lion now looks like this: 

Lion graphic with the legs fixed


garbas (approved):
:heart:

PR #678, opened 2021-03-05
Copyright notice: NixOS -> NixOS contributors

roberth:
This used to be a rather vague statement, because "NixOS" is not
the name of a person or legal entity.
Nobody has signed away their copyright, so it remains with the
individual contributors.
This is not legal advice.


garbas (approved):

garbas:
Thank you!

PR #700, opened 2021-04-22
index.tt: add emails of Foundation board members

asymmetric:
To make it easier to contact them.

Ideally though, the foundation would have its own email address.


edolstra:
The foundation does have its own email address: foundation@nixos.org. I think it's better to use that than individual email addresses.

asymmetric:
Ping :)

garbas (approved):
:heart:

PR #704, opened 2021-05-08
update-blog: add cchardet dependency

lukegb:
aiohttp will use cchardet to determine the charset of the page it
retrieves if it's available and isn't specified in the Content-Type
header. In the case of some of the sources it isn't specified, which
leads to problems when parsing them.

This fixes the failed updates - https://github.com/NixOS/nixos-homepage/runs/2535903129?check_suite_focus=true.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -77,7 +77,7 @@ rec {
         mkPyScript (with pkgs.python3Packages; [ click toml ]) "shuffle-commercial-providers";
 
       update_blog =
-        mkPyScript (with pkgs.python3Packages; [ aiohttp click feedparser ]) "update-blog";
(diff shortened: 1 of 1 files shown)


lukegb:
cc @garbas since I think this is blocking the live site from updating at the moment

The alternative approach here is to just assume that the content is always utf-8, which also seems reasonable and might be better.

samueldr:
@lukegb what would it look like to assume this is always UTF-8?

lukegb:
https://github.com/NixOS/nixos-homepage/blob/master/scripts/update-blog.py#L125

--> `await response.text('utf-8')`

samueldr:
Thanks, won't be my call, but that makes the investigation easier on @garbas for whenever he has the time to look into it :).

garbas (approved):
Thank you :heart: for taking the time to fix this and sorry that it took so long to review it.

PR #709, opened 2021-05-20
Add CORS policy for matrix client endpoint

mweinelt:
Untested, but based on https://github.com/NixOS/nixos-homepage/pull/709 and https://docs.netlify.com/routing/headers/.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -430,3 +430,8 @@
   to = "/css/nixos-site.css"
   status = 302
   force = true
+
+[[headers]]
+  for = "/.well-known/matrix/client"
+    [headers.values]
+    Access-Control-Allow-Origin = "*"


netlify.toml
```diff
@@ -430,3 +430,8 @@
   to = "/css/nixos-site.css"
   status = 302
   force = true
+
+[[headers]]
+  for "/.well-known/matrix/client"
```
grahamc, line 435: cool, thanks!

```suggestion
  for = "/.well-known/matrix/client"
```


maybe? not sure, is the indentation right? the red highlighting is suspicious :)

mweinelt: Yeah, I was seeing that as well. My TOML-foo is worryingly non-existant.

grahamc:
Thank you!

PR #728, opened 2021-06-29
Fix typo in donate.tt

CRTified:
It should probably say "money" instead of "mone".


CRTified:
Duplicate of #729 :sweat_smile:

PR #738, opened 2021-08-22
Add the NixOS Foundation's statement of inclusion

grahamc:
Established May 7 this year with Armijn, Rob, and Eelco. At the time,
it was primarily shared on Twitter:

https://twitter.com/grhmc/status/1390775249424338944

I should have added it to the website then, but now is a good time
too.

### DIFF
--- a/community/index.tt
+++ b/community/index.tt
@@ -10,6 +10,13 @@
 <section class="community-description">
   <p>Nix and NixOS are developed and used by a diverse and welcoming community
     from all around the world.</p>
+  <p>The NixOS Foundation aims to promote participation without regard to gender,
+    sexual orientation, disability, ethnicity, age, or similar personal
+    characteristics.</p>
+  <p>We want to strive to create and foster community by providing an
+    intentionally welcoming and safe environment where all feel valued and cared
+    for, and where all are given opportunity to participate meaningfully. The
+    Foundation will work with the community in service of this goal.</p>
 </section>
 
 <section class="community-communication">
--- a/site-styles/pages/community.less
+++ b/site-styles/pages/community.less
@@ -1,6 +1,10 @@
 .community-description {
   .generic-layout();
-  .lead();
+
+  // Only apply `lead` to the first paragraph of the community statement.
+  & > :first-child {
+	.lead();
+  }
 }
 
 .community-banner {


grahamc:
image

andir (approved):

layout.tt
```diff
@@ -117,6 +117,18 @@ BLOCK navigationLink %]
...
+            The NixOS Foundation aims to promote participation without
+            regard to gender, sexual orientation, disability, ethnicity,
+            age, or similar personal characteristics. We want to strive
+            to create and foster community by providing an intentionally
+            welcoming and safe environment where all feel valued and cared
+            for, and where all are given opportunity to participate
+            meaningfully. The Foundation will work with the community in
+            service of this goal.
+          </span>
+        </div>
```
samueldr, line 131: Unclassified `<div>` are forbidden. Use a `<section>`, and add a class e.g. `footer-statement`.

The `<span>` serves no purpose here. It should be removed. Though a `<p>` would be needed as this is more akin to *copy text* than some form of *functional text*. So rather than outright removing the `<span>` a `<p>` should be used.

samueldr:
Currently this would require a tiny bit of edits to the common styles to fit in the style guides.

It would require what amounts to [*adding this to all but last `section`s of the footer*](https://github.com/NixOS/nixos-common-styles/blob/94b4ed63d18d52dfa5d146d909f054006427304a/src/components/footer.less#L123-L126), and applying [*the font size change to all `.lower` elements instead of explicitly to some*](https://github.com/NixOS/nixos-common-styles/blob/94b4ed63d18d52dfa5d146d909f054006427304a/src/components/footer.less#L99).

I guess if adding such a lengthy blerb at the bottom of the site is desired, I can handle doing the change.

grahamc:
I would appreciate that help! :)

edolstra:
The footer should be kept to a minimum and is not the right place for this kind of statement. The community page (https://nixos.org/community/index.html) seems like a better place.

domenkozar:
It would be great to have a CoC, but this is a very good start :)

domenkozar (approved):

samueldr:
This style change comes from a discussion with Graham.

PR #769, opened 2021-12-05
manual: Prevent the stylesheet from eating the admonition contents

jtojnar:
When an admonition contained other elements than paragraphs (e.g. pre), they were removed. Additionally, custom titles were replaced in single-paragraph admonitions by generic ones and all titles were stripped in multi-paragraph ones.

With this patch, the stylesheet should preserve all contents of admonitions when trying to Bootstapify them.

### Example

Noticed by @maralorn and reported on [Matrix](https://matrix.to/#/!kjdutkOsheZdjqYmqp:nixos.org/$pwoosV97vjUhIwO-Lh-18MJvvvDTAmF67sHC-FazFDs?via=nixos.org&via=matrix.org&via=nixos.dev).

#### On the website
[Note in manual on the website](https://nixos.org/manual/nixos/stable/index.html#sec-booting-from-usb)

#### Untransformed manual
Note in manual built in nixpkgs repo

#### Fixed (current style)

Note after the fix

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -104,11 +104,11 @@
         "nixpkgs": "nixpkgs_3"
       },
       "locked": {
-        "lastModified": 1619640096,
-        "narHash": "sha256-gMnaDRsfLVmnBCoXwgjGJZmqBd/XUGpdZY5UkUEroEM=",
+        "lastModified": 1638915714,
+        "narHash": "sha256-w5qIrnLiaWm/YhIEpgwQrhr3dZgnBP//a7ueDSm63ho=",
         "owner": "NixOS",
         "repo": "nixos-common-styles",
-        "rev": "d5787c2a12922059712862e62ca2a324062b2954",
+        "rev": "fa19c7b909ae283851b1b580b1583f4a6375be29",
         "type": "github"
       },
       "original": {
--- a/scripts/bootstrapify-docbook.xsl
+++ b/scripts/bootstrapify-docbook.xsl
@@ -139,12 +139,24 @@
   <xsl:template name="alert-class">
     <xsl:param name="admonition-type"/>
     <xsl:choose>
+      <xsl:when test="$admonition-type = 'caution'">
+        <xsl:value-of select="'alert-danger'"/>
+      </xsl:when>
+      <xsl:when test="$admonition-type = 'danger'">
+        <xsl:value-of select="'alert-danger'"/>
+      </xsl:when>
+      <xsl:when test="$admonition-type = 'important'">
+        <xsl:value-of select="'alert-warning'"/>
+      </xsl:when>
       <xsl:when test="$admonition-type = 'note'">
         <xsl:value-of select="'alert-info'"/>
       </xsl:when>
-      <xsl:otherwise>
-        <xsl:value-of select="concat('alert-', $admonition-type)"/>
-      </xsl:otherwise>
+      <xsl:when test="$admonition-type = 'tip'">
+        <xsl:value-of select="'alert-info'"/>
+      </xsl:when>
+      <xsl:when test="$admonition-type = 'warning'">
+        <xsl:value-of select="'alert-warning'"/>
+      </xsl:when>
     </xsl:choose>
   </xsl:template>
 
@@ -158,7 +170,7 @@
 
     in which case we turn it into a single line as prefered by Bootstrap.
   -->
-  <xsl:template match="x:div[(@class='warning' or @class='note') and count(*) = 2]">
+  <xsl:template match="x:div[(@class='caution' or @class = 'danger' or @class = 'important' or @class = 'note' or @class = 'tip' or @class = 'warning') and count(*) = 2]">
     <xsl:element name="div">
       <xsl:attribute name="class">alert <xsl:call-template name="alert-class"><xsl:with-param name="admonition-type" select="@class" /></xsl:call-template></xsl:attribute>
       <strong><xsl:apply-templates select="x:h3[@class='title']/node()" />:</strong><xsl:text> </xsl:text><xsl:apply-templates select="x:p/node()" />
@@ -169,7 +181,7 @@
     Or there can be more elements, in which case we will switch it to “.alert-block” and
     change the heading size to h4 as expected by Bootstrap.
   -->
-  <xsl:template match="x:div[(@class='warning' or @class='note') and count(*) > 2]">
+  <xsl:template match="x:div[(@class='caution' or @class = 'danger' or @class = 'important' or @class = 'note' or @class = 'tip' or @class = 'warning') and count(*) > 2]">
     <xsl:element name="div">
       <xsl:attribute name="class">alert <xsl:call-template name="alert-class"><xsl:with-param name="admonition-type" select="@class" /></xsl:call-template> alert-block</xsl:attribute>
       <h4><xsl:apply-templates select="x:h3[@class='title']/node()" /></h4>
--- a/scripts/bootstrapify-docbook.xsl
+++ b/scripts/bootstrapify-docbook.xsl
@@ -130,23 +130,53 @@
 
   <xsl:template match="x:div[@class='titlepage' and ../@class = 'book']" />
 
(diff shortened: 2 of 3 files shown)


jtojnar:
@samueldr Looks like we no longer use bootstrap and alert-block and alert-warning is not implemented. Also there are some weird classes like `.alert.-warning` in the generated CSS stylesheet.

jtojnar:
The`.notice-box` appears to be defined in:

https://github.com/NixOS/nixos-common-styles/blob/d5787c2a12922059712862e62ca2a324062b2954/src/components/notice-box.less

But the theme does not appear to have a red colour I could use for the danger:

https://github.com/NixOS/nixos-common-styles/blob/d5787c2a12922059712862e62ca2a324062b2954/src/theme.less#L36-L52

Also maybe we should just use the common styles directly?

cc @garbas

jtojnar:
Curiously, Bootstrap 3 seems to [actually lack `alert-block`](https://getbootstrap.com/docs/3.4/components/#alerts) so we would need to take it from [Bootstrap 2](https://getbootstrap.com/2.3.2/components.html#alerts).

samueldr:
Since we're not actually using bootstrap, at all, I don't think we need to take `alert-block` from any of them.

Though, with that said, the styles for the manuals were authored with the intent to *stay compatible* with Bootstrap 2 semantics because of their usage in `bootstrapify-docbook`. Note that **yes it is Bootstrap 2 semantics**; the website before the re-design was using Bootstrap 2.

https://github.com/NixOS/nixos-homepage/blob/0e971a0d3994d4ce47a75ade7d914531018c03be/site-styles/pages/manual.less#L126-L135

it wasn't worth the effort to change the manual generation's implementation since *anyway* the move to a new Markdown toolchain is *just around the corner*.

I can update the styles to add missing compatibility classes. Can you make a list of what you think is missing for this PR, so I'm less likely to miss something?

And yes, it's RSCSS as you noticed.

jtojnar:
I can switch to using the `.notice-box` classes directly if desired. And since, unless we opt for the most bare-bones CommonMark dialect, the updated manual will still have admonitions (they are available in all documentation toolchains currently in consideration), I think it is not unreasonable to take admonitions into consideration.

At minimum we need `.alert-danger`/`.notice-box.-danger` that is red or similar. Here are a few admonitions from the manual to test the stylesheets on. We can of course adjust the XSLT stylesheet any way we need to better match our design language.

```html
<div class="alert alert-danger"><strong>Caution:</strong> 
Packages that are installed with your python derivation, are not accesible by <code class="literal">hy</code> this way.
</div>
<div class="alert alert-warning"><strong>Warning:</strong> 
This is a warning.
</div>
<div class="alert alert-info alert-block"><h4>On macOS</h4><pre class="programlisting hljs nix" xml:space="preserve">$ diskutil list
[..]
/dev/diskN (external, physical):
<span class="hljs-comment"><span class="hljs-comment">#:                       TYPE NAME                    SIZE       IDENTIFIER</span></span>
[..]
$ diskutil unmountDisk diskN
Unmount of all volumes on diskN was successful
$ sudo dd <span class="hljs-attr"><span class="hljs-attr">if=nix.iso</span></span> <span class="hljs-attr"><span class="hljs-attr">of=/dev/rdiskN</span></span>
</pre><p>
Using the 'raw' <code class="literal">rdiskN</code> device instead of
<code class="literal">diskN</code> completes in minutes instead of hours.
After <code class="literal">dd</code> completes, a GUI dialog "The disk
you inserted was not readable by this computer" will pop up,
which can be ignored.
</p></div>
<div class="alert alert-info alert-block"><h4>Note</h4><p>
Support for <code class="literal">NIXOS_LUSTRATE</code> was added in
NixOS 16.09. The act of "lustrating" refers to the
wiping of the existing distribution. Creating
<code class="literal">/etc/NIXOS_LUSTRATE</code> can also be used on
NixOS to remove all mutable files from your root partition
(anything that's not in <code class="literal">/nix</code> or
<code class="literal">/boot</code> gets "lustrated" on the
next boot.
</p><p>
lustrate /ˈlʌstreɪt/ verb.
</p><p>
purify by expiatory sacrifice, ceremonial washing, or some
other ritual action.
</p></div>
```

Then we might want to improve the styles maybe at least make the heading colour match the admonition colour but we probably should not block on that.

For comparison, this is what Bootstrap 2 did (as a side note `.alert-warning` has actually been introduced in Bootstrap 3, in Bootstrap 2, it was the default `.alert` style):

```css
.alert {
  padding: 8px 35px 8px 14px;
  margin-bottom: 20px;
  text-shadow: 0 1px 0 rgba(255, 255, 255, 0.5);
  background-color: #fcf8e3;
  border: 1px solid #fbeed5;
  -webkit-border-radius: 4px;
     -moz-border-radius: 4px;
          border-radius: 4px;
}

.alert,
.alert h4 {
  color: #c09853;
}

.alert h4 {
  margin: 0;
}

.alert .close {
  position: relative;
  top: -2px;
  right: -21px;
  line-height: 20px;
}

.alert-success {
  color: #468847;
  background-color: #dff0d8;
  border-color: #d6e9c6;
}

.alert-success h4 {
  color: #468847;
}

.alert-danger,
.alert-error {
  color: #b94a48;
  background-color: #f2dede;
  border-color: #eed3d7;
}

.alert-danger h4,
.alert-error h4 {
  color: #b94a48;
}

.alert-info {
  color: #3a87ad;
  background-color: #d9edf7;
  border-color: #bce8f1;
}

.alert-info h4 {
  color: #3a87ad;
}

.alert-block {
  padding-top: 14px;
  padding-bottom: 14px;
}

.alert-block > p,
.alert-block > ul {
  margin-bottom: 0;
}

.alert-block p + p {
  margin-top: 5px;
}
```

samueldr:
If everything from docbook is semantically represented correctly enough in the DOM you shared previously, it'll do. No need to review the manual DOM implementation until we choose the actual documentation toolchain IMO.

If you think things could be better for admonitions, please share, we're free to do **anything**. Don't worry about the specifics of the CSS/LESS implementation, just worry about the DOM, semantics and what's missing.

I'll wave the magic less wand and fix everything up. Nothing you've said seems to be an issue. Note that adding icons shouldn't require anything different than what's here.

jtojnar:
Just to sum up [Matrix discussion](https://matrix.to/#/!kjdutkOsheZdjqYmqp:nixos.org/$O3Vw0qdizz5P-VMtPUWJzfLwzeopAI6VmeUcOoPeIaw?via=nixos.org&via=matrix.org&via=nixos.dev), we decided to go with always having block “title” (even for single-paragraph admonitions) since we are no longer held by Bootstrap style guides. The transformed content might look something like this:

```css
aside.admonition-block.-${admonitionType}
    > div.title
    > p
    > pre
    > p
```

We are using `div` instead of `h3` for the title since it does not really fit into the document outline.

samueldr:
We'll use three styles, we don't really need more, and to be fair if I had to distinguish between tips, note, and important, it would be through iconography rather than colours.

image


Caution and Warning are the same, still I don't think they warrant different colours.

image


Danger is uniquely red, but not in use at this point in time.

image

samueldr:
Hmm, I would update the flake.lock file with the commit ID, but *somehow* it was chosen that squash and merge is better on the common styles repo, so the revision ID of the tip of the PR is not going to be present in the main branch once merged.

¯\\\_(ツ)\_/¯

jtojnar:
Looks good. The bottom padding appears to slightly larger than the top one in the screenshot but that is orthogonal to these changes.

We use at least one caution admonition so that one will use danger.

Why not merge the common styles first and then update flake.lock here to the final commit id?

samueldr:
> Why not merge the common styles first and then update flake.lock here to the final commit id?

The PR may take a while to get merged. It's blocked on requiring a review, and I'm not sure who else than @garbas can review. And I wanted the preview to be updated, but I see that you're not part of the homepage repo, so previews don't get generated for this PR...

garbas:
I assume this PR needs to be updated to latest `nixos-common-styles` revision.

garbas (approved):

PR #774, opened 2021-12-10
Always suggest multi-user install install

### LINKED ISSUE #616: Three different commands to install Nix?
**Describe the issue**

On the following pages, two different commands are described how to install Nix. Shouldn't these be the same commands?

**Page links**
- https://nixos.org/download.html#nix-quick-install

```
$ curl -L https://nixos.org/nix/install | sh
```

- https://nixos.org/guides/install-nix.html
```
sh <(curl -L https://nixos.org/nix/install) --daemon
```

Thanks

kubukoz:
Closes #616

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@
 
 /result*
 /styles
+.direnv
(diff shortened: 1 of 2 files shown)


garbas:
This was actually (also) fixed as part of #780

kubukoz:
lovely :)

PR #785, opened 2022-01-11
Remove installer flags deprecated as of Nix 2.4

Enzime:
As of Nix 2.4, the `--darwin-use-unencrypted-nix-store-volume` flag was
deprecated and prints a warning. Single user installations on macOS are
no longer supported so the `--daemon` flag is unnecessary.

### DIFF
--- a/download.tt
+++ b/download.tt
@@ -73,7 +73,7 @@
       </article>
 
       <article id="nix-install-macos">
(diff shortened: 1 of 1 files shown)


Enzime:
@garbas When they changed from `OS X` to `macOS` it's stylized with a lowercase M

PR #789, opened 2022-01-17
Add dev setup and sync SVG embed

samueldr:
(See: https://github.com/NixOS/nixos-common-styles/pull/8)

Huh... I hadn't intended to *optimize* the build process, but in verifying it wasn't making things worse, I realized that it improved build output sizes.

```
-r--r--r-- 1 root root 260K Dec 31  1969 old-site/styles/index.css
-r--r--r-- 1 root root 219K Dec 31  1969 new-site/styles/index.css

-r--r--r-- 1 root root  59K Dec 31  1969 old-site/styles/blog.css
-r--r--r-- 1 root root  30K Dec 31  1969 new-site/styles/blog.css

-r--r--r-- 1 root root 2.7M Dec 31  1969 old-site/styles/community.css
-r--r--r-- 1 root root 1.2M Dec 31  1969 new-site/styles/community.css
```

The actual important changes here:

 - New instructions and script to streamline hacking on `nixos-common-styles`
 - Use the new SVG embedding

These combined makes hacking on *anything* in the common styles a cinch. Things *just worked* (though were not exactly instant, but quick enough).

I have verified that this works by hacking on a still WIP branch where I actually implement the [requested features](https://github.com/NixOS/nixos-common-styles/issues/7) in storybook, and validate that changes aren't breaking the main site.

I also verified that changing SVGs affect the site live, without a further rebuild or manual step.

* * *

## TODO

 - [ ] Fix transcluded SVGs


flake.nix
```diff
@@ -119,7 +119,17 @@ rec {
             export NIX_STATE_DIR=$TMPDIR
 
             rm -f site-styles/common-styles
-            ln -s ${nixos-common-styles.packages."${system}".commonStyles} site-styles/common-styles
```
garbas, line 122: Since we will have to repeat this for status.nixos.org it might be good to distribute this functions with `nixos-common-styles`

scripts/nixos-common-styles-dev-setup.sh
```diff
@@ -0,0 +1,20 @@
...
+
+echo " :: Setting up development environment to use a local NixOS common styles checkout."
+
+if [ -e site-styles/common-styles ]; then
+	echo -n "Removing possibly stale symlink: "
+	rm -vf site-styles/common-styles
+fi
+
+echo -n "Linking with local checkout: "
+ln -vs "$(readlink -f "$local_checkout/src")" site-styles/common-styles
```
garbas, line 20: I wonder if we could do this just in the `serve` command. And instead of symlinking we would have a variable in `Makefile`. Then on a reload we would run `make COMMON_STYLES_DIR=...`.

garbas (commented):
I like the direction you are going. We might want to move some parts into `nixos-common-scripts` to be able to share them when starting work on `status.nixos.org`.

garbas:
This is not going to be needed with the new work.

PR #795, opened 2022-02-01
Fix broken URL to release notes

ooesili:
The URL for the nix package manager release notes does not include the
patch version when the patch version is `0`, i.e. on new minor and major
releases. This currently causes a broken link on download page to the
release notes of the latest version of nix.

For example, the download page currently generates a link for
`release-notes/rl-2.6.0.html`, however the actual URL for the release
notes is `release-notes/rl-2.6.html`.

This commit removes the patch version from the URL to the release notes
when the patch version is `0`.


garbas:
@ooesili Thank you! :1st_place_medal:

PR #808, opened 2022-02-24
Update commercial-support.html.in

drupol:
This should not be displayed in the website frontend.

### DIFF
--- a/community/commercial-support.html.in
+++ b/community/commercial-support.html.in
@@ -1,13 +1,3 @@
-
-  To start developing run:
-      serve
-
-  and go to the following URL in your browser:
-      https://127.0.0.1:8000/
-
-  It will rebuild the website on each change.
-
-
     <li>
       <a href="https://www.numtide.com/">
         <div>


samueldr:
Hi! :wave:

Uuuh... We already removed that, a couple of time... so definitely there's an underlying problem that needs fixing.

I've looked into it, no need for *you* to try your hand at fixing, I'm leaving the note for other website maintainers.

* * *

Ah ha!

@garbas it's added by the cron task.

 - https://github.com/NixOS/nixos-homepage/commit/1d46b32da85d3aa3481dc8f4ce3eb17fbe6672bd

Because of this:

 - https://github.com/NixOS/nixos-homepage/blob/c0b76ea536fdb532ad65105e7df09acd838cb4bd/scripts/update.sh#L14-L15

drupol:
How about using JavaScript for shuffling the data ? This would prevent rebuilding the site every hour and better for the environment.

I used that in here: https://not-a-number.io/projects/

WDYT ?

fricklerhandwerk:
#858

PR #810, opened 2022-03-01
fixing categories formatting

tomberek:


garbas:
I don't think this will actually works since `blog/categories.tt` is actually updated from `scripts/update-blog.py`.

Issue #833, opened 2022-04-22
How to contribute: help us help you


LoveIsGrief:
# The problem(s)

## Which files do I edit?

> if you are familiar with CSS look at the [issues tagged with design tag](https://github.com/NixOS/nixos-homepage/issues?q=is%3Aissue+is%3Aopen+label%3Adesign).

This is the only clear thing I will know to edit: CSS files. For content and bugs, I don't know where to start. There's no overview of the technologies used or how they are used. Is there a static site generator? Are things just loaded into a `public_html` directory and served using CGI scripts?  
If there's a problem, I probably have to find out first what we're using to then investigate which component could be failing, to then how it's failing and what can be done to resolve the issue. If everybody, individually first has to figure out what's being used, it takes away valuable time from the individual that they could have spent on the next step.

# `nix` requirement

The `README.md` just provides a block on how to run the homepage with live-reload. 

| README |
| ------------ |
| ![afbeelding](https://user-images.githubusercontent.com/2829538/164699189-b92b9918-81dc-4c65-82c6-b8a8be528a05.png) |

but in order to run it, I need `nix`. Were I just a passerby who saw a CSS error or a content error, I'd have to spend a good chunk of time first installing everything to just contribute minor changes. Especially for content, requiring more than an editor seems like overkill to me.

In my case, I'm still testing nixOS on another computer that doesn't have all my secrets, so booting that up just to contribute is pretty meh.

# Suggestions

## List the technologies used

Helps save time for contributors: they'll know which competencies to have in order to contribute to a certain aspect of the project.
The CSS line is a good start, but content and bugs definitely need more information.

I shouldn't have to read a bunch of different files to try and find out what this `.tt` extension is and what possibilities it provides (still don't know what it actually is and if I can get syntax highlighting for it).

## A `nix`-less contribution flow

Since I don't know which technologies are used, I'm assuming the site is a generated, static one.

It would be nice to be able to "just" clone the repo, make edits, push them and see them somewhere. Github actions should allow that pretty easily if I'm not mistaken.

## Overview of how things work together

As stated in Problems>"Which files do I edit?", it really helps to at least know which files are for what. Having an overview of how they are connected together is a nice to have, but definitely possible because [Github supports Mermaid diagrams](https://about.gitlab.com/handbook/tools-and-tips/mermaid/).

As a first step, it might even be just enough to have `README.md`s in the various folders

[...]

PR #837, opened 2022-04-30
Build Github pages for pull requests

LoveIsGrief:
Since this project generates a static site, one can take advantage of [Github Pages](https://pages.github.com/) to host it. In the case of this PR, it introduces the ability to review the the static site generated by pull requests.

Another benefit is that modifications need not require the lengthy installation of `nix` (related to #833)

# How it works

The PR will build the static site, modify the `pages` branch by adding a new directory with the name of the PR branch in the `/pages/` folder and updating `/index.html` with the list of directories therein, then finally push it. This allows for PRs to reviewed by navigating to the github pages URL (example below). 

# Required setup

 - Create `pages` branch
 - Use `pages` branch as the [source for Github Pages](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site) on this project

# Example result

LoveisGrief/nixos-homepage#1 showcases the [result](https://loveisgrief.github.io/nixos-homepage/)

# Possible pitfalls

**Branch name collision**

Since the branch name is used I'm not sure what the source branch names look like for forked repos. PRs with the same branch name might overwrite each other, but I'm not sure. I used [`github.head_ref`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context). Using `github.ref_name` gave me `1/merge` which I assume is the PR number with `/merge` (?).

### DIFF
--- a/.github/actions/generate-pages/action.yml
+++ b/.github/actions/generate-pages/action.yml
@@ -1,0 +1,31 @@
+name: Update pages
+description: Updates the page for the current branch
+runs:
+  using: composite
+  steps:
+    - name: Build
+      shell: bash
+      run: |
+        set -x
+        # Find the name of the branch/tag
+        # ref_name doesn't work for pull requests as head_ref points to the source ref
+        ref_name="${{ github.head_ref || github.ref_name }}"
+        tmp_page_dir="/tmp/${ref_name}"
+        
+        # Update the "webroot" for the generated homepage
+        repo_name="$(echo ${{ github.repository }} | cut -d / -f 2)"
+        sed -i -re "s|ROOT = .+|ROOT = \"/$repo_name/pages/$ref_name/\"|g" Makefile
+        grep ROOT Makefile
+        
+        # Build the homepage
+        nix build
+        
+        # Undo changes to Makefile
+        git checkout -- .
+        
+        # Checkout pages branch and be ready to commit the generated homepage
+        mkdir -p "$tmp_page_dir"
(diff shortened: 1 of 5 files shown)


garbas:
Thank you for putting the time into this, but the preview is already generated (published via Netlify) if branch originates from current repository. This used to be a limitation of GitHub at some point, but it looks like we could also enable it also for other PRs where branch does not originates from this repo.

garbas:
@LoveIsGrief Not correct. You don't need to a Netlify account to preview a website. You can either run a local development server or you run `nix-build` and serve `./result` folder locally (which what local development server is doing).

LoveIsGrief:
I'm thinking more of the PR review. Someone makes a PR, says they vertically centered a div. You want to have a look at the result.

You can either pull the branch and preview it locally, or (with my PR) go to the github page and see it live. Also, if you're on your phone or on the way and would like to have a look, there won't be a need to wait till you have access to a machine with nix.

Right now, that isn't possible, even if you were to remove the condition on the CI step that publishes to netlify. More work would probably have to be done, no? 
This PR removes the netlify dependency for PRs.

garbas:
As I said above, the preview is already created. But at the time github was missing some features regarding security we are not allowing just anybody to create a preview, just for PRs that are created from a branch on this repo. Now that GitHub has much better security story there we could consider enabling this also for other PRs.

There is no need to involve GitHub pages since Netlify provides this functionality for us.

PR #816, opened 2022-03-19
Local setup under macOs

lockejan:
another attempt to resolve https://github.com/NixOS/nixos-homepage/issues/515

After https://github.com/NixOS/nixos-homepage/pull/584 has been closed, I just tested it again and it still works.
The README on master evolved a bit since then, which is why I cleaned up the last proposal of @bxrt a bit.

It's not a completely outlined scenario how to hack on the project under macOS but it should suffice to get started @garbas .

PR #838, opened 2022-04-30
Add docker-compose

LoveIsGrief:
This PR removes the dependency on `nix` to be in the developer's environment and adds docker as an alternative.

Related to #833

# Reasoning

I'm probably not the only person here who would like to contribute to the homepage and nix/nixos documentation,
 but doesn't necessarily have `nix` itself installed on my dev machine or in my dev environment.
There are probably other usecases as well, but for me personally, I main a non nixOS distro (opensuse) and don't want to take
 my chances of messing up this install. So, nixOS is on a separate laptop which is purely for testing nixOS and doesn't have
 my developer secrets.

This helps me contribute and might help others contribute who just swing by and don't want to install yet another tool
 in order to contribute just some HTML or CSS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM nixos/nix
+
+RUN echo 'experimental-features = nix-command flakes' | tee -a /etc/nix/nix.conf
+
+WORKDIR /app
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+services:
+  nixos-homepage:
+    build:
+      context: .
+    environment:
+      HOST: "0.0.0.0"
+    ports:
+      - 8000:8000
+    volumes:
+      - .:/app
+    command:
+      - nix
+      - develop
+      - --command
+      - serve
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ If you feel lost where and how to contribute, ask the [marketing team](https://n
 
 # How to help?
 
-To run local development instance follow this steps:
+To run local development instance follow this steps to start a local server
 
     $ git clone git@github.com:NixOS/nixos-homepage.git
     $ cd nixos-homepage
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nixos/nix
+
+RUN nix-channel --add https://nixos.org/channels/nixpkgs-unstable \
+    && nix-channel --update \
+    && nix-env -iA nixpkgs.nixFlakes \
+    && echo 'experimental-features = nix-command flakes' | tee -a /etc/nix/nix.conf
```
garbas, line 6: Above run command should be replaced with:

```
RUN echo 'experimental-features = nix-command flakes' | tee -a /etc/nix/nix.conf
```

This is because `nixos/nix` image already has latest `nixpkgs-unstable` channel inside the image and latest nix version is already installed (currently that is 2.8.0). But I would also enable flakes, since I would use them later on.

docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    build:
+      context: .
+    environment:
+      HOST: "0.0.0.0"
+    ports:
+      - 8000:8000
+    volumes:
+      - .:/app
+    command:
+      - nix-shell
```
garbas, line 12: For command rather use `nix develop --command serve` to be also asked to use the binary cache to speed up the build.

garbas (changes requested):
This PR does not really remove the dependency on Nix. It just moved the Nix setup into docker image.

An attempt to do this was started in https://github.com/NixOS/nixos-homepage/pull/816 once.

Can you also document the current setup (bellow) the current one.

Dockerfile
LoveIsGrief: Done

docker-compose.yml
LoveIsGrief: done

garbas (approved):

PR #840, opened 2022-05-02
how nix works: reword introduction

fricklerhandwerk:
This is fallout from [documenting nix architecture](https://github.com/NixOS/nix/pull/6420), especially regarding a more general terminology which is hopefully more friendly to beginners.

I have more additions in the pipeline, and the feature sections could still use some love to add more detail to and stay consistent with the introduction, as well as getting more well-structured internally.

@garbas @thufschmitt @Ericson2314 @radvendii

### DIFF
--- a/guides/how-nix-works.tt
+++ b/guides/how-nix-works.tt
@@ -12,6 +12,28 @@
   <div>
     <h1 id="overview">Overview</h1>
     <p>
+      Nix can
+      <ul>
+        <li>
+          run any packaged program ad-hoc without installing it
+        </li>
+        <li>
+          establish temporary environments with all required dependencies
+        </li>
+        <li>
+          install programs and keep them up to date
+        </li>
+        <li>
+          provide reproducible, portable development environments, user environments, and entire operating system configurations.
+        </li>
+      </ul>
+    </p>
+
+    <p>
+      This is achieved as follows:
+    </p>
+
+    <p>
       Nix enforces <a hrefs="#complete-dependencies">complete, exact dependencies</a> both at build time and run time.
     </p>
     <p>
@@ -29,23 +51,6 @@
     <p>
       Nix build plans are declared in a <a href="#nix-language">purely functional configuration language</a>, the Nix expression language.
       It allows precise specification of build inputs and build instructions, and arbitrary composition of packages in a flexible, concise manner.
-    </p>
-    <p>
-      In practice, Nix can
-      <ul>
-        <li>
-          run programs ad-hoc without installing them
-        </li>
-        <li>
-          establish temporary environments with all required dependencies
-        </li>
-        <li>
-          install programs and keep them up to date
-        </li>
-        <li>
-          provide reproducible, portable development environments, user environments, and entire operating system configurations.
-        </li>
-      </ul>
     </p>
 
     <h1 id="features">Features</h1>
--- a/guides/how-nix-works.tt
+++ b/guides/how-nix-works.tt
@@ -66,10 +66,10 @@ We currently have no machinery to render mermaid diagrams directly.
 flowchart
   subgraph store
     direction LR
-    input[build input] --> |referenced by| drv[build step] --> |builds| out[build output]
+    input[build input] -.-> |referenced by| drv[build step] -.-> |builds| out[build output]
   end
-  cli[command line interface] --> |evaluates| expr[configuration language] --> |evaluates to| drv
-  cli --> |manages|store
+  cli[command line interface] -.-> |evaluates| expr[configuration language] -.-> |evaluates to| drv
+  cli -.-> |manages|store
 ```
 -->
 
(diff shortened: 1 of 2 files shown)


guides/how-nix-works.tt
```diff
@@ -11,40 +11,44 @@
...
-      they never change after they have been built. Nix stores packages in the
-      <em>Nix store</em>, usually the directory <tt>/nix/store</tt>, where each
-      package has its own unique subdirectory such as
+      Nix provides <em>reproducible builds</em> and <em>correct deployment</em> of build results by ensuring <a hrefs="#complete-dependencies">complete, exact dependencies</a> both at build time and run time.
     </p>
-
-    <pre class="file-listing">/nix/store/b6gvzjyb2pg0kjfwrjmg1vfhh54ad73z-firefox-33.1/</pre>
-
     <p>
-      where <tt>b6gvzjyb2pg0…</tt> is a unique identifier for the package that
```
edolstra, line 25: I feel that by removing the only technical detail on this page (namely how Nix stores packages), we can't really call this page "How Nix works" anymore. It becomes more a listing of features, but we already have https://nixos.org/ and https://nixos.org/explore.html, so we should figure out how this page is related to those.

guides/how-nix-works.tt
```diff
@@ -231,13 +241,24 @@
         system</a>.
     </p>
 
+    <h2 id="multi-user-support">Multi-user support</h2>
+
+    <p>
+      Starting at version 0.11, Nix has multi-user support. This means that
```
edolstra, line 247:
```suggestion
      Nix has multi-user support, meaning that
```

guides/how-nix-works.tt
fricklerhandwerk: > I feel that by removing the only technical detail on this page (namely how Nix stores packages), we can't really call this page "How Nix works" anymore.

Yes, the overview is supposed to be superficial. These details should go into the subsections, and I didn't want to touch them yet to keep the PR reasonable.

I'm also on the fence whether that page should instead be called "What Nix does", but that depends how we would  elaborate the subsections.

> It becomes more a listing of features, but we already have https://nixos.org/ and https://nixos.org/explore.html, so we should figure out how this page is related to those.

The front page is pretty clear IMHO, it's just the shortest possible description of what Nix can do. There still should be a complete list of major capabilities. Indeed https://nixos.org/explore.html overlap with the new introduction, and if you ask me they should be merged. Unfortunately having a design with illustrations baked in does not really allow for rearranging contents. Also, listing a disjoint set of features does not indicate how they are connected conceptually (that essentially everything falls out naturally from "complete, exact dependencies"), this is why I added that prose overview.

@garbas What does the marketing team think about this?

garbas: While [Explore](https://nixos.org/explore.html) is to answer a question "Why would I even look into Nix?", we can assume that visitors who come here are already interested into Nix and fall into 2 different groups:
1. Already have Nix installed and use it for simple tasks (eg. simple development environments). They waited some time to really see if Nix is the tool they want to invest into. Now they want to understand how things work.
2. Before they use any tool they want to understand how it works. They might now have time to go over +20 pages (eg. nix pills), but would love to read few pages to get the basic idea how things work.

Above two groups are why this page is needed. So *How Nix/NixOS works* page can be technical. We should be able to explain in few pages basic ideas how Nix/NixOS works. We don't have to focus on *Why this is good* too much. I personally think you can explain much more (and much quicker) with a nice graphic. You don't have to design the graphic just need to provide the idea and I'll try to reach out to the designer to do the job.

Important to know about the current content of the page. Current content of the page is a copy/paste from two older pages which talked about Why you should use Nix. Ideally would would completely rewrite this page with much better explanations.

garbas:
@fricklerhandwerk Would you be interested in bigger rewrite of this page?

fricklerhandwerk:
@garbas Yes, I have more material in the making, and really I don't want to leave that page in such an inconsistent state.

guides/how-nix-works.tt
fricklerhandwerk: Yes, as said I think technical details should go into subsections. I would also like to add diagrams. Right now the quickest path would be to drop in SVG files, but that is not very easy to maintain.

garbas:
Could this PR be merged into #841 content wise. I think the content of both PRs could be a provide a nice opening.

fricklerhandwerk:
@garbas Could we merge this? Further documentation work will produce more updates to the feature sections eventually. Otherwise this PR will hang around for more weeks.

fricklerhandwerk:
@garbas Could we merge this? Further documentation work will produce more updates to the feature sections eventually. Otherwise this PR will hang around for more weeks.

Ericson2314:
Should we mere this? Let's not let the perfect be the enemy of the good.

fricklerhandwerk:
Closing this since that work continues on the Nix manual. Maybe we can include that section on the web site once it's good enough.

PR #841, opened 2022-05-03
How Nix works: add architecture overview

fricklerhandwerk:
I still have no idea how to make `<aside>` visible. @garbas any help on the LESS?


guides/how-nix-works.tt
```diff
@@ -236,6 +236,67 @@
...
+    </p>
+
+<!--
+We currently have no machinery to render mermaid diagrams directly.
+
+```mermaid
+flowchart
+  subgraph store
+    direction LR
+    input[build input] --> |referenced by| drv[build task] --> |builds| out[build output]
```
garbas, line 252: `-->` in this line breaks the HTML comment

garbas:
@fricklerhandwerk I'm not sure how would you like to render aside element, can you share an example or explain with words. The less code is in `./site-styles/` folder. I'm not sure we have some specific styles for this page, but if we need to we can add it.

I think the architecture section that you added at the end of the "Nix" part, should actually be merged with the first paragraph. I think that would be a great opening if we properly merge the content of both paragraphs.

garbas (changes requested):

fricklerhandwerk:
@garbas `aside` should be rendered like warnings or notes in the nix manual, [see an example](https://nixos.org/manual/nix/stable/advanced-topics/distributed-builds.html). Essentially a colored background box with padding. (Using `blockquote` as currently in the manual is the wrong data type.)

Will combine the PRs once I get around to it, have it close to the top of the stack.

fricklerhandwerk:
Closing, merged into #840.

PR #861, opened 2022-06-04
Fix self link in RSS feed.

kevincox:
Previously it pointed at a URL which just redirected here.


garbas:
@kevincox Thank you for the PR and sorry for late review.

PR #868, opened 2022-06-21
mention you can install other desktops

raboof:
I saw there was some confusion about this somewhere, though I don't remember where...


garbas:
@raboof Thank you!

PR #872, opened 2022-07-01
Update docker example (nixpkgs#94636)

roberth:
Avoid https://github.com/NixOS/nixpkgs/issues/94636

To summarize, adding packages to `contents` is bad, because it duplicates self-referential packages.
Instead use

    contents = buildEnv { ... };

Or keep it simple like this example with Cmd or Entrypoint.

This container has only command, so we don't bother with
PATH and make it the Entrypoint so it receives the container
args as process args.


roberth:
Only `buildImage` suffers from the problem, not `buildLayeredImage`.

PR #882, opened 2022-08-11
make nix.dev official

fricklerhandwerk:
as decided here:
https://github.com/NixOS/nix.dev/issues/285#issuecomment-1198131047

### DIFF
--- a/learn.tt
+++ b/learn.tt
@@ -209,21 +209,21 @@
 
 <section class="learn-otherresources">
   <div>
-    <h1>Other learning resources</h1>
+    <h1>Learning resources</h1>
     <ul>
+      <li class="clickable-whole">
+        <h3>nix.dev</h3>
+        <p>
+          Guides to getting things done in the Nix ecosystem.
+        </p>
+        <a href="https://nix.dev">Read more</a>
+      </li>
       <li class="clickable-whole">
         <h3>Unofficial Wiki</h3>
         <p>
           A user-maintained wiki for Nix and NixOS.
         </p>
         <a href="https://nixos.wiki">Read more</a>
-      </li>
-      <li class="clickable-whole">
-        <h3>nix.dev</h3>
-        <p>
-          An unofficial and opinionated guide for developers getting things done using the Nix ecosystem.
-        </p>
-        <a href="https://nix.dev">Read more</a>
       </li>
       <li class="clickable-whole">
         <h3>Nix Pills</h3>


edolstra:
I don't think we should make nix.dev "official" yet (or at least clarify what "official" means). It's currently opinionated in ways that would have major ramifications if they become the official position of the Nix project - e.g. one of the first sections is about pinning Nixpkgs using `niv`. There is also a policy issue around commercial services: nix.dev promotes cachix, which is understandable since it's Domen's site. But if nix.dev becomes official we should decide what we would do if others want to add references to their commercial services.

fricklerhandwerk:
Yes, please clarify what "official" means. It's a blocker to centralizing documentation without re-doing enormous amounts of work.

Here is the relevant issue to track this: https://github.com/NixOS/nix.dev/issues/277

houstdav000:
Coming here from a related thread of project "support"/recommendations as in [awesome-nix](https://github.com/nix-community/awesome-nix).

Given the mission statement:
> The NixOS Foundation's role is to support the infrastructure and development of the NixOS project as a whole. The exact nature and scope of this might vary over time.

From <https://github.com/NixOS/nixos-foundation>, it is left unclear what "support", "NixOS project", and "private work" mean. It seems that was intentional at the time for flexibility, but the lack of clear boundaries now leaves not only what makes a resource "official", but what kind of actions can be "official".

It's clearly unreasonable for a single organization to maintain _everything_ in a community, but also potentially stifiling. In that regard, it may be useful to consider what others are doing. Canonical, for example, breaks up their resources across multiple forums, a discourse, governance bodies, blogs, and wikis.

There are already other sites too with _semi_-community-driven efforts, like [nix-community.org](https://nix-community.org) and [buildwithnix.org](https://builtwithnix.org).

Perhaps it might be best instead of converting every resource to "official" ones, to define scope on what's official and provide a designated "community website", whether that be the wiki, one of the above, someone's site, or a new `community.nixos.org` site that's under the `nix-community` purview (if that's an actual group of people and not just a repository catch-all. Not actually sure myself).

Either way, it sounds like a scope-definition issue imo.

domenkozar:
> I don't think we should make nix.dev "official" yet (or at least clarify what "official" means). It's currently opinionated in ways that would have major ramifications if they become the official position of the Nix project - e.g. one of the first sections is about pinning Nixpkgs using `niv`. There is also a policy issue around commercial services: nix.dev promotes cachix, which is understandable since it's Domen's site. But if nix.dev becomes official we should decide what we would do if others want to add references to their commercial services.

`niv` article was written before flakes were even released as Nix. I'm fine flakes replacing `niv` of course, once they are a stable feature.

As for the commercial services, I wouldn't be transferring all my work to the community expecting a monopoly (I find it a bit weird to even propose that).

asymmetric:
Making it official should IMO include moving it under nixos.org.

fricklerhandwerk:
Closing since the "Learn" button links directly to nix.dev and the bottom bar is full.

PR #913, opened 2022-10-11
60,000 -> 80,000 packages in nixpkgs

tuxiqae:
According to https://github.com/NixOS/nixpkgs/blob/cfaec958cc81014eae1b8f35827c7c46b4c74207/README.md there are more than 80,000 packages in the nixpkgs repository. 

(According to https://repology.org/repository/nix_stable_22_05 there are about 86K total.)

### DIFF
--- a/index.tt
+++ b/index.tt
@@ -92,7 +92,7 @@
       The Nix Packages collection
       <strong>(<a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>)</strong>
       is a set of
-      <strong class="-highlighted">over 80&#160;000 packages</strong> for the Nix package manager.
+      <strong class="-highlighted">over 80&nbsp;000 packages</strong> for the Nix package manager.
     </p>
     <form method="get" action="https://search.nixos.org/packages">
       <div>
--- a/index.tt
+++ b/index.tt
@@ -92,7 +92,7 @@
       The Nix Packages collection
       <strong>(<a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>)</strong>
       is a set of
-      <strong class="-highlighted">over 60&#160;000 packages</strong> for the Nix package manager.
+      <strong class="-highlighted">over 80&#160;000 packages</strong> for the Nix package manager.
     </p>
     <form method="get" action="https://search.nixos.org/packages">
       <div>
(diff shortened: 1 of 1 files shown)


index.tt
```diff
@@ -92,7 +92,7 @@
       The Nix Packages collection
       <strong>(<a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>)</strong>
       is a set of
-      <strong class="-highlighted">over 60&#160;000 packages</strong> for the Nix package manager.
+      <strong class="-highlighted">over 80&#160;000 packages</strong> for the Nix package manager.
```
fricklerhandwerk, line 95: Not sure why `&nbsp;` was used here.

```suggestion
      <strong class="-highlighted">over 80,000 packages</strong> for the Nix package manager.
```

tuxiqae: I just saw this commit 86b2e28c9aace45485dacdc7e302870bb1254c64

tuxiqae: Looks good, thanks

fricklerhandwerk: @tuxiqae Nope, sorry, this broke the build. For some reason something in there doesn't accept `&nbsp;`. Would be great if you could force-push your original version over this, then I'll merge it.

tuxiqae: Sure, but I will only be able to do so in a couple of hours.

samueldr: The reason is that `&nbsp;` is HTML, and this is handled through an XML-based tooling.

index.tt
```diff
@@ -92,7 +92,7 @@
       The Nix Packages collection
       <strong>(<a href="https://github.com/NixOS/nixpkgs">Nixpkgs</a>)</strong>
       is a set of
-      <strong class="-highlighted">over 60&#160;000 packages</strong> for the Nix package manager.
+      <strong class="-highlighted">over 80&nbsp;000 packages</strong> for the Nix package manager.
```
garbas, line 95:
```suggestion
      <strong class="-highlighted">over 80&#160;000 packages</strong> for the Nix package manager.
```

garbas (approved):

tuxiqae:
@garbas Don't you prefer to remove everything but the first commit and merge that?

garbas:
> @garbas Don't you prefer to remove everything but the first commit and merge that?

Not really... I could just use GitHub UI to revert this change from my phone.

PR #937, opened 2022-10-31
Adding nixcon 2022 to the list of nixcon conferences

garbas:


garbas:
hmm ... i still need to fix the logo. but first need to put kids to bed.

Issue #698, opened 2021-04-18
/nixos/manual/unstable/options.html redirects to /manual/nixos/stable/unstable/options.html


sternenseemann:
This should of course instead redirect to `/manual/nixos/unstable/options.html`. Seems like an oversight while setting up these redirects.

timfenney:
Was `/nixos/manual/unstable/options.html` reachable before? What happened?

timfenney:
As I mentioned in my Draft PR (https://github.com/NixOS/nixos-homepage/pull/980), **it seems we need the netlify cli tool for the netlify redirect rules to take place in development.**

@fricklerhandwerk do you have any idea how to proceed with this? I followed the instructions to fire up the nix shell for development, but I don't have a netlify cli in my PATH it seems.

timfenney:
Unfortunately this hasn't been super-clear to me.

This issue claims this redirect is a problem:
`/nixos/manual/unstable/options.html` -> `/manual/nixos/stable/unstable/options.html`

However, when I look at @garbas commit above, it seems that this rule is causing the behaviour:
```
a68a4e2ce0 (Rok Garbas         2020-08-21 14:27:49 +0200 163) [[redirects]]
a68a4e2ce0 (Rok Garbas         2020-08-21 14:27:49 +0200 164)   from = "/nix/manual/*"
a68a4e2ce0 (Rok Garbas         2020-08-21 14:27:49 +0200 165)   to = "/manual/nix/stable/:splat"
a68a4e2ce0 (Rok Garbas         2020-08-21 14:27:49 +0200 166)   status = 302
a68a4e2ce0 (Rok Garbas         2020-08-21 14:27:49 +0200 167)   force = true
```
... but since "unstable manuals were added" with this commit, it seems like the issue is complaining about something that was never a viable URL?

@garbas did this URL exist prior to this commit here? If not I think we can close this :sweat_smile:

sternenseemann:
The URL did exist as proven by my browser history, it was a viable URL before the website was redesigned (so it predates this repository, I think).

PR #980, opened 2023-02-03
Add netlify redirect rule.

timfenney:
This is an effort to fix https://github.com/NixOS/nixos-homepage/issues/698

My theory of what is going on:
- the redirects are supposed to redirect from `unstable` URLs to `stable`
- the failure mentioned in the issue is because `unstable` is being captured and passed into the URL to be redirected to, via the `splat`
- since the first matching rule is applied (only), I added a more specific rule above the rule which I think caused the problem

**It seems like we need the netlify cli tool for netlify.toml redirect rules to take effect in development.**

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -173,6 +173,12 @@
   force = true
 
 [[redirects]]
+  from = "/nixos/manual/unstable/*"
+  to = "/manual/nixos/stable/:splat"
+  status = 302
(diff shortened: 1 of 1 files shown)


timfenney:
Aha I can test on an Ubuntu machine ... I will test the redirect rule then.

PR #999, opened 2023-03-02
add flox to commercial support page

jurraca:
flox is a tool built on Nix, providing a CLI and platform for sourcing from Nixpkgs, building with Nix, and publishing software in a declarative way. We also partner with organizations looking to leverage Nix and flox.


fricklerhandwerk (approved):
This is legit. A third party should confirm though, since @garbas is now a flox employee. @zimbatm?

garbas:
But we need to fix the build first. It looks like it was broken some time ago.

garbas:
I've fixed the CI build&deploy github action and I've reopened an updated PR. It is here #1000

PR #1014, opened 2023-03-13
Link to installation section

### LINKED ISSUE #1013: There is no information on how to run the installer
**Introduce your stance**

The Homepage currently says: 

> The minimal installation ISO image does not contain the graphical user interface, and is therefore a lot smaller. You have to run the installer from the console. It contains a number of rescue tools

(under `Download > Minimal ISO image`)

Emphasis: "run the installer from the console".

But it does not say _how_ to actually run the installer.

**Page links**

https://nixos.org/download.html

---

Reported to me in person by @p-fruck

matthiasbeyer:
We should not only tell users that they need to run the installer, we also need to tell them how to. Thus, link to the install section of the manual here.

---

Closes #1013 
CC @fricklerhandwerk

### DIFF
--- a/download.tt
+++ b/download.tt
@@ -194,7 +194,10 @@
 
         <p>The minimal installation ISO image <strong>does not contain the
             graphical user interface</strong>, and is therefore a lot smaller.
-          You have to <strong>run the installer from the console</strong>. It
+          You have to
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Thanks a lot!

PR #1021, opened 2023-03-23
 Add a description meta tag 

garbas:
Continuation of #1016

### DIFF
--- a/layout.tt
+++ b/layout.tt
@@ -36,6 +36,7 @@ BLOCK navigationLink %]
   <link rel="shortcut icon" type="image/png" href="/favicon.png" />
   <meta property="og:image" content="/logo/nixos-lores.png" />
   <meta name="google-site-verification" content="ir-07nYvo3u3x_VmkTO1wCfYJ8uC-SrVBGR7hZgqPSE" />
+  <meta name="description" content="A package manager and operating system providing an unique approach to packaging and system configuration" />
 
   [%# This has to happen ASAP, even before jQuery is initialized. %]
   <script>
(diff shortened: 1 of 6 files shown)


garbas:
@fricklerhandwerk @gilice Do you maybe have time to do a quick review?

layout.tt
```diff
@@ -24,10 +24,23 @@ BLOCK navigationLink %]
...
+
 <html lang="en" class="without-js">
 
 <head>
-  <title>[% title %]</title>
+
+  [% IF title != "" %]
+  <title>[% title %] | Nix &amp; NixOS </title>
+  [% ELSE %]
+  <title>Nix &amp; NixOS | Reproducible builds and deployments</title>
```
fricklerhandwerk, line 39:
```suggestion
  <title>[% title %] | Nix </title>
  [% ELSE %]
  <title>Nix | Reproducible builds and deployments</title>
```

Except for the odd announcement, the current web site is exclusively about Nix and clearly not about NixOS. We have unambiguous evidence that beginners have a hard time distinguishing the different software components and their purpose and role. It of course needs more work to introduce the other projects and systems, but let's keep it focused on Nix for now.

fricklerhandwerk (approved):
I assume the code works, cannot test it right now.

layout.tt
garbas: I'm leaving both in the title as we are also offering the download of NixOS.

And meanwhile we'll figure out the messaging on the website to introduce both Nix and NixOS ideas.

PR #1022, opened 2023-03-23
Display versioned manuals on nixos.org

jlesquembre:
Generate documentation for the latest 3 nix versions. 3 is a random choice, we can change how many versions we want to display.

Screenshot:

screenshot__2023-03-23_161527

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,6 +93,24 @@ $(NIX_MANUAL_STABLE_OUT): $(call rwildcard, $(NIX_MANUAL_STABLE_IN), *)
 	mkdir -p $(NIX_MANUAL_STABLE_OUT)
 	cp --no-preserve=mode,ownership -RL $(NIX_MANUAL_STABLE_IN)/* $(NIX_MANUAL_STABLE_OUT)
 
+NIX_MANUAL_2_12_IN ?= /no-such-path
+NIX_MANUAL_2_12_OUT = manual/nix/2.12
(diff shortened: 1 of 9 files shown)


garbas (changes requested):
👍 on the intent to show 5 version of documentation for Nix. And thank you for putting in the work @jlesquembre !

But we should improve the implementation, because it puts more work on the marketing team to keep everything in sync. I hope to explain this bellow.

In current situation Nix team has full control what gets deployed to the website, by just pushing to certain branches -> `master` and `latest-release`. Then an hourly github action will update the website.

I would like to keep this workflow.

As far as I can understand this PR, this would not be the case anymore, since if a patch release is done for older versions we would need to manually intervene and bump it. This also means that each release we would have to manually bump the versions (since we are showing 5 versions with this PR).

Can we figure it out that his happens automatically without any intervention from anybody.

Maybe this version bump could happen in `scripts/update.sh` (this is the hourly github action script) keeping website up-to-date.

jlesquembre:
@garbas thanks for your feedback.
I added a script to automate version updates, called from `scripts/update.sh` as you suggested. 
The version list is in the `nix_versions.json` file. 
The UI is still the same as in the screenshot, there are 5 different versions. But documentation for more versions is generated. The goal is to prevent dead links when new nix versions are released.

fricklerhandwerk (approved):
LGTM

garbas:
Sorry for taking long to review this, but I have some time planned to review this over the coming weekend.

jlesquembre:
@garbas no worries, thanks for reviewing it :)

PR #1030, opened 2023-04-03
Added wavelens company to community

DerDennisOP:
Wavelens is currently also using NixOS for their servers.


fricklerhandwerk:
That page is about companies providing commercial support for Nix users, right @garbas? In that case this addition would be  off topic.

garbas (commented):
> That page is about companies providing commercial support for Nix users, right @garbas? In that case this addition would be off topic.

Yes, this page is meant to list companies providing commercial support for Nix users.

PR #1038, opened 2023-04-12
Add nixcademy.com to commercial support page

tfc:


garbas (approved):

tfc:
That was quick. Thank you very much @garbas!

PR #1043, opened 2023-04-19
treewide: add mastodon to footer

gilice:
I didn't actually check if the icon works, so don't merge without checking


garbas:
@fricklerhandwerk did you check if icon is correct? If I remember correctly this PR does not add it.

fricklerhandwerk:
image

garbas:
@fricklerhandwerk that was not the change. The footer does not have an icon.

fricklerhandwerk:
Oh indeed, that Mastodon icon was on the community page.

Issue #854, opened 2022-05-30
Hourly update [ci skip] spamming the commits


davidak:
e.g. https://github.com/NixOS/nixos-homepage/commit/45a5e300664f0e3817d5fcad18feceec187a3b28

you can't find actual commits anymore

fricklerhandwerk:
The logo shuffling should definitely be on the front-end. If possible we should prune the history, even if it breaks clones.

drupol:
Indeed, I think the shuffling should be made in Javascript, on the front-end.

This would:

1. Prevent spamming the commit log with hourly updates
2. Foster eco-programming and spare energy

Am I the only one worrying about those details?

JulienMalka:
I have always found this situation to be ridiculous, I am favorable to changing this.

drupol:
I just submitted a PR to fix this, let me know what you think.

drupol:
This issue is fixed, it can now be closed.

PR #1052, opened 2023-05-13
Use client-side shuffle on "Commercial support" page

drupol:
This PR:

- Is related to #854 
- Add a Javascript shuffling the section, the shuffling is updated every hour
- Rename the script `shuffle-commercial-providers` into `update-commercial-providers`

Additional things:
- JSFiddle script: https://jsfiddle.net/gb63joLy/

:warning: There will be conflict with the file `community/commercial-support.html.in` during the lifetime of this PR because that file gets updated hourly randomly... hence the conflicts. If it gets accepted, I'll rebase it properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -255,7 +255,7 @@ index.html: blog/announcements-rss.xml blog/index.html
 community/commercial-support.html: community/commercial-support.html.in
 
 community/commercial-support.html.in: community/commercial-support.toml
-	shuffle-commercial-providers --input community/commercial-support.toml > $@
+	update-commercial-providers --input community/commercial-support.toml > $@
 
 
 ### Check
--- a/js/nixos-site.js
+++ b/js/nixos-site.js
@@ -1,3 +1,25 @@
+$.fn.shuffleChildren = function() {
+  $.each(this.get(), function(index, el) {
+      var random = function(s = Date.now()) {
+        return function() {
+            s = Math.sin(s) * 10000;
+
+            return s - Math.floor(s);
+        };
+      };
+
+      var $el = $(el);
+      var $find = $el.children();
+
+      $find.sort(function() {
+          return 0.5 - random((new Date).getHours());
+      });
+
+      $el.empty();
+      $find.appendTo($el);
+  });
+};
+
 $(function () {
   // Special "pseudo-global" variable to track whether we are synthetically
   // activating an event. In that case some semantics are different.
@@ -324,4 +346,5 @@ $(function () {
   // This way tabs and panes are active as intended.
   handleNavigation();
 
+  $("section.community-commercial-support > ul").shuffleChildren();
 });
--- a/js/nixos-site.js
+++ b/js/nixos-site.js
@@ -1,22 +1,22 @@
-$.fn.shuffleChildren = function() {
+// FisherYates implementation
+$.fn.shuffleChildren = function(seedParam = Date.now()) {
+  var randomish = function(s) {
+    s = Math.sin(s) * 10000;
+    return s - Math.floor(s);
+  };
+
   $.each(this.get(), function(index, el) {
-      var random = function(s = Date.now()) {
-        return function() {
-            s = Math.sin(s) * 10000;
+    var $el = $(el);
+    var $find = $el.children();
+    var len = $find.length;
+    var seed = randomish(seedParam);
 
-            return s - Math.floor(s);
-        };
-      };
+    while (--len) {
+      let randIndex = Math.floor(seed * (len + 1));
+      [$find[randIndex], $find[len]] = [$find[len], $find[randIndex]];
+    }
 
-      var $el = $(el);
-      var $find = $el.children();
-
-      $find.sort(function() {
-          return 0.5 - random((new Date).getHours());
-      });
-
-      $el.empty();
-      $find.appendTo($el);
+    $el.empty().append($find);
   });
 };
 
@@ -346,5 +346,5 @@ $(function () {
   // This way tabs and panes are active as intended.
   handleNavigation();
 
-  $("section.community-commercial-support > ul").shuffleChildren();
+  $("section.community-commercial-support > ul").shuffleChildren((new Date).getHours());
 });
--- a/scripts/update-commercial-providers.py
+++ b/scripts/update-commercial-providers.py
@@ -1,5 +1,4 @@
 import click
-import random
 import toml
 
 TEMPLATE = """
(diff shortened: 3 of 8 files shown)


scripts/update-commercial-providers.py
```diff
@@ -22,8 +22,6 @@ def main(input):
     with open(input) as f:
         providers = toml.load(f)["commercial-provider"]
 
-    random.shuffle(providers)
```
jtojnar, line 25: You can also remove the import.

Makefile
```diff
@@ -255,7 +255,7 @@ index.html: blog/announcements-rss.xml blog/index.html
 community/commercial-support.html: community/commercial-support.html.in
 
 community/commercial-support.html.in: community/commercial-support.toml
-	shuffle-commercial-providers --input community/commercial-support.toml > $@
+	update-commercial-providers --input community/commercial-support.toml > $@
```
jtojnar, line 258: It should be possible to just have

```suggestion
community/commercial-support.html: community/commercial-support.toml
	update-commercial-providers --input community/commercial-support.toml > $@
```

and remove the `.in` file from git.

scripts/update.sh
```diff
@@ -11,7 +11,7 @@ nix flake lock \
   --update-input nix-pills \
   --update-input nix-dev
 
-echo "Shuffle commercial providers ..."
+echo "Updating commercial providers..."
 nix develop --command shuffle-commercial-providers --input community/commercial-support.toml > community/commercial-support.html.in
```
jtojnar, line 15: This should become unnecessary then.

drupol: Removed.

Makefile
drupol: Removed.

scripts/update-commercial-providers.py
drupol: Done.

Makefile
drupol: Removing `community/commercial-support.html.in` is not a good idea since it is included in `
community/commercial-support.html`.

js/nixos-site.js
```diff
@@ -1,3 +1,25 @@
+$.fn.shuffleChildren = function() {
+  $.each(this.get(), function(index, el) {
+      var random = function(s = Date.now()) {
```
fricklerhandwerk, line 3: What's wrong with just shuffling randomly on page load? Would be simpler. The general idea is to present companies without preferred ordering, so it shouldn't matter how exactly that is achieved.

fricklerhandwerk (approved):
Thank you so much, this was long overdue. @garbas when this merged, could we please do the unthinkable for once, and prune the commit history on master?

js/nixos-site.js
drupol: Depends if we want to have the same behavior as before. Before this PR, the page was updated every hour.
By seeding the PRNG with the current `hour` of the day, we make sure that the order will change every hour.
If you don't want that behavior and have a different order at each page load, I can restore the original version I did, I don't mind.

fricklerhandwerk: Let @garbas take a look, he's ultimately responsible.

js/nixos-site.js
```diff
@@ -1,3 +1,25 @@
...
+            return s - Math.floor(s);
+        };
+      };
+
+      var $el = $(el);
+      var $find = $el.children();
+
+      $find.sort(function() {
+          return 0.5 - random((new Date).getHours());
+      });
```
pacien, line 16: This looks awfully similar to the biased [Microsoft shuffle].
The [Fisher-Yates shuffle] would be a safer bet :)

[Microsoft shuffle]: https://www.robweir.com/blog/2010/02/microsoft-random-browser-ballot.html
[Fisher-Yates shuffle]: https://en.wikipedia.org/wiki/Fisher-Yates_shuffle#The_modern_algorithm

pacien (changes requested):

js/nixos-site.js
drupol: I just pushed a commit implementing the Fisher Yates algo.

js/nixos-site.js
```diff
@@ -1,3 +1,25 @@
+// FisherYates implementation
+$.fn.shuffleChildren = function(seed = Date.now()) {
+  var random = function(s) {
```
pacien, line 3: Nitpick: could you rename this to `randomish`, to underline that the
distribution is not uniform, and to avoid shadowing the browser's `random`?

js/nixos-site.js
```diff
@@ -1,3 +1,25 @@
...
+    return s - Math.floor(s);
+  };
+
+  $.each(this.get(), function(index, el) {
+    var $el = $(el);
+    var $find = $el.children();
+    var len = $find.length;
+
+    while (--len) {
+      let randIndex = Math.floor(random(seed) * (len + 1));
```
pacien, line 14: Here, the "random" source is called (re-initialised) with the seed at each
turn. That doesn't seem right.

pacien (changes requested):

js/nixos-site.js
pacien: The seed changes at each hour, but each call also re-initialisis the sequence
it seems? That should be clealer in context:
https://github.com/NixOS/nixos-homepage/pull/1052#discussion_r1193202657

js/nixos-site.js
drupol: Ok I got your point. Will fix that.

js/nixos-site.js
drupol: Browser's random is `Math.random()`. Should we really rename it?

js/nixos-site.js
pacien: The function is called with `(new Date).getHours()`, setting the seed to that.
This approach generates 24 permutations, one for each hour in a day.
But then, we only cycle through those same permutations every day.

More fundamentally, it might just be preferable to just drop this hourly 
change and just shuffle on each page load instead.

This remove the need to implement a custom sequence PRNG in JS,
which is bad for everyone's sanity.

drupol: IMHO, I also thing this is the best option, see JSFiddle: https://jsfiddle.net/gb63joLy/1/

garbas:
@drupol Thank you for taking the time to do this.

I'm going to close this PR and open it in some branch in this repo. This way we will get a preview.

Makefile
jtojnar: It is just an intermediary step. The suggested change above should make it unnecessary.

PR #1056, opened 2023-05-16
Use client-side shuffle on "Commercial support" page

garbas:
Was #1052

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -255,7 +255,7 @@ index.html: blog/announcements-rss.xml blog/index.html
 community/commercial-support.html: community/commercial-support.html.in
 
 community/commercial-support.html.in: community/commercial-support.toml
-	shuffle-commercial-providers --input community/commercial-support.toml > $@
+	update-commercial-providers --input community/commercial-support.toml > $@
 
 
 ### Check
--- a/js/nixos-site.js
+++ b/js/nixos-site.js
@@ -1,3 +1,25 @@
+$.fn.shuffleChildren = function() {
+  $.each(this.get(), function(index, el) {
+      var random = function(s = Date.now()) {
+        return function() {
+            s = Math.sin(s) * 10000;
+
+            return s - Math.floor(s);
+        };
+      };
+
+      var $el = $(el);
+      var $find = $el.children();
+
+      $find.sort(function() {
+          return 0.5 - random((new Date).getHours());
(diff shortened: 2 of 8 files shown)


js/nixos-site.js
```diff
@@ -1,3 +1,25 @@
...
+    var seed = randomish(seedParam);
+
+    while (--len) {
+      let randIndex = Math.floor(seed * (len + 1));
+      [$find[randIndex], $find[len]] = [$find[len], $find[randIndex]];
+    }
+
+    $el.empty().append($find);
+  });
+};
```
pacien, line 21: (Repeating/rephrasing my comments from the previous pull request:)

For the Fisher-Yates shuffle to work, a new random number has to be used
each time an index is picked for swapping elements.

Also, the sinful function (used here to work around the lack of seedable PRNG
in the web browser) isn't fair.

How about dropping the hourly seed thing and just use `Math.random()` in the
`while` loop?

pacien (changes requested):

js/nixos-site.js
pacien: This: https://github.com/NixOS/nixos-homepage/pull/1052#discussion_r1193377878

drupol: I support this and I think that we should drop the hourly seed thingie.

Makefile
```diff
@@ -255,7 +255,7 @@ index.html: blog/announcements-rss.xml blog/index.html
 community/commercial-support.html: community/commercial-support.html.in
 
 community/commercial-support.html.in: community/commercial-support.toml
```
jtojnar, line 257: Copying my comment from https://github.com/NixOS/nixos-homepage/pull/1052#discussion_r1193009005:

> It should be possible to just have
> 
> ```suggestion
> community/commercial-support.html: community/commercial-support.toml
> ```
> 
> and remove the `.in` file from git.

js/nixos-site.js
fricklerhandwerk: @drupol can you update this on your branch and reopen the original PR? Then we can merge something at all. I will hit the button if you ping me, let's fix forward at the worst.

drupol: I created a new PR @ https://github.com/NixOS/nixos-homepage/pull/1081

PR #1062, opened 2023-05-29
download.tt: Mention WSL systemd support

doronbehar:


download.tt
```diff
@@ -92,6 +92,13 @@
 
       <article id="nix-install-windows">
         <h2>Windows (WSL2)</h2>
+        <h3>Multi-user installation (Requires WSL with systemd enabled)</h3>
+        <p>
+          WSL versions 0.67.6 and above has systemd support. Follow <a href="https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/">this guide</a> to configure it, and then install Nix using:
```
fricklerhandwerk, line 97:
```suggestion
          WSL versions 0.67.6 and above has systemd support. Follow <a href="https://devblogs.microsoft.com/commandline/systemd-support-is-now-available-in-wsl/">Microsoft's systemd guide</a> to configure it, and then install Nix using:
```

The link text should be self-describing.

fricklerhandwerk (approved):
Thanks! Small wording suggestion.

Could you please also add this to the Nix manual and nix.dev? It's unfortunate that we have three places for this, but until we decide on one source of truth we unfortunately have to maintain all of them.

doronbehar:
> Thanks! Small wording suggestion.
> 

Committed!

> Could you please also add this to the Nix manual

Regarding the Nix manual, what page have you had in mind? I was thinking only of:

```diff
diff --git i/doc/manual/src/installation/supported-platforms.md w/doc/manual/src/installation/supported-platforms.md
index 8ca3ce8d4..768259a44 100644
--- i/doc/manual/src/installation/supported-platforms.md
+++ w/doc/manual/src/installation/supported-platforms.md
@@ -5,3 +5,5 @@ Nix is currently supported on the following platforms:
   - Linux (i686, x86\_64, aarch64).
 
   - macOS (x86\_64, aarch64).
+
+  - Windows Subsystem for Linux
```

> [...] and nix.dev?

Seems good to me:

https://nix.dev/tutorials/install-nix

> It's unfortunate that we have three places for this, but until we decide on one source of truth we unfortunately have to maintain all of them.

Indeed it's unfortunate - I had no idea nix.dev contained that page and I don't see why would we need it.

fricklerhandwerk:
> Regarding the Nix manual, what page have you had in mind?

- [Installation](https://github.com/NixOS/nix/blob/bed2fe23123134420027e46866ef6ce334fd231c/doc/manual/src/installation/installation.md) (that one also needs an update of the `curl | sh` lines)
- [Installing binary](https://github.com/NixOS/nix/blob/bed2fe23123134420027e46866ef6ce334fd231c/doc/manual/src/installation/installing-binary.md)

Looking at all of that after quite some time – ugh, these pages really need some work.

> I had no idea nix.dev contained that page and I don't see why would we need it.

nix.dev is now the official documentation resource. I'd rather remove the Nix installation instructions from nixos.org and link the *Download* button to nix.dev. It would be a first step to reduce the duplication. @NixOS/documentation-team and @NixOS/marketing-team what do you think?

06kellyjac:
I know this is about installing nix multi user to generic WSL but is it worth also mentioning NixOS-WSL?
https://github.com/nix-community/NixOS-WSL

I guess it counts as 3rd party so we might be hesitant to do so. nix-community is already referred to but only within the blog section

---

Also just to clarify it's WSL**2** but is versioned as having an `0.` Major? (`0.67.6`)
Or was systemd support in previous WSL releases?

garbas:
@doronbehar I'd rever the change back to WSL2 ... https://learn.microsoft.com/en-us/windows/wsl/install#check-which-version-of-wsl-you-are-running

Since Nix only works with WSL2.

garbas:
@06kellyjac Can we make it official? Put it in the Windows store/marketplace? Do you want to help out with this?

fricklerhandwerk:
@Garbas we'll fix the broken tutorial links and revert their removal to keep the status quo (that was simply an accident), but we still have to make a call on long-term maintenance. 

Multiple sources of truth for installation instructions produced a tangible (albeit not terrible) amount of work in the past months. I'd like to get down to one instance, doesn't matter to me where it lives in principle. The Nix manual is the candidate where it would likely make most sense, as it would be self-contained.


The first steps tutorials on nixos.org also require additional maintenance and cognitive overhead, and while we continue developing the presentation on nix.dev, they will keep looking careless because some things simply don't translate well, and no one has the time or motivation to take care of that.

fricklerhandwerk:
@Garbas we'll fix the broken tutorial links and revert their removal to keep the status quo (that was simply an accident), but we still have to make a call on long-term maintenance. Do you have a concrete counter proposal that you can commit to maintain over at least the next half year?

Multiple sources of truth for installation instructions produced a tangible (albeit not terrible) amount of work in the past months. I'd like to get down to one instance, doesn't matter to me where it lives in principle. The Nix manual is the candidate where it would likely make most sense, as it would be self-contained.


The first steps tutorials on nixos.org also require additional maintenance and cognitive overhead, and while we continue developing the presentation on nix.dev, they will keep looking careless because some things simply don't translate well, and no one has the time or motivation to take care of that.

06kellyjac:
> @06kellyjac Can we make it official? Put it in the Windows store/marketplace? Do you want to help out with this?

Interesting. I don't use Windows or NixOS-WSL so I wouldn't be very helpful here but I'll CC some of the people working on that repo to get their thoughts: @nzbr @k900

K900:
We could, but it would require some infra effort to handle signing and such, and the benefits are questionable.

fricklerhandwerk (approved):

nzbr:
> > @06kellyjac Can we make it official? Put it in the Windows store/marketplace? Do you want to help out with this?
> 
> Interesting. I don't use Windows or NixOS-WSL so I wouldn't be very helpful here but I'll CC some of the people working on that repo to get their thoughts: @nzbr @K900

It'd be nice to get it to the MS Store some day, but for that nix-community/NixOS-WSL/pull/94 needs to be finished, and we'd need to get some sort of official NixOS account on there. We're probably going to aim for being on Scoop and maybe Winget first

garbas:
@nzbr Thank you for working on this.

Geting Nix/NixOS to any of the MS Store/Scoop/Winget would mean being available to many new users. I wonder if @NixOS/foundation would be willing to sponsor it, or maybe we open a fundraising like for the [documentation team](https://opencollective.com/nixos/projects/documentation-project).

Would somebody from the NixOS-WSL contributors be interested to work on this? If needed I (and rest of the @NixOS/marketing-team) can help with fundraising campaign.

K900:
The hard part with getting on the Store isn't really funding, it's getting the infra set up to handle signing and such.

PR #1077, opened 2023-06-19
index: replace Reproducible with Declarative.

Foxboron:
NixOS does not support reproducible builds* so having this be the first word gives people the wrong impression considering how well-known this project is.

"Declarative" conveys the intent better.

* As defined by the Reproducible Builds projects

### DIFF
--- a/index.tt
+++ b/index.tt
@@ -33,7 +33,7 @@
 <section class="home-hero">
   <div>
     <div class="blurb">
-      <h1>Reproducible builds <span>and deployments.</span></h1>
+      <h1>Declarative builds <span>and deployments.</span></h1>
       <p>
         Nix is a tool that takes a unique approach to package management and
         system configuration. Learn how to make reproducible, declarative and


fricklerhandwerk (approved):

Foxboron:
To be clear; if there is a better replacement for "Reproducible Builds"  in the context of NixOS, I'm all ears. I was contemplating "Reproducible Environments" but a longer tagline might need more UX considerations.

garbas (changes requested):
Not true, there are different way of reproducibility. Ask 10.people you'll probably get 10 different answers. 

The topic of messaging is going to be discussed at the next marketing meetings.

Foxboron:
Not at all. It is quite clearly defined.

>Reproducible builds are a set of software development practices that create an independently-verifiable path from source to binary code

https://reproducible-builds.org/

See also https://reproducible-builds.org/docs/definition/

This is currently a work in progress for NixOS, but not something that is supported to the point where it causes confusion like this:  https://en.wikipedia.org/w/index.php?title=Comparison_of_Linux_distributions&diff=prev&oldid=1151180812

garbas:
I don't argue that is not clearly defined. What I'm arguing is that peoples perception of if something is reproducible varies quite a bit. I've heard numerous times Docker images are reproducible.

Also - last time we discussed the messaging on the front page - the `Declarative` word doesn't capture everything we want to say. And so we went forward with `Reproducible`.

Now the discussion about the new messaging was hanging around for some time. But with the recent work (eg, survey & soon newsletter, social postings, ...) just being done, we'll be able to prioritize it. What I can assure you is that we will stay away from the word `Reproducible` for numerous reasons.

Foxboron:
The loose usage from projects like NixOS has contributed to this perception,  which is part of the point here.

I'm happy that "Reproducible Builds" will either be reworded or have a less prominent place on the homepage.

Foxboron:
And to be clear, the issue is not `Reproducible`. The issue is `Reproducible Builds`.

Foxboron:
It's been almost 3 months now.

What is the status?

fricklerhandwerk:
@Garbas I agree with the change an merge this now for the sake of not letting people hang. But I'm not religious about this, feel free to revert if you strongly disagree.

Foxboron:
@fricklerhandwerk Thanks!

SeqBra:
This should be reversed, nixos allows for reproducible builds, phrases and even words can have multiple definitions

piegamesde:
Please start a new discussion in the issue tracker or on the forum

PR #1088, opened 2023-07-06
ci: `cron`: Update from hourly to daily

drupol:
Hi,

This opinionated PR proposes the following changes:

1. Changes the `cron` workflow from running `hourly` to `daily`
2. Remove obsolete options that are enabled by default on some actions
3. Reformat the file using `prettier`: `nix run nixpkgs#nodePackages.prettier -- --write .github/workflows/cron.yml`

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -1,7 +1,7 @@
 name: "Hourly Build & Deploy to Netlify"
 on:
   schedule:
-  - cron: '0 * * * *'
+  - cron: '0 0 * * *'
 jobs:
   hourly-build-and-deploy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/cron.yml
```diff
@@ -1,63 +1,59 @@
...
-        fetch-depth: 0
+      - name: Checking out the repository
+        uses: actions/checkout@v3
 
-    - name: Installing Nix
-      uses: cachix/install-nix-action@v22
-      with:
-        nix_path: nixpkgs=channel:nixos-unstable
-        extra_nix_config: |
-          experimental-features = nix-command flakes
```
fricklerhandwerk, line 20: Why do we disable `flakes`? Doesn't this break the... everything?

drupol: It is enabled by default.

garbas (approved):

PR #1102, opened 2023-08-10
learn: link to First Steps tutorial

osslate:
The link to "First steps with Nix" point to nix.dev. This changes the link to a more specific tutorial on nix.dev titled "First Steps", which makes more sense given the title of the button.


garbas (approved):
Good catch. Thank you.

PR #1117, opened 2023-09-07
Remove Black River Software [no longer operating] from service providers

eamsden:
I'm no longer operating Black River Software, so it should be removed from the list of Nix/NixOS service providers.


fricklerhandwerk:
Merging after a due diligence facts check. Thanks for keeping the website up to date!

eamsden:
Thank you for merging quickly!

PR #1122, opened 2023-09-16
Add responsive footer element

tfc:
As discussed in the meeting, this is my attempt to mimic the original footer of nixos.org, but described with tailwind CSS.

### DIFF
--- a/src/layouts/Footer.astro
+++ b/src/layouts/Footer.astro
@@ -1,6 +1,82 @@
 ---
+import { Icon } from 'astro-icon'
 ---
 
 <footer>
-  Footer
+  <style>
+    .nixlogo {
+      content: "";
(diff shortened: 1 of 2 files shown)


src/layouts/Footer.astro
```diff
@@ -1,6 +1,82 @@
 ---
+import { Icon } from 'astro-icon'
 ---
 
 <footer>
-  Footer
+  <style>
+    .nixlogo {
```
tfc, line 7: not sure what's the best way to do this. The good part about this is that it is defined in the same file as it's used. but if we need this svg somewhere else too, we would need a more central place.
Of course, it could also just be an  tag that links to an svg file... not sure why it is like this in the original nixos.org implementation?

thilobillerbeck (approved):

thilobillerbeck:
looks good :+1: 

since this is already in main, there is no need to merge imo

garbas:
@tfc awesome to have so quick turn around. I've protected the `main` branch going forward so we go via review process. I'll close this PR since it is already in `main`.

src/layouts/Footer.astro
garbas: I think we should use the custom icon set as described in astro-icon.

PR #1124, opened 2023-09-18
Domen has stepped down from the moderation team

alyssais:
Thanks for your work Domen!


domenkozar:
Due the team doing really good work and lack of time on my end, I better focus my efforts in the foundation.

Thank you for doing this (usually) unthankful job 🙏

PR #1129, opened 2023-09-22
index: reword Nix selling points

orangecms:
This clarifies the conceptual understanding of Nix and adds links
to external references and the Nix wiki.

### DIFF
--- a/index.tt
+++ b/index.tt
@@ -70,7 +70,7 @@
         <h2>Reproducibility</h2>
         <p>
           Nix <strong>builds packages in isolation</strong> from each other to enable
-          <a href="https://reproducible-builds.org/">reproducible builds</a>.
+          <a href="https://reproducible.nixos.org/">reproducible builds</a>.
           The <a href="https://nixos.wiki/wiki/Binary_Cache">Nix cache</a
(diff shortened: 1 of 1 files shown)


orangecms:
Whether the proposed change of order is desirable / sensible might be debatable; my idea is that reproducibility builds strongly on top of the declarative approach, and that in turn is the main point of Nix.

The suggestion here is to stress the goals and the means, offering direct entry points to the project to make it more accessible.

orangecms:
I just found https://reproducible.nixos.org/ - I'd want to substitue the reference to the external reproducible-builds.org for it, since it explains it as well and show the reproducibility status.

garbas:
Sorry for long review delay, but I finally came around to review it.

Right now (for the next month) the focus with the website is to switch to a more "modern" and setup. This was holding us back for some time. Immediately after that we are going to look how to change the messaging on the website. If you have time I would welcome you to join the [marketing team](https://nixos.org/community/teams/marketing) and join the discussion.

For the above reason I will close this PR and consider it in a month. Thank you @orangecms for the effort and caring. I hope you join the marketing team and help us promote Nix/NixOS even better.

PR #1140, opened 2023-10-30
commercial-support: fix url for serokell

Mic92:
Somehow this got mixed up with mayflower. I am not re-adding mayflower since I have sources that most (all?) NixOS-related people will have left the company by next year.

### DIFF
--- a/community/commercial-support.tt
+++ b/community/commercial-support.tt
@@ -105,7 +105,7 @@ methodology and workflow
     </li>
 
     <li>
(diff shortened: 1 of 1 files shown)


WilliButz (approved):

globin (approved):

drupol (commented):
Looks like serokell is listed twice ?

Mic92:
> Looks like serokell is listed twice ?

I don't think so? Where do you see that?

PR #1141, opened 2023-10-31
link to nix.dev for installation docs

mightyiam:
Another small step toward showing users the way into nix.dev.

### DIFF
--- a/learn.tt
+++ b/learn.tt
@@ -8,7 +8,7 @@
   <ul>
     <li>
       <div class="clickable-whole">
-        <a href="[%root%]download.html#download-nix">
+        <a href="https://nix.dev/install-nix">
           [% PROCESS svg path="site-styles/assets/gfx-learn-install.svg" %]
           <button>Install Nix</button>
         </a>


garbas (changes requested):
I'm 👎 for this change since both pages contain the same information.

I assume this is the area where _"tensions"_ between docs and marketing team arise. We need to agree what is in the domain of either team. I'm not sure how to structure this discussion, but any idea is welcome. Maybe starting with a simple call would be a start and we can see where it will take us.

I'm not saying that docs team shouldn't document how to install Nix, especially the advance/detailed version of the installation. For the happy path I think this should stay on the main website. But lets discuss this further...

fricklerhandwerk:
I argue we should do this for at least two reasons:
- We currently still have at least three places where installation is shown.

  In the past months there was at least one instance where this was super annoying: someone had a very good change that made the shell command compatible with `fish`, and we had to update those three places in different repos, and missed the fourth one. Much later I removed the one from the Nix README which now points to nix.dev.

  We'll likely have to update it again and again if we move forward with the DetSys installer. I'm also considering pointing people to the same Nix release that is shipped with NixOS, because using Home Manager, where Nix is pinned to that version in NixOS, can break fresh installs that have a newer Nix. So there will be churn.

  Let's have one place for the happy path (nix.dev is on a good way to provide the complete happy path) and one for all the details (that's the Nix manual).

- The @NixOS/marketing-team is objectively much less active than Nix maintainers and the documentation team. I would rather have you focus on your priorities rather than running this kind of day-to-day noise by you.

- We're already moving all the Nix-specific documentation, including the manual, to nix.dev, and all of that is actively maintained. Let's just continue this, it seems to work. I would very much like to see nixos.org eventually expand on NixOS, because we don't really have anyone taking care of that and no other obvious place to put that information except for the NixOS manual (and that still needs a lot of love before one can work on it for real).

garbas:
> I argue we should do this for at least two reasons:
> 
>     * We currently still have at least three places where installation is shown.
>       In the past months there was at least one instance where this was super annoying: someone had a very good change that made the shell command compatible with `fish`, and we had to update those three places in different repos, and missed the fourth one. Much later I removed the one from the Nix README which now points to nix.dev.

I agree there should be one place for this and that place should be on the nixos.org website and detailed instructions in the documentation.

>       We'll likely have to update it again and again if we move forward with the DetSys installer. I'm also considering pointing people to the same Nix release that is shipped with NixOS, because using Home Manager, where Nix is pinned to that version in NixOS, can break fresh installs that have a newer Nix. So there will be churn.

That can be easily changed in current setup. We currently follow what is released by the Nix team. I don't mind changing that.

>       Let's have one place for the happy path (nix.dev is on a good way to provide the complete happy path) and one for all the details (that's the Nix manual).

I disagree nix.dev should be the place for this. But I also think nix.dev should be a redirect to nixos.org and that the docs subdomain should be used to not confuse the user that that is documentation. 

>     * The @NixOS/marketing-team is objectively much less active than Nix maintainers and the documentation team. I would rather have you focus on your priorities rather than running this kind of day-to-day noise by you.

I don't consider this day-to-day noise. I consider this fundamental disagreement in direction. That is why I called for a call where we could align our idea and directions.

>     * We're already moving all the Nix-specific documentation, including the manual, to nix.dev, and all of that is actively maintained. Let's just continue this, it seems to work. I would very much like to see nixos.org eventually expand on NixOS, because we don't really have anyone taking care of that and no other obvious place to put that information except for the NixOS manual (and that still needs a lot of love before one can work on it for real).

I think this is our main point of disagreement and I think you know it, since we discussed it many times. As I wrote I'd like discuss the direction we are taking here since I fundamentally disagree with it.

mightyiam:
Thank you.

PR #1145, opened 2023-11-05
Astro/donate page [RFR]

djacu:
- [x] Waiting on [$astro/landingpage](https://github.com/NixOS/nixos-homepage/tree/astro/landingpage) to merge to take advantage of the Button component.
- [x] Need to fix styling on hyperlink tags.
- [x] Need to fix styling for top header and paragraph.
- [x] Fix external logos on small width screens.

PR #1147, opened 2023-11-07
Astro/landingpage [WIP]

thilobillerbeck:
There are still some components missing but to continue on with the other pages (#1145 for example) by using this pages components it would be a good idea to merge the current state into main.


djacu (approved):
lgtm

tomberek (approved):

PR #1148, opened 2023-11-07
nix.dev: Fix for non-flake

infinisil:
See https://github.com/NixOS/nix.dev/pull/761


fricklerhandwerk:
We should instead remove the dependency altogether: https://github.com/NixOS/nixos-homepage/pull/1149

PR #1149, opened 2023-11-07
remove dependency on nix.dev

fricklerhandwerk:
it's not used for anything


infinisil:
Oh damn you're right, it was removed in ec82512afc3dec6e242ea32bbce700b7d8ef6832

tomberek (approved):

PR #1156, opened 2023-11-15
Update moderation team

piegamesde:
- piegames joined (https://discourse.nixos.org/t/welcome-piegames-to-the-moderation-team/35469)
- Alyssa is leaving
- grahamc is leaving

### DIFF
--- a/community/teams/moderation.tt
+++ b/community/teams/moderation.tt
@@ -21,7 +21,6 @@ href="https://github.com/NixOS/rfcs/blob/master/rfcs/0102-moderation-team.md">RF
   <aside>
     <h2>Members</h2>
     <ul>
-      <li><a href="https://discourse.nixos.org/u/grahamc">Graham Christensen</a> (gchristensen)</li>
       <li><a href="https://discourse.nixos.org/u/mic92">Jörg Thalheim</a> (Mic92)</li>
       <li><a href="https://discourse.nixos.org/u/hexa">Martin Weinelt</a> (Hexa)</li>
       <li><a href="https://discourse.nixos.org/u/garbas">Rok Garbas</a> (garbas)</li>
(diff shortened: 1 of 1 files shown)


ryantm (approved):

mweinelt (approved):

alyssais (approved):

grahamc:
And welcome pie!!! You’re going to do great, and be a valued member I’m sure. What a good day!

community/teams/moderation.tt
```diff
@@ -21,12 +21,12 @@ href="https://github.com/NixOS/rfcs/blob/master/rfcs/0102-moderation-team.md">RF
   <aside>
     <h2>Members</h2>
     <ul>
-      <li><a href="https://discourse.nixos.org/u/qyliss/">Alyssa Ross</a> (qyliss / alyssais)</li>
       <li><a href="https://discourse.nixos.org/u/grahamc">Graham Christensen</a> (gchristensen)</li>
```
grahamc, line 24: I just realized I forgot to formalize my stepping down. Mind incorporating that here? I haven't been an active member since early 2023.

```suggestion
```

alyssais: Would you mind also demoting yourself from admin and leaving the relevant Matrix rooms?  (#moderators:nixos.org, #moderationteam:nixos.org)

Because you're an admin, we can't remove you, and it's not good if people who're no longer on the team are still idling in the rooms, where they can see confidential reports and stuff.

grahamc: For sure, should be all set. Can you double check?

By the way, we should also:

1. identify any rooms I'm admin of where mjolnir isn't an admin
2. demote me from admin to default in all rooms where mjolnir is an admin

This is like a million rooms, though, and there is no chance I could do this by hand. Is that mjolnir "invite and setup" script still around? Might be able to modify it.

mweinelt: It is still here; https://github.com/grahamc/matrix-management-tools/blob/main/process-all-rooms.sh

We could never use it, because the @admin:nixos.org account isn't server admin, and we therefore couldn't rely on `synadm`.

garbas (approved):

PR #1163, opened 2023-11-22
fix mastodon icon in footer

garbas:

PR #1167, opened 2023-11-27
Allow triggering manual updates

infinisil:
This enables the [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) trigger, allowing the daily updates to be trigger manually from https://github.com/NixOS/nixos-homepage/actions/workflows/cron.yml

I want this so I can link people to new sections in the Nixpkgs/NixOS manual without having to wait for the daily update.

Also, so I can force an update already now, since updates were broken [for 5 days](https://github.com/NixOS/nixos-homepage/pull/1163#issuecomment-1827677297).

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -2,6 +2,8 @@ name: "Hourly Build & Deploy to Netlify"
 on:
   schedule:
     - cron: "0 0 * * *"
+  workflow_dispatch:
+
 jobs:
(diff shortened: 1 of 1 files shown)


infinisil:
Thanks! Just started a manual run: https://github.com/NixOS/nixos-homepage/actions/runs/7005172022

PR #1183, opened 2023-12-06
docs team: Remove @zmitchell

infinisil:
Asked to be removed from the team due to time constraints, thanks for all the contributions!


fricklerhandwerk:
Thanks @zmitchell! You moved the whole thing forward by a lot. Would be great to see you around some time again.

PR #1185, opened 2023-12-07
Several Astro Rework Enhancements

thilobillerbeck:
This PR changes the following things:

- adjust default tailwind font sizes to be mroe in line with the currently live NixOS page
- adjust the container width to match up with the original page
- make `<head>` section be generated through the `Astro/SEO` plugin and match up with the original page
- update several dependencies (mainly Astro to version 4)
- add 404 page and adjust the Layout to always cover the full page vertically
- omit `Vegur `font and render the "NixOS Foundation" logo as a path-only SVG (this is the only thing we used Vegur for)
- Switched `Overpass `and `Roboto `to their variable font counterparts to be more flexible with font rendering and save up on network traffic as well as import complexity
- add some aria labels in the layout where they were missing
- compress HTML in the final output to reduce bundle size

Just a quick heads up regarding the font, since the variable font allows for more weights what is now `light` will become `extralight`, this maybe leads to some changes needed to be made in the resulting page, but this is not a huge deal imo and the only thing that is affected by changing `Roboto ` to `Roboto Flex`.


djacu (approved):
lgtm

PR #1188, opened 2023-12-09
Astro/Fix Tag Component

djacu:
Fix Tag component:
- so hrefs work.
- uses the correct 'nix blue' value.
- so the background fits the contents. Without the `w-fit` class it looks like the bottom tag instead of the top.

Screenshot 2023-12-08 at 19-46-50 Explore

### DIFF
--- a/src/components/ui/Tag.astro
+++ b/src/components/ui/Tag.astro
@@ -22,6 +22,13 @@ const {
 const localClassList = `${colorMap[color]} w-fit rounded-lg font-bold !text-white !no-underline text-sm px-[0.5rem] py-[0.25rem]`;
 ---
 
-<a href={href} class:list={[localClassList, classList]}>
-  <slot/>
-</a>
+{href
+  ?
+    <a href={href} class:list={[localClassList, classList]}>
+      <slot/>
+    </a>
+  :
+    <span href={href} class:list={[localClassList, classList]}>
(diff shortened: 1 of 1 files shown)


src/components/ui/Tag.astro
```diff
@@ -15,11 +16,12 @@ const colorMap = {
...
 
-const localClassList = `${colorMap[color]} rounded-lg font-bold text-white text-sm px-[0.5rem] py-[0.25rem]`;
+const localClassList = `${colorMap[color]} w-fit rounded-lg font-bold !text-white !no-underline text-sm px-[0.5rem] py-[0.25rem]`;
 ---
 
-<span class:list={[localClassList, classList]}>
+<a href={href} class:list={[localClassList, classList]}>
   <slot/>
-</span>
+</a>
```
thilobillerbeck, line 27: I guess we should render either span or a if href is present or not since especially accesibility tools such as screenreaders will always communicate there is an anchor tag present even if it is not one.

```
{href ? <a ..../> : <span ...../>}

thilobillerbeck (changes requested):

src/components/ui/Tag.astro
djacu: Good idea

thilobillerbeck (approved):

PR #1194, opened 2024-01-11
doc team: Add @danielsidhion

infinisil:


DanielSidhion (approved):
Approving to show I accept to be part of the team!

PR #1197, opened 2024-01-11
as mentioned on chat removing roberth

henrik-ch:
@roberth


roberth (approved):
Thank you for taking care of this @henrik-ch.

For context, I'm involved in too many things right now, which makes me disproportionately less effective. I won't be too far away, and I do still want to help with module system docs for instance, at some point, but being on this list creates a non-zero expectation that I can't meet for the time being.

PR #1199, opened 2024-01-12
workflows/cron: Pin Nix to 2.19.2, fix updates

infinisil:
The current default for cachix/install-nix-action@v24 is 2.19.1, which annoyingly completely removed support for `--update-input`, which the homepages udpate script was relying upon. Only 2.19.2 reinstated support for that flag with a deprecation notice.

Also by pinning we prevent such problems in the future. Notably in this case this caused the docs on nixos.org to not update **for 2 months**.

I tested the script locally. After merging I'll manually trigger the action

### DIFF
--- a/.github/workflows/cron.yml
+++ b/.github/workflows/cron.yml
@@ -16,6 +16,9 @@ jobs:
         uses: cachix/install-nix-action@v24
         with:
           nix_path: nixpkgs=channel:nixos-unstable
+          # Pin Nix so that CI doesn't randomly break again
+          # And we don't want 2.19.1 because that doesn't have backwards compatible support for --update-input
+          install_url: https://releases.nixos.org/nix/nix-2.19.2/install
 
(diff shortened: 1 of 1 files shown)


DanielSidhion (approved):

infinisil:
Also annoyingly, the update script now _requires_ Nix version 2.19 or greater, anything older doesn't work. 

Actually let me update CI to just pin Nix 2.19.2 instead, that's gonna be simpler and better for the future. It also makes the update script work for the Nix versions people actually have installed (since 2.19 isn't even the default in NixOS)

DanielSidhion:
I think the current solution to pin the version to 2.19.2 in CI is good to go, although we may have to revisit this in the future again when the flag is removed. Hopefully by then we'll be able to switch to `nix flake update` without the chance of someone being on a version of nix old enough not to support it.

PR #1213, opened 2024-01-31
Astro/fix package lock

djacu:
The package-lock.json file was changed in #1208. This should fix it.


djacu:
Picked the wrong base so checks aren't running correctly. Closing and re-opening.

PR #1218, opened 2024-02-04
well-known: delegate matrix client/server to matrix.nixos.org

mweinelt:
We are moving away from EMS and towards self-hosted infrastructure.

cc https://github.com/NixOS/nixos-org-configurations/pull/336


mweinelt:
Needs a rebase on main.

garbas:
Moved to #1367

PR #1220, opened 2024-02-06
Update the steering committee

infinisil:
See https://discourse.nixos.org/t/rfc-steering-committee-rotation-2023-24/36197/13?u=infinisil

Ping @GetPsyched @jonringer @Mmesch


GetPsyched:
Let's keep my username lowercase as well in the spirit of the rest?

infinisil:
Those are the canonical usernames as they are on Discourse, I don't think's it's important to keep them lowercase, but I don't really mind :)

PR #1222, opened 2024-02-14
fix [commercial-support]: remove enlambda from service providers

hugosenari:
The URL isn't working anymore.

https://www.enlambda.com/

@fricklerhandwerk :-D


garbas:
@tomberek can we also adjust apply this in `main` branch?

PR #1225, opened 2024-02-23
Include manuals + pills as post build step

garbas:
and remove not used files


garbas:
Still on the TODO list is listing the AMIS

thilobillerbeck:
> [removed a little too much :)](https://github.com/NixOS/nixos-homepage/pull/1225/commits/54e536d12d4869ad518df8ac15e7037f76b39853)

haha, I was just about to complain that dev mode does not work, but this seems to be fixing it :D

thilobillerbeck:
@garbas what about the `site-styles` folder? Did you find anyhting that breaks when removing it? Otherwise I think we can omit it as well.

garbas:
> @garbas what about the `site-styles` folder? Did you find anyhting that breaks when removing it? Otherwise I think we can omit it as well.

`site-styles` folder is removed

garbas:
I think this is now ready for review and potential merge once issues are addressed.

thilobillerbeck (approved):
looks good

PR #1244, opened 2024-03-01
Update events

tommorris:
London group seems to have relocated to Eventbrite.

Removing Lausanne and Seattle as they seem to have closed.


garbas:
Migrated PR to #1366

PR #1301, opened 2024-03-04
Fix link on Download page

garbas:


thilobillerbeck:
holding this for now since I also do some breaking changes on /download

PR #1311, opened 2024-03-06
Fix image scalings and rework download page

### LINKED ISSUE #1259: [Techstack-beta] Download: no graceful degradation without JavaScript
On the current implementation, the content presentation is enhanced for JavaScript use only when JavaScript is available. This leads to the whole content being visible by default if scripts are disabled.

On the new implementation, the navigation "tabs" for the section are present, but ineffective.

image

### LINKED ISSUE #1260: [Techstack-beta] Download: navigation clumsy on narrow
Plainly and simply, "vertical tabs list" doesn't work for narrow navigation. On any site.

For the original design, the decision was to make it so that on narrow those became "collapsible boxes".

The same section is active here in this contrived extra-long screenshot:

image

On real world device metrics, the issue becomes notable in that the context of where the user is can be off-screen, which makes reading sections one after the other harder to deal with.

image

### LINKED ISSUE #1261: [Techstack-beta] Download: EC2 section breaks width on narrow
The EC2 table pushes the content width over the edge on narrow dimensions.

image

### LINKED ISSUE #1276: [Techstack-beta] Homepage: flair images aspect ratio not respected
In this screenshot, the outline was added via `img { outline: 1px solid lime; }`. The only edit I made was added the width of the image after the fact, and aligning them in height. Both are from the new website.

You can observe that from 830px to 924px, the image has the same height, but ends-up being resized in width only.

wat

(On Firefox)

I have not spotted any other place where this is happening, but whatever the root cause is, check for other locations using the same pattern.

### LINKED ISSUE #1277: [Techstack-beta] Donate: acknowledgement logos scale is off (in different ways)
Both following screenshots are from the new implementation:

image
image

Observe the *Hetzner* and *antithesis* logos not scaling up like the others.

*Combined* to this issue here, the different scales picked for the logos here depending on the width of the window makes for an even more jumpy page than one would expect, with the logos going small, then medium, then small, then huge.

### LINKED ISSUE #1287: [Techstack-beta] Download: wrong link used in the fix
From #1283.

https://github.com/NixOS/nixos-homepage/commit/90f8e1f155b04221d49596e84fa69b418b610eb9#diff-75536eec7eac6d295419edc14756f523db498a2a56b106c10634452d2ceed33bL10

Link should be to `/nixos/manual/index.html#sec-booting-from-usb`.

### LINKED ISSUE #1292: [Techstack-beta] Site-wide: console and code blocks styles inconsistent
Both inconsistent within the new site, and inconsistent with the previous implementation.

* * *

Here's a screenshot from the Download page, where the new implementation is proper for a commands block:

image

* * *

Here's, comparatively, a screenshot from the *How Nix Works* page, showing an off-model commands block:

image

* * *

You may have noticed I'm using *commands block* here. This is because a stylistic decision was made to have code blocks on light, and terminal sessions on dark.

Source code blocks are always off-model in the new implementation.

image

* * *

Side-note: the only *story* in the *stories* section is off-model in the old site.

### LINKED ISSUE #1294: [Techstack-beta] Download: No way to deep-link
In the previous implementation, sections of the download page could be linked to directly.

 - https://nixos.org/download#nixos-virtualbox

thilobillerbeck:
- closes #1277 
- closes #1276 
- closes #1287 
- closes #1292
- closes #1259 
- closes #1260 
- closes #1261 
- closes #1294


thilobillerbeck:
TODO #1294

garbas (approved):

PR #1313, opened 2024-03-07
Fix inconsistent capitalizing

garbas:
Also some I moved some content from `src/content` since it makes it hard to navigate many files while it can be all done in one file.


thilobillerbeck (approved):
Looks good, but I think the nixcon section will get pretty long over time, maybe we should finde a proper solution for that in the future

PR #1326, opened 2024-03-08
[Textstack Beta] fix dividers, images, box sizes etc.

### LINKED ISSUE #1258: [Techstack-beta] Explore: images are larger than needed when narrow
Comparatively to the current implementation, this takes-up valuable vertical real-estate when scrolling for what is flair.

image

I know that this first comparison screenshot is somewhat contrived, but on a real world device metrics, it still has consequences:

image

### LINKED ISSUE #1288: [Techstack-beta] Design: Narrow menu button accessibility problems
Follow-up to the incorrect/incomplete fix for #1253.

* * *

In web development, generally speaking, `tabindex` shouldn't be used, and *if* it is needed, it generally means that it's used to cover another bug of deficiency in the design.

> [!WARNING]  
> Warning: You are recommended to only use `0` and `-1` as tabindex values. Avoid using `tabindex` values greater than `0` and CSS properties that can change the order of focusable HTML elements ([Ordering flex items](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Ordering_flex_items)). Doing so makes it difficult for people who rely on using keyboard for navigation or assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.
— https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

So, already, `tabindex=2` (and 1 for the logo) here should be considered unexpected as it makes keyboard navigation happen in an order different from visual/expected.

* * *

Using Firefox's Accessibility developer tool we see:

image

The warning links to:

 - https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Keyboard#avoid_using_tabindex_attribute_greater_than_zero

* * *

Let's move to Chromium with the Vimium extension. Vimium is a *really amazing* tool to find what I call **accessibility smells**. In itself it is not exactly (but also is) an accessibility tool, but the way it combs through the page to find element to navigate to is still representative of how different tools see the site.

To reduce noise, I use the Explore page, which is relatively simple.

Pressing `f` (follow link / activate item) one can see that on the new implementation the element is not seen as one that can be activated.

image

* * *

Still in Chromium, its Accessibility developer tool differs greatly from Firefox's. (And I kinda prefer its presentation when dealing with a site.)

Comparing the *Full-Page Accessibility Tree* (ignoring anything outside the banner, and focusing on the narrow menu navigation), we can see that there is no mention of the widget to toggle the menu in the new implementation.

image

So as far as tools using that model of the accessibility tree, the menu does not exist in a narrow view.

* * *

Quoting from [SCR35](https://www.w3.org/WAI/WCAG21/Techniques/client-side-script/SCR35.html):

>  In order to ensure that scripted actions can be invoked from the keyboard, they are associated with "natively actionable" HTML elements (links and buttons). The `onclick` event of these elements is device independent. While "`onclick`" sounds like it is tied to the mouse, the `onclick` event is actually mapped to the default action of a link or button. The default action occurs when the user clicks the element with a mouse, but it also occurs when the user focuses the element and hits enter or space, and when the element is triggered via the accessibility API.

In this instance, the fix to make the thing keyboard-accessible is:

 - Use no `tabindex` at all
 - Removing the keyboard handling (which is sometimes incomplete since it only handles *Enter*).
 - Replacing the generic semantics-free DOM element (the `<div>`) by, preferably, a `<button>`.

(`<a>` should be reserved for links that do page navigation, ensuring none of its subtle semantics are left, if it needs to be customized.)

Since `onClick` is used (via the `click` event listener) the *implicit and intrinsic* behaviour of the `<button>` makes it reachable via tab navigation, and can be activated via any user-agent method of activating the button, without having to consider every possible case.

### LINKED ISSUE #1295: [Techstack-beta] Asciinema examples: No way to deep-link
In the previous implementation, examples from the home page could be linked to directly.

 - https://nixos.org/#asciinema-demo-example_4

### LINKED ISSUE #1307: [Techstack-beta] Site-wide: Seams on slanted boxes
This is a ***really*** annoying issue, believe me, I know it.

I had to deal with it at the time, too.

The easiest (but not the only) way to reproduce is to use a mobile browser on a mobile device, here either Firefox on Android or Chrome on Android.

Go to a page with slanted boxes, and zoom in then out. It might not even be needed to zoom in/out, depending on the device metrics.

<details><summary>Large mobile screenshot</summary>

image

</details>
<details><summary>Large mobile screenshot</summary>

image

</details>
<details><summary>Large mobile screenshot</summary>

image

</details>

This can also happen with precision scrolling on desktop, but given the human-error input in precision scrolling, this is basically impossible to reproduce faithfully.

image

### LINKED ISSUE #1321: [Techstack-beta] Site-wide: Banners missing the hover state
https://github.com/NixOS/nixos-homepage/issues/1305 was closed with #1315 though incomplete.

> (Not pictured: how it is clickable... hard to screenshot... But the whole light blue part is clickable, **see the pointer cursor when you hover**.)

The whole banner being clickable shouldn't come as a surprise for mouse users. This is why the cursor should change to the pointer (hand pointer).

### LINKED ISSUE #1322: [Techstack-beta] Homepage banner not migrated to new component (to fix #1305)
https://github.com/NixOS/nixos-homepage/issues/1305 was closed with #1315 though incomplete.

I had assumed that with the improved tech stack, there wouldn't be issues where a design pattern is re-used at multiple locations, but implemented in a different ways, and that changing the behaviour of one wouldn't be a chore needing site-wide edits. And with that in mind, that *all* banners would be changed when fixing #1305.

I'll keep in mind that I do have to point out all locations on other issues.

image

### LINKED ISSUE #1324: [Techstack-beta] Blog: Filter checkbox visible
#1312 added a regression to the visuals of the filter.

The checkbox implementing the selection toggle is now clearly visible, as a smudge in the top/left of the element.

(Screenshots on Firefox, just in case somehow it isn't an issue in Chrome.)

image

image

thilobillerbeck:
- fixes #1324 
- fixes #1322 
- fixes #1321 
- fixes #1258 
- fixes #1295 
- fixes #1307 
- fixes #1288


samueldr:
## Slanted regression on homepage

With "medium"(?) sized viewports, there is no real padding under the search box.

<details><summary>Screenshots</summary>

image

image

</details>

Maybe not as much as there was beforehand, but definitely more than current. That would match the previous design.

## #1295

With 222144c7df4d0c2f59a63b803fc9bd4772bb6244 I can see it is possible to navigate to a deep-linked player, and it starts player as expected. Though I'm maybe missing the intended way for getting to the anchor (other than knowing you can inspect and put the anchor manually).

## #1258

Maybe I'm missing something, but comparing the last preview (https://65eb76d84ab57006ca156cfd--nixos-homepage.netlify.app/) and the current deployed site (https://main--nixos-homepage.netlify.app/) the explore page is identical in narrow mode.

samueldr:
## Slanted bits on homepage

(I can open a new issue if you prefer, as this is not strictly a regression from this PR here... exactly... but would be linked to how I would fix it normally...)

The slanted bit at the bottom of the search page had the "bar" bit decreasing in size more and more.

The original image is intended to be used scaled at 1:1, so that the bit is the same size on either side. Also the bit at the top should be visible in all viewport widths.

<details><summary>Old</summary>

image

</details>
<details><summary>main</summary>

image

</details>
<details><summary>This PR</summary>

image

</details>

* * *

<details><summary>Bits I'm talking about</summary>

image

</details>

samueldr (changes requested):
(See previous comments)

thilobillerbeck:
**Slanted bit**
yes, I kmow it still looks off. I'm already trying to bring it back to its original form while preserving the divider component. From what I remember the previous implementation used CSS backgrounds right? (Guess taken from in front of my phone :P) Maybe I can also add a second mode to the divider which inlines dividers and restore the way it was before.

**Explore page**
I resized them to 9/10 width, Its a subtile change, but when debugging it was visible.

thilobillerbeck:
everything should be adressed

thilobillerbeck:
> (I can open a new issue if you prefer, as this is not strictly a regression from this PR here... exactly... but would be linked to how I would fix it normally...)

I think this is better, since it directly addresses the changes in the PR instad of cycling through PRs and issues again and again. Thx btw. for revieweing so intensely, while moving this whole thing over, one for sure looses attention for detail :D

garbas (approved):

PR #1335, opened 2024-03-13
Fix some team usernames and make it clear they're for discourse

infinisil:
In another PR I'll also add a team with GitHub usernames instead.


thilobillerbeck (approved):
looks good, thx for suggesting this change and fix some of the names

infinisil:
I now opened #1337 which introduces github usernames

PR #1339, opened 2024-03-13
Download: fix tab navigation by removing unnecessary span

thilobillerbeck:
Removing the download tabs  `<span>` attribute since it causes problems and is unneeded anyway


infinisil (approved):
Can confirm that this fixes the issue, thanks!

PR #1347, opened 2024-03-14
community: Split talk section according to moderation status

samueldr:
I have left Stack Overflow in the main section because it is an already moderated property on the web.

Feel free to add a commit moving it if you believe this is not appropriate.

This also changes the style of the section slightly by not growing the buttons, as they become a bit too *ｗｉｄｅ*.

### Preview

Wide:

image

Narrow:

image

### DIFF
--- a/src/components/pages/community/CommunityLink.astro
+++ b/src/components/pages/community/CommunityLink.astro
@@ -21,6 +21,7 @@ const { href, iconName, name = "", description = "", type = "full" } = Astro.pro
     <a href={href} class="!no-underline !text-white">
       <div class:list={[
         "flex gap-2 text-xl font-bold",
+        (type !== "full" || description === "") && "h-full",
         type === "mobile_icon_desktop_name" && "flex-col md:flex-row items-center justify-center md:justify-start",
         ]}>
         <Icon
@@ -33,6 +34,6 @@ const { href, iconName, name = "", description = "", type = "full" } = Astro.pro
           type === "mobile_icon_desktop_name" && "hidden md:block",
           ]}>{name}</span>}
       </div>
-      {type === "full" && <div class="mt-2">{description}</div>}
+      {description !== "" && type === "full" && <div class="mt-2">{description}</div>}
     </a>
   </li>
--- a/src/components/pages/community/CommunityLink.astro
+++ b/src/components/pages/community/CommunityLink.astro
@@ -1,0 +1,38 @@
+---
+import { Icon } from "astro-icon";
+
+interface Props {
+    href: string;
+    iconName: string;
+    name?: string;
+    description?: string;
+    type?: "full" | "mobile_icon_desktop_name" | "icon";
+}
+
+const { href, iconName, name = "", description = "", type = "full" } = Astro.props;
+
+
+---
+
+<li class:list={[
+  "bg-nixdarkblue flex-1 rounded-xl p-4 hover:box-shadow-xl hover:bg-nixsemidarkblue transition-all cursor-pointer",
+  type === "icon" && "flex flex-col items-center justify-center",
+  ]}>
+    <a href={href} class="!no-underline !text-white">
+      <div class:list={[
+        "flex gap-2 text-xl font-bold",
+        type === "mobile_icon_desktop_name" && "flex-col md:flex-row items-center justify-center md:justify-start",
+        ]}>
+        <Icon
+          name={iconName}
+          alt={"logo of " + name}
+          class="w-8 h-8"
+        />
+        {type !== "icon" && <span class:list={[
+          "leading-8 ml-2",
+          type === "mobile_icon_desktop_name" && "hidden md:block",
+          ]}>{name}</span>}
+      </div>
+      {type === "full" && <div class="mt-2">{description}</div>}
+    </a>
+  </li>
--- a/src/pages/community.astro
+++ b/src/pages/community.astro
@@ -1,6 +1,5 @@
 ---
 import { getEntry, getCollection } from "astro:content";
-import { Icon } from "astro-icon";
 import { Image } from 'astro:assets';
 
 import Layout from "../layouts/Layout.astro";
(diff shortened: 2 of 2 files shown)


thilobillerbeck:
Good suggestions, allthough I'm not really happy with the fact that this now looks pretty uneven. I have the following suggestion, feel free to discuss:

- add names of the media to the the second row of the "where to talk section"
- style the outside spaces like the first row of the so neither the buttons nor the space around looks too empty

Or

- split "Where to talk" and "Outside Spaces" into two grid columns on wide screens and make the button grid feel the width then

I think the secound suggestion will use less vertical screen space and the horizontal one more effectively. Maybe giving that a try will make this section look more consistent and ordered looking.

samueldr:
I don't know tailwind and this website's use of tailwind, so I know I won't be able to achieve any result.

Though I'm fine with the design being reviewed, what's important here is the content.

Given that the design could take a while to settle on, it may be a good idea to split that as an additional task.

thilobillerbeck:
I absolutely agree, though I don't want this to go live without deteriating the design, bc. this changes the secionts layout. Would you be confortable with a review regarding your knowledge of tailwind (I mean, at least you seemd to figure out how to change flex behavior) or should I either PR your code or redirect this PR to a different branch and apply the changes quickly and leave them for re-review?

samueldr:
If you can redirect this to a PR with the edited implementation, please do.

Though note that you also can directly push to branches when this is checked. (Defaults to checked I think?)

image

And I am fine if you do here.

thilobillerbeck:
> Though note that you also can directly push to branches when this is checked. (Defaults to checked I think?)

Yeah this has been buggy for me multiple times. I hope this works now.
I'll apply my fixes then directly.

thilobillerbeck:
Since previews for forks do not work I post some screenshots here

localhost_4321_community(iPhone 12 Pro) (1)
localhost_4321_community(iPhone 12 Pro)
image

thilobillerbeck:
@samueldr could you review my changes and tell if this matches your original intention?

samueldr:
Pretty much okay, I was mainly focused on the content here.

The only caveat is I would have used `type="mobile_icon_desktop_name"` and not used a description. IMO the descriptions should be kept for the more important parts only.

samueldr:
Not sure how proefficient you are with a patches-based git workflow, but here's the patch for how I would have seen it:

```diff
From b4599962e7ec392de5facd6034dfc63e14ea42a1 Mon Sep 17 00:00:00 2001
From: Samuel Dionne-Riel <samuel@dionne-riel.com>
Date: Sat, 16 Mar 2024 20:57:50 -0400
Subject: [PATCH] community: Drop descriptions for unofficial spaces

Signed-off-by: Samuel Dionne-Riel <samuel@dionne-riel.com>
---
 src/pages/community.astro | 5 ++---
 1 file changed, 2 insertions(+), 3 deletions(-)

diff --git a/src/pages/community.astro b/src/pages/community.astro
index 5ac292c90..fe79c0ab0 100644
--- a/src/pages/community.astro
+++ b/src/pages/community.astro
@@ -103,13 +103,11 @@ const platformsOutside = [
     href: "https://discord.gg/RbvHtGa",
     name: "Discord",
     iconName: "simple-icons:discord",
-    description: "This community run Discord server is a place for real-time help, development chat, and off-topic community discussion. Please note that this Discord is not affiliated or moderated by the NixOS organization.",
   },
   {
     href: "https://www.reddit.com/r/NixOS/",
     name: "Reddit",
     iconName: "simple-icons:reddit",
-    description: "The NixOS subreddit is a place for news, discussion and questions about NixOS and Nix. Please note that this subreddit is not affiliated or moderated by the NixOS organization.",
   },
 ];
 
@@ -186,7 +184,8 @@ import nixosFoundationLogo from "../assets/image/nixos-foundation-logo.svg";
                 href={platform.href}
                 name={platform.name}
                 iconName={platform.iconName}
-                description={platform.description} />
+                description={platform.description}
+                type="mobile_icon_desktop_name" />
             ))
           }
         </ul>
-- 
2.43.2
```

* * *

Extra thing I was futzing with quickly. Used some defensive programming with with the community link widget to not become "weird" (taller than expected) when a blank or no description is given.

```diff
From e98ccd61aea1adb4c4e64ee470431bba947fbb17 Mon Sep 17 00:00:00 2001
From: Samuel Dionne-Riel <samuel@dionne-riel.com>
Date: Sat, 16 Mar 2024 20:58:19 -0400
Subject: [PATCH] CommunityLink: Handle description-less full items

This prevents a bogus empty element from being added.

Signed-off-by: Samuel Dionne-Riel <samuel@dionne-riel.com>
---
 src/components/pages/community/CommunityLink.astro | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/src/components/pages/community/CommunityLink.astro b/src/components/pages/community/CommunityLink.astro
index 8be9d50f8..b3cf8ed88 100644
--- a/src/components/pages/community/CommunityLink.astro
+++ b/src/components/pages/community/CommunityLink.astro
@@ -33,6 +33,6 @@ const { href, iconName, name = "", description = "", type = "full" } = Astro.pro
           type === "mobile_icon_desktop_name" && "hidden md:block",
           ]}>{name}</span>}
       </div>
-      {type === "full" && <div class="mt-2">{description}</div>}
+      {description !== "" && type === "full" && <div class="mt-2">{description}</div>}
     </a>
   </li>
\ No newline at end of file
-- 
2.43.2
```


I can send those to the branch if you agree with the change, and prefer it.

thilobillerbeck:
> Pretty much okay, I was mainly focused on the content here.
> 
> The only caveat is I would have used `type="mobile_icon_desktop_name"` and not used a description. IMO the descriptions should be kept for the more important parts only.

This was something I was concerned about too, though this makes the buttons look pretty wide agian when stretched like the official section. That's what drove me to that decision. But maybe I'm a bit too concerned at that point.

thilobillerbeck:
@samueldr just applied your suggestions :+1:  and fixed text and icons in the icon and text only buttons not being centered when shrinking to certain sizes. 

This is how it looks now
image

thilobillerbeck:
btw. I treid to shrink down the unofficial row to 2/3 of the full size so it fits the upper buttons, but imo. it looks uneven, but I'll leave it here as an inspiration 
image

samueldr:
I'll go with your judgement here, both options look fine, though I prefer the latter one.

Again, my main concern here is with the content :).

thilobillerbeck (approved):

thilobillerbeck:
self-approving this since design was discussed and reviewed through the conversation

PR #1353, opened 2024-03-16
teams: use matrix.to instead of app.element.io for links

katrinafyi:
This gives users the choice to use their own Matrix client instead of Element, and it also shows a nicer landing page instead of immediately going to the Element web app. This is already used for most of the Matrix links, there are just a few which use app.element.io.

Builds successfully and all affected links seem to work correctly. I've copied the links below if you would like to try them:
- https://matrix.to/#/#docs:nixos.org
- https://matrix.to/#/#cuda:nixos.org
- https://matrix.to/#/#foundation:nixos.org

Old link for comparison:
- https://app.element.io/#/room/#docs:nixos.org


thilobillerbeck (approved):
Also thanks for this one :)

PR #1360, opened 2024-03-20
Show shell prompt in macOS install instructions

wleightond:
This bugged me on the download page, so I know it'll bug others. Every set of instructions for the other platforms shows the `$` prompt (or [other prompt where relevant](https://github.com/NixOS/nixos-homepage/blob/main/src/content/download/04-nix-docker.mdx)), but the macOS one doesn't. 

This commit just adds a `$ ` for consistency. 

Seemed like a simple enough change that I just opened a PR; let me know if there needs to be a corresponding issue.


thilobillerbeck (approved):
looks good, thanks for contributing <3

PR #1368, opened 2024-03-22
Download page: open the OS specific tab on page load

### LINKED ISSUE #1115: Download page: open the OS specific tab on page load
**Introduce your stance**

MacOS users generally want to install Nix on their own system. I can't exclude the possibility that they want to install it on Ubuntu, but that's what the tab menu is for.

**Describe the issue**

When opening the download page on Safari, it shows the Linux instructions.

This comes across as inattentive, even if it's not a huge problem.

**Page links**

https://nixos.org/download

**Additional context**

Probably there's a more clever way to check for macOS rather than just Safari.

VedantWankhade:
Closes #1115 
Uses `navigator.userAgent` to find user's platform and sets the initial active tab accordingly.

### DIFF
--- a/src/components/pages/download/DownloadAccordion.astro
+++ b/src/components/pages/download/DownloadAccordion.astro
@@ -67,6 +67,23 @@ let downloadOptions = (await getCollection("download")).filter(
   const menuContainer = acc.querySelector(".download-accordion-menu");
 
   let activePanel = panels[0];
+
+  if (id === "download-nix-accordion") {
+    const userPlatform = navigator.userAgent.toLowerCase();
+    let os = "linux";
+    if (userPlatform.includes("macintosh")) {
+      os = "macos";
+    } else if(userPlatform.includes("windows")) {
+      os = "windows";
+    }
+    for (let i = 0; i < panels.length; i++) {
+      if (panels[i].id === `nix-install-${os}`) {
+        activePanel = panels[i];
+        break;
+      }
+    }
+  }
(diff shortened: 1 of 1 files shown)


src/components/pages/download/DownloadAccordion.astro
```diff
@@ -67,6 +67,23 @@ let downloadOptions = (await getCollection("download")).filter(
...
+      os = "macos";
+    } else if(userPlatform.includes("windows")) {
+      os = "windows";
+    }
+    for (let i = 0; i < panels.length; i++) {
+      if (panels[i].id === `nix-install-${os}`) {
+        activePanel = panels[i];
+        break;
+      }
+    }
```
thilobillerbeck, line 84:
```suggestion
    activePanel = acc.querySelector(`#nix-install-${os}`);
```

thilobillerbeck (changes requested):
Thanks for contributing, I have one small suggestion

src/components/pages/download/DownloadAccordion.astro
VedantWankhade: That's clever. Actually I tried the same approach the first time but I used `panels` to query on but obviously that didnt work and somehow I overlooked the `acc` variable.

VedantWankhade:
I have made changes according to the suggestion.

thilobillerbeck (approved):
Thanks for your contribution

PR #1374, opened 2024-03-30
Adapt visible nix to RFC-166

dasJ:
This should cause soon-to-be Nix users to see well-formatted code from the beginning :)


thilobillerbeck (approved):
looks good, thanks for contributing <3

PR #1378, opened 2024-04-01
Force GPU rendering to fix asciinema on Safari 16 and below

albertchae:
Use translateZ to force GPU rendering in Safari to fix asciinema rendering

Follow up to remaining issue on https://github.com/NixOS/nixos-homepage/issues/1323

It looks like newer versions of safari have fixed this https://github.com/asciinema/asciinema-player/issues/258
but if we want to support older versions, this [translateZ GPU hack](https://michaeluloth.com/css-translate-z/) seems
to work, at least for keeping the black terminal part within the parent containers border.

The sizing of the player font can still be incorrect, so one workaround is to make the `cover.scenario` 
asciicast slightly wider than necessary to accomodate


albertchae:
@thilobillerbeck @garbas do we have any data on the breakdown of browser + version of visitors to nixos.org? If the number of users on Safari < 16 is a small percentage, maybe we should just close this PR instead of merging a hack

thilobillerbeck:
I just did not have the time to review, will do tomorrow

albertchae:
> I just did not have the time to review, will do tomorrow

@thilobillerbeck oh sorry just to be clear I wasn't making a comment about code review time, just wondering if it's worth doing the `translateZ` hack if it turns out most (maybe > 99%?) Safari visitors to nixos.org are on the latest version. I usually give open source PRs at least 1-2 weeks before pinging for reviews so no worries there :D

src/styles/base.css
```diff
@@ -150,6 +150,10 @@
     position: relative !important; /* Override the absolute positioning of the play button (TODO: remove this hack) */
 }
 
+pre.ap-terminal {
+  transform: translateZ(0);
+}
```
thilobillerbeck, line 155: Moving this into the base layer and using tailwinds @apply directive would be awesome to keep everything consistent

thilobillerbeck (changes requested):
Looks alright, though I am not sure if we want to merge this or not, it seems like a pretty minor issue that would be rarely noticed if at all and is patched already on a browser level anyway

albertchae:
> Looks alright, though I am not sure if we want to merge this or not, it seems like a pretty minor issue that would be rarely noticed if at all and is patched already on a browser level anyway

I agree about possibly not merging it because it is already patched in the browser but it is most noticeable on ios mobile views, e.g. an iPhone 14 Pro





And the other thing to consider is that on iOS, all browsers (including Firefox and Chrome) use webkit under the hood (at least prior to [this EU ruling](https://www.theverge.com/2024/1/26/24052067/mozilla-apple-ios-browser-rules-firefox)) 

And AFAICT the only way to upgrade to Safari 17 + the associated webkit is to do a full iOS or macOS major version upgrade. You can't just upgrade the browser.

So those are the only reasons it might be worth doing this hack, if we have the user traffic from these browser versions to support it.

src/styles/base.css
albertchae: I'm not super familiar with tailwind but I'm not sure if this is what you meant https://github.com/NixOS/nixos-homepage/pull/1378/commits/a1ccacd5092361be4497272d6cf4e7d5e0f9744d. I had to add the asciinema-player js to tailwind's content scan otherwise the class would get stripped out

PR #1385, opened 2024-04-08
Fix nix pills build with new mdBook output

DanielSidhion:
Since https://github.com/NixOS/nix-pills/pull/233 got merged, the nix pills are now built with mdBook. In this process, we removed the `revCount` and `shortRev` attributes from the mdBook output.

This PR removes those attributes from here to ensure the nix pills can still be built with the new changes.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -85,11 +85,11 @@
     "nix-pills": {
       "flake": false,
       "locked": {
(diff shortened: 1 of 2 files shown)


thilobillerbeck (approved):
looks alright, thanks :)

PR #1387, opened 2024-04-11
Adjust cover video to work with latest nixpkgs

### LINKED ISSUE #1221: Asciinema demo contains not reproducible examples
**Introduce your stance**

There has been a broken build reported recently in the unofficial Discord, following the ["Demo"](https://nixos.org/#asciinema-demo-cover).

**Describe the issue**

The succesfull build of the expression defined in `redis-minimal.nix` in that demo can not be reproduced anymore with nixpkgs newer than `nixos-21.09`. The ressource itself is unclear about which `nixpkgs` was used for this, therefore I brute-forced all channels between 2020 and today.

**Page links**

* the demo: https://nixos.org/#asciinema-demo-cover
* the exact code in question: https://github.com/NixOS/nixos-homepage/blob/master/demos/cover.scenario#L80-L94

**Additional context**

garbas:
fixes #1221

### DIFF
--- a/public/demos/cover.scenario
+++ b/public/demos/cover.scenario
@@ -85,21 +85,23 @@ $ # The redis packaged in nixpkgs is not optimized for small size
 $ # Let us fix this!
 $ cat -n redis-minimal.nix
     1  {
-    2    pkgs ? import <nixpkgs> { },
-    3  }:
-    4  pkgs.redis.overrideAttrs (old: {
-    5    # no need for systemd support in our docker image
-    6    makeFlags = old.makeFlags ++ [ "USE_SYSTEMD=no" ];
-    7    # build static binary with musl
-    8    preBuild = ''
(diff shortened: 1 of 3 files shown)


garbas:
@Welding-Torch @NobbZ Can you verify that an example now works also for you?

thilobillerbeck (approved):

NobbZ:
I still do not like that there is no hint about what nixpkgs version this was tested against.

To be honest, relying on the nix path pointing to a compatible version of nixpkgs (it existing at all) and breaking if not, is very bad PR for the "reproducibility" we always try to hold up as the primary thing.

Of course, we do not want to introduce flakes here, unnecissarily increasing complexity of the demo, though at least mentioning that it was tested against a certain channel version, and that far older and far newer versions might require adjustements.

PR #1388, opened 2024-04-11
Fix daily CI update - blog/announcements.xml is not used anymore

garbas:


garbas:
I noticed Daily cron job was broken and this will fix it.

thilobillerbeck (approved):

PR #1390, opened 2024-04-12
Minor minimization of inputs redundancy in flake.nix

Rexcrazy804:
Rewrote the input declarations to fall within an overarching inputs structure scope. This manages to vividly distinguish the inputs from the succeeding cache configurations before the outputs declaration.


thilobillerbeck (approved):
looks good, thanks for your contribution <3

PR #1398, opened 2024-04-21
Explain what each code scenario snippet do

ibizaman:
Explain the important parts of the snippets instead of letting the user - possibly new to nix - try
to figure out what they do.

The update in the `flake.nix` file is because I didn't understand at first that the `npm install` was actually required to run `nix run dev`.


thilobillerbeck (approved):
Looks good, thanks for your contribution <3

Issue #1010, opened 2023-03-07
How nix{,os} works page has a grub screenshot


gilice:
Since Grub is no longer the default bootloader appiied with `nix-generate-config` (replaced by `systemd-boot`), the screenshot should be updated.

(Fun fact: the screenshot is almost 10 years old!)

Page links: https://nixos.org/guides/how-nix-works.html#rollbacks
Suggested edit: replace with a screenshot of systemd-boot

brnl:
![image](https://github.com/NixOS/nixos-homepage/assets/3243133/167766c3-1a6c-4008-b811-5327221fe771)

^ Does this work for you? I just cropped to 1024 px wide, maintaining aspect ratio. The original is also small in size.

PR #1401, opened 2024-04-22
fix: #1010 

sudanchapagain:
Extremely simple fix for #1010.
- Updated the text.
- Added image from the same issue.
- Removed previous image after greping for it in the code base and it isn't used anywhere else.

i also have a question why is image aligned that way with

```
article img:not(.img-noadjust) {
    @apply float-right inline ml-8 h-48 w-auto;
}
```

I don't think having a image be that small and kept in two column layout is necessary.


thilobillerbeck (approved):
Looks good, thanks for your contribution <3

Issue #1336, opened 2024-03-13
Make team member links distinguishable by an Icon


thilobillerbeck:
**Describe the glitch**
With multiple different paths a team member name link could lead you, it is good to show upfront where the link directs to. This would probably be best done by an icon. (one for discourse, github, etc...)

**Additional context**
This comes in context with #1335

thilobillerbeck:
Halting this until https://github.com/NixOS/nixos-homepage/issues/1340 is fixed

PR #1403, opened 2024-04-24
Make team member links distinguishable by an Icon

albertchae:
Closes https://github.com/NixOS/nixos-homepage/issues/1336

- Display github/discourse icons next to member links
- Need to set `inline-block` to display the icons inline without unnecessary line breaks
- Moved title higher and added line breaks in between name and links to better accommodate multiple links
- Also added some github examples for the marketing team so we can see what it looks like if there are both

### DIFF
--- a/src/content/teams/07_marketing.mdx
+++ b/src/content/teams/07_marketing.mdx
@@ -4,6 +4,7 @@ description: Maintains this website, responsible for social media outreach, coll
 members:
 - name: Rok Garbas
   discourse: garbas
+  github: garbas
   title: Leader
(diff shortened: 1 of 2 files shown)


albertchae:
Discussed in marketing meeting: @garbas suggested using only the icon instead of handle, and go back to inlining after name

thilobillerbeck:
> Discussed in marketing meeting: @garbas suggested using only the icon instead of handle, and go back to inlining after name

sounds like a good idea, maybe we can even think of including a popover if people want to have the name on the site directly

PR #1404, opened 2024-04-25
remove ryantm from moderation team

ryantm:
I don't have enough time to keep on top of my moderation tasks, so I'm stepping down from the team.


thilobillerbeck (approved):
Sad to see you leave, thanks for your efforts!

PR #1410, opened 2024-05-01
Fix Eelco's Nix team title

infinisil:
Having spoken with Nix team members, it looks like all are equally empowered, and that it's inaccurate to call Eelco the team lead.

See also https://discourse.nixos.org/t/nixos-foundation-board-giving-power-to-the-community/44552/7?u=infinisil

Ping @NixOS/nix-team, please review


roberth (approved):

roberth:
As @infinisil describes, this is an adjustment in our external communication, so not a demotion.

thilobillerbeck (approved):

PR #1427, opened 2024-05-17
Proofreading "How Nix Works"

jangdan:
This pull request adds a missing "be" to the body of ["How Nix Works"](https://github.com/NixOS/nixos-homepage/blob/022865693e96ea1c32335c18aecd93388ce9dd2a/src/pages/guides/how-nix-works.mdx).


djacu (approved):
Looks good to me.

PR #1436, opened 2024-05-21
50-nixos-isoimage: Reference plasma6 ISOs for 24.05

mweinelt:
The ISO name will change slightly due to the update from KDE Plasma version 5 to 6.

Blocked until we switch to 24.05 downloads on the homepage.

### DIFF
--- a/src/content/download/50-nixos-isoimage.mdx
+++ b/src/content/download/50-nixos-isoimage.mdx
@@ -41,8 +41,8 @@ The graphical installation ISO image **contains the graphical NixOS installer** 
   <span>
     <Button
       color="semidarkblue"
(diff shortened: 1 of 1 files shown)


thilobillerbeck:
stupid question, but why don't we name the image just plasma, would fit more into the naming scheme imo.

thilobillerbeck:
Stupid question, but why don't we name the image just plasma? would fit more into the naming scheme imo.

mweinelt:
That is a question for *tada* @K900.

K900:
Because we may want to have both, at least for a while.

mweinelt:
Superseded by #1453

PR #1437, opened 2024-05-22
[24.05] Add updated instructions about NixOS AMIs

arianvp:
The new AMIs aren't uploaded yet as the jobset didn't finish. But this adds updated instructions about  how to use them!


arianvp:
Related: https://github.com/NixOS/amis/pull/132

thilobillerbeck (approved):
I wanted to suggestion interpolating `{NIXOS_VERSION}` in the code blocks. While I cannot find a solution (since interpolating a code block is difficult for obvious reasons) this seems good for now. Thanks for proposing these changes. :)

tomberek (approved):
Thanks!

PR #1439, opened 2024-05-23
community: do not promote unmoderated spaces

nyabinary:
Cease promoting unofficial and umoderated (in regarding the values of our current CoC) spaces
Discussion on the following topic: https://github.com/NixOS/nixos-homepage/issues/1438

### DIFF
--- a/src/pages/community.astro
+++ b/src/pages/community.astro
@@ -98,19 +98,6 @@ const platformsOther = [
   },
 ];
 
-const platformsOutside = [
-  {
-    href: "https://discord.gg/RbvHtGa",
-    name: "Discord",
-    iconName: "simple-icons:discord",
-  },
-  {
-    href: "https://www.reddit.com/r/NixOS/",
-    name: "Reddit",
-    iconName: "simple-icons:reddit",
-  },
-];
-
 import nixosFoundationLogo from "../assets/image/nixos-foundation-logo.svg";
 ---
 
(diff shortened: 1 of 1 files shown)


mweinelt:
> unsanctioned

By whom? Who is the final authority in these questions, if I may ask?

nyabinary:
> > unsanctioned
> 
> By whom? Who is the final authority in these questions, if I may ask?

Presumably the board is until the new governance is formed (I did word that a bit off though, let me edit it :P)

thilobillerbeck:
I am a bit confused that this PR popped up without any prior discussion keeping in mind this is the root idea the issue was made for. I for one think it is indeed a good idea to also promote community channels but I definitely agree with @mweinelt s concerns on that topic. Though I do not agree with seizing the whole unofficial community section like this PR suggests, we cannot bear the load of moderating that many platforms with the current assembly of the moderation team and keeping out spaces with people eager to build up engaged Nix communities would be a bad sign to the community. I encourage the discussion under #1438 to come up with clear and concise but also fair communication about not officially moderated spaces.

thilobillerbeck:
> I believe you're thinking it from the wrong way around here, @thilobillerbeck.
> 
> We should remove the unmoderated spaces since they are currently unmoderated. We can work with those spaces to sync up the practices and moderation, and then add them back in.

That's a great idea, I would really like to see this becoming q reality.

> At least one of these locations right now is a place where the people removed from this community congregate to spread FUD.

Absolutely, but should we punish the other space for the one behaving badly? We could just remove the one spreading FUD

> Frankly, I'm surprised you *closed it* this eagerly. Are you taking a stance that the subreddit and the discord should be left on the site?

I just want discussions to take place in one place and not at two in parallel. This issue can be reopened anytime. I don't know where I am taking a stance here apart from expressing my opinion and directing the discussion back to the issue.

samueldr:
> Absolutely, but should we punish the other space for the one behaving badly? We could just remove the one spreading FUD

I have only had second-hand reports for the other location. The other location may be worse depending on how accurate and pervasive the behaviour from the second-hand reports are, and one's opinion.

So the reason I did not push for the other location is because *I do not personally know*. But if you want me to say more, I can go public with the second-hand reports.

> I just want discussions to take place in one place and not at two in parallel. This issue can be reopened anytime. I don't know where I am taking a stance here apart from expressing my opinion and directing the discussion back to the issue.

It wasn't clear why it was closed, sorry if this felt strongly worded. I'm really becoming tired of how the inability to act quickly in the NixOS org and community makes us look bad.

PR #1446, opened 2024-05-26
redirect all Nix manual URLs to nix.dev

fricklerhandwerk:
nix.dev serves stable URLs to the manual, and has redirects for the
`stable` and `unstable` mutable URLs. it also re-uses page-level
redirects that are supplied by the Nix manual itself, so we don't have
to take care of that here any more.

Made with @roberth at #ZurichZHF

### DIFF
--- a/public/netlify.toml
+++ b/public/netlify.toml
@@ -113,44 +113,8 @@
   force = true
 
 [[redirects]]
-  from = "/manual/nix"
-  to = "/manual/nix/stable"
-  status = 302
-  force = true
-
-[[redirects]]
-  from = "/manual/nix/:version/expressions/expression-language.html"
-  to = "/manual/nix/:version/language/index.html"
-  status = 302
-  force = true
-
-[[redirects]]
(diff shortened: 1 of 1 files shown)


thilobillerbeck (approved):

thilobillerbeck:
Great to see so much output from that event :)

infinisil:
This doesn't work! https://nixos.org/manual/nix/stable goes to https://nix.dev/manual/nix/2.18/stable/, which is a 404

roberth:
It's great to see versioned docs for Nix. :rocket: 
Also the new domain is a great fit, because Nix isn't NixOS.

This change inspired the new example in the docs for `remap` in
- https://github.com/NixOS/nixpkgs/pull/298665

Profpatsch:
> Also the new domain is a great fit, because Nix isn't NixOS.

and here we have the source of our issues

roberth:
@Profpatsch what issues are you referring to?

PR #1470, opened 2024-06-17
Fix redirect for hydra manual

srd424:
The redirect for the hydra manual link points to the job "manual" which seems to have been deleted in 2022, replaced by "manual.x86_64-linux" and "manual.aarch64-linux".

I don't know how much the manuals differ by architecture, if at all, but I'm guessing x86-64 is reasonable target here.


thilobillerbeck (approved):
looks good, thx for your contribution <3

PR #1472, opened 2024-06-21
Change Github Actions trigger type to pull_request_target to prevent tampering and allow workflow from forks

thilobillerbeck:
Changing the deploy trigger to `pull_request_target` to allow the CI to use our secrets for deployments regarding PRs from forks, making PR checks quicker and less error prone over time.

See here for details: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:
       - "main"
-  pull_request:
+  pull_request_target:
 
 permissions:
   contents: "write"


crertel:
Question: is this something that could be exploited to mine bitcoins or something, or will CI runs need manual approval?

thilobillerbeck:
First of all the runners the builds run on aren't ours anyway. Since we are an open source project, we fall under terms and pricing which allows us to use GitHub Actions for free.

Secondly as of my understanding, the current setup would allow for that, the change would run the CI script that is in the branch that should be merged to not in the branch merged from. This is also why GitHub Actions doesn't deploy PRs for staging that are not from our repo, since with a malicious script the secret would be extracted. It's in that regard more of a safeguard than before.

thilobillerbeck:
Approval is currently set to be manually required for any external contribution. Though this PR is for what happens after approval.

crertel:
Awesome, thank you for the explanation!: :)

crertel (approved):

tomberek (approved):

PR #1481, opened 2024-06-25
fix yet another typo in the redirects

fricklerhandwerk:
sorry everyone, this mess is currently not getting better as quickly as it should be.


thilobillerbeck (approved):
No worries, mistakes are human :D great you spotted it!

PR #1484, opened 2024-06-26
update wiki links

jopejoe1:
Replace unofficial wiki links with official wiki links


thilobillerbeck (approved):
Looks good, thx for contributing <3

PR #1489, opened 2024-06-28
Format/init

hsjobeki:
Adds formatting checks to the CI

with using "prettier" with "prettier-plugin-astro"

Locally developers can run:

`nix run .#format` to format the src directory`

Alternatively / Additionally:

Use any LSP that has `prettier` support.
 
or run:

`npm run format`
`npm run format:check`

Ci executes:

`nix build .\#checks.x86_64-linux.format`

Which fails if the formatting is not in the required format which will ensure the format stays consistent.


tomberek:
I've not looked at the workflow yet, but likely would want a .git-blame-ignore-revs

hsjobeki:
> I've not looked at the workflow yet, but likely would want a .git-blame-ignore-revs

- [x] Added .git-blame-ignore-revs

thilobillerbeck (changes requested):
Looks good :+1: could you update your branch to the current state of `main`? Then I'll merge it. :)

tomberek (approved):
Looks good overall. Let's just do a conflict resolution.

thilobillerbeck (approved):

PR #1567, opened 2024-10-21
Fix typo in AWS AMI search instructions

arianvp:
Filters combine with OR, not AND. So the instructions were wrong.


arianvp:
Also fixed here: https://github.com/NixOS/amis/pull/175

thilobillerbeck (approved):
Alright, thanks for fixing that :)

PR #1573, opened 2024-11-05
Add security page

crertel:
This PR is to start a discussion and hopefully get us an easy-to-find and short but useful security page.

image

I don't know what should go in these sections, so I need comment from @fricklerhandwerk @tomberek @NixOS/security  and I guess whoever else knows about these things.

My thinking is that if you hit the front page with a question about security that a) there should be something on the navbar right there about security (since, you know, we are trying to take it seriously) and b) clicking on that should take you to a page with clear, actionable instructions to either get you submitting your concern or answering your likely question in short order. This PR gets us in a position to do that.


crertel:
Prior art:

* [What Debian does, linked from their front page](https://www.debian.org/security/)
* [What Arch does, linked from their navbar](https://security.archlinux.org/)
* [What OpenSuse does, which took many levels of clicking through non-obvious contribution pages](https://en.opensuse.org/openSUSE:Security_team)
* [What Gentoo does, took a couple of clicks](https://www.gentoo.org/support/security/)
* [What RHEL does, took some clicking](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9#Security)
* Lix has an email in their footer.

crertel:
Do we not have a `security@nixos.org` or similar alias?

LeSuisse:
Hello,

Currently there is a `Security` link in the footer linking to https://nixos.org/community/teams/security/ providing information to privately report security issue. We get some notifications regularly so this process is already mostly working.

Do we need another page?

> Do we not have a `security@nixos.org` or similar alias?

A mailing list exists with our 3 emails, I'm however not entirely sure how it is managed.

mweinelt:
Managed by the infra team.

crertel:
@LeSuisse:

> Do we need another page?

So, this is why I've got the strawman up. I think the team page is good for telling about the team, but in terms of getting people the information or next steps they need to file a report it's kinda rough. The optics are that we in recent memory had at least one large dropped ball, and it'd probably help to make it painfully obvious that we take security seriously and have changed *something* as we improve our processes from that dropped ball.

It may well be that we just need to change that page and perhaps link to it more prominently; I'm just spitballing here and getting feedback.

@mweinelt:
 
> Managed by the infra team.

So, do we currently have that alias pointing at that mailing list? If so, does that mailing list include the current security team? If it doesn't, how hard would it be to add them to it--and if they don't want to be added, is there some other thing that would make more sense?

tomberek:
Conceptually seems reasonable. I'd expect the content should be driven by the Security team as well as deciding if they want this change. It would give them more visibility to communicate issues and for people to communicate with them.

crertel:
@mweinelt @NixOS/security just bumping this...do we have a `security@` alias and does it point at the current security folks? I checked the infra repo and didn't see an obvious definition, and it's unclear to me if the mailing list exists in the normal sense or in the "there is some mailing list under some alias somewhere" sense.

mweinelt:
We do maintain a security@ alias.

jfly:
@crertel, we currently use ImprovMX to manage the `security@` alias/list. We're planning to move to a self hosted alternative, which is tracked by https://github.com/NixOS/infra/issues/485.

> do we have a security@ alias and does it point at the current security folks?

I don't know have access to ImprovMX, so I can't confirm this. Perhaps @mweinelt or @Mic92 could?

tomberek:
@mweinelt Can you help fill in the content portions of the PR?

Mic92:
risicle and tgerbet and hexa receive security@

tomberek:
I mean this: https://github.com/NixOS/nixos-homepage/pull/1573/files#diff-fc516bcf6f7500940e2fe9a73c1a4ace1f7b74f009c6e6872a032a66d9061979R29

mweinelt:
I'd much rather the author did research prior art and came up with a proposal that we could iterate on. Otherwise it shifts too much of the work onto me.

crertel:
Somewhat underscoring this, I'll note that I sent out a test email to `security@` on 2024-11-20, and only heard back from (that I saw) Robert Scott (on the 21st, and I think that's technically within 24 hours if I look at the timestamps more closely--and thank you Robert!!!).

Anyways, there's definitely some room for improvement here.

thilobillerbeck:
@crertel Maybe you could explain your ideas in the marketing matrix channel and we can help shape them. :) We have open ears for fresh ideas and then we can iterate more quickly on those. This would save you time always building out prototypes as PRs. Or maybe draw them up first using your favourite graphic design software or a piece of paper. :)

I appreciate your effort, though I feel it'd be more efficient for you and us if you work with us directly. :)

Issue #830, opened 2022-04-20
Academic papers not visible on the website


fricklerhandwerk:
Quite some time ago the research papers on Nix and NixOS were removed from the website. 
- https://github.com/NixOS/nixos-homepage/commit/748c41d19350dbbf9f6f747d20df794bf2d43be1
- https://github.com/NixOS/nixos-homepage/commit/552a59b4031bf1a7fb896a0cad4faa35511eee39

@edolstra What was the reason was to do that, except academic papers being "scary"? Most essential things in them are still true, but most importantly they are the only material which explain motivation, rationale and architecture of Nix and NixOS.

This is only my personal anecdatum, but the theses were the central resource for me to understand what Nix is and why it is conceptually superior to anything else out there. Otherwise I would not have bothered to fight my way into making use of it. Here is [what the research section looked like](http://web.archive.org/web/20170705180623/http://nixos.org/docs/papers.html) when I started out.

On top of that it's part of the project's and community's history. I agree with https://github.com/NixOS/nixos-homepage/issues/446#issuecomment-697486801 that it should be kept accessible.

Maybe we can find a nice spot for the papers somewhere?

sambacha:
> Quite some time ago the research papers on Nix and NixOS were removed from the website.
> 
> * [748c41d](https://github.com/NixOS/nixos-homepage/commit/748c41d19350dbbf9f6f747d20df794bf2d43be1)
> * [552a59b](https://github.com/NixOS/nixos-homepage/commit/552a59b4031bf1a7fb896a0cad4faa35511eee39)
> 
> @edolstra What was the reason was to do that, except academic papers being "scary"? Most essential things in them are still true, but most importantly they are the only material which explain motivation, rationale and architecture of Nix and NixOS.
> 
> This is only my personal anecdatum, but the theses were the central resource for me to understand what Nix is and why it is conceptually superior to anything else out there. Otherwise I would not have bothered to fight my way into making use of it. Here is [what the research section looked like](http://web.archive.org/web/20170705180623/http://nixos.org/docs/papers.html) when I started out.
> 
> On top of that it's part of the project's and community's history. I agree with [#446 (comment)](https://github.com/NixOS/nixos-homepage/issues/446#issuecomment-697486801) that it should be kept accessible.
> 
> Maybe we can find a nice spot for the papers somewhere?

Preach Brother [fricklerhandwerk](https://github.com/fricklerhandwerk), this is kind of important

[...]

PR #1580, opened 2024-11-14
re-add papers/research section

mschwaig:
I re-added this section, based on https://web.archive.org/web/20170518051949/http://nixos.org/docs/papers.html and recent works I am aware of.

~This is just a draft, I still need to check if I got all of the right information.~

Since the original page was at http://nixos.org/docs/papers.html and this is at http://nixos.org/research, it would also be nice to set up a redirect, or change the URL.

I'm happy to review PRs related to this list.
To make that process easier maybe we should
* add a short notice with a link before the list which sends people here
* introduce a label for those issues

Feel free to help me work on the text!

I think @fricklerhandwerk wanted to review this.


fricklerhandwerk:
Nice!

Closes https://github.com/NixOS/nixos-homepage/issues/830

Off the top of my mind, a seminal paper that deeply concerns Nix is [Build systems a la carte](https://www.microsoft.com/en-us/research/uploads/prod/2018/03/build-systems.pdf).

I have a few small nits on the wording, and will add a few comments. Other than that this is really really good, and we can add more papers as we go.

src/pages/research.astro
```diff
@@ -0,0 +1,1170 @@
...
+---
+
+<Layout title="Research and Scientific Publications">
+  <PageHeader text="Research and Scientific Publications" />
+  <Container class="grid px-12 pt-16">
+    <p class="text-xl font-bold text-gray-500">
+      Nix is a research project who's time to solve today's software
+      distribution issues has come.
+    </p>
+  </Container>
```
fricklerhandwerk, line 824:
```suggestion
```

Semantically this would have been an `<aside>` btw. In any case I wouldn't want any of this to sound like advertisement but instead do the topic of the page justice and present things as a matter of fact.

src/pages/research.astro
```diff
@@ -0,0 +1,1170 @@
...
+      <p>
+        Nix got started as a research project by Eelco Dolstra and his
+        collaborators, which over time attracted a dedicated community based on
+        its revolutionary ideas.
+      </p>
+      <p>
+        As a part of this growing community, a new crop of researchers has
+        discovered Nix as a fitting tool to tackle todays issues of scientific
+        and binary reproducibility, software distribution and supply chain
+        security.
```
fricklerhandwerk, line 839:
```suggestion
        Nix started as a research project by Eelco Dolstra and his collaborators at Utrecht University around 2003.
        Since then, scientists from multiple institutions have published their work on repeatable computation and reliable, secure software distribution.
      </p>
      <p>
        This collection traces the continued exploration of theoretical foundations and practical applications of the ideas underlying Nix and its ecosystem.
```

I'd want to go away from using flashy adjectives, and find a phrasing that doesn't go out of date quickly.

src/pages/research.astro
```diff
@@ -0,0 +1,1170 @@
...
+                              </svg>
+                              <span class="ml-2 text-sm font-bold text-gray-400 move-button">
+                                DOI
+                              </span>
+                            </a>
+                          )}
+                        </div>
+                        <div class="ml-2">
+                          {paper.abstract && (
+                            <button
```
fricklerhandwerk, line 966: There's no reason whatsoever for this to be javascript-only. It could easily be a `<details>` element with the appropriate styling. Not a blocker, but I'd strongly prever to (eventually) not have to use any scripting

src/pages/research.astro
```diff
@@ -0,0 +1,1170 @@
...
+                                class="h-5 w-5"
+                                viewBox="0 0 20 20"
+                                fill="currentColor"
+                              >
+                                <path d="M8 3a1 1 0 011-1h2a1 1 0 110 2H9a1 1 0 01-1-1z" />
+                                <path d="M6 3a2 2 0 00-2 2v11a2 2 0 002 2h8a2 2 0 002-2V5a2 2 0 00-2-2 3 3 0 01-3 3H9a3 3 0 01-3-3z" />
+                              </svg>
+                              <span class="ml-2 text-sm font-bold text-gray-400 move-button">
+                                CITE
+                              </span>
```
fricklerhandwerk, line 909: Without scripting this could lead to a BibTex raw file (not a blocker)

fricklerhandwerk (approved):
This is awesome, absolutely great stuff, and it seems to keep coming. I'd merge this with the prose suggestions and would help get fixups in if you're still interested in polishing the UX.

thilobillerbeck (changes requested):
One nitpick I have is that the buttons on the publications (Preprint, Download, etc.) have a contrast of 2.53:1 when comparing with the background, which is not that great. It'd be great if you could change this to a color that at least suffices AA contrast. :)

src/pages/research.astro
mschwaig: It does not work with `<details>` or at least I cannot make it work without significantly changing the overall layout and compromising the design. I made it work using a checkbox now similar to this: https://stackoverflow.com/questions/41220717/collapse-without-javascript/41220875#41220875.

src/pages/research.astro
```diff
@@ -0,0 +1,1131 @@
...
+    publicationInfo: {
+      type: "conference",
+      conference:
+        "11th International Workshop on Software Configuration Management (SCM-11)",
+      location: "Portland, Oregon, USA",
+      pages: "102–117",
+    },
+    preprintOrArchiveUrl: "https://nixos.org/~eelco/pubs/iscsd-scm11-final.pdf",
+  },
+];
```
thilobillerbeck, line 796: Maybe this could be split up into a separate content collection, see https://docs.astro.build/en/guides/content-collections/

src/pages/research.astro
```diff
@@ -0,0 +1,1131 @@
...
+                  </article>
+                ))}
+              </div>
+            </section>
+          ))
+        }
+      </div>
+    </Container>
+  </div>
+</Layout>
```
thilobillerbeck, line 1032: Some parts of these templates look really repetitive to me, maybe those could be split up into reusable components. This is currently really hard to read tbh. though it does its job, so good job htere :)

src/pages/research.astro
```diff
@@ -0,0 +1,1131 @@
...
+
+<style>
+  .btn-action {
+    @apply p-2 text-gray-400 hover:text-gray-600 rounded-full hover:bg-gray-100 transition;
```
thilobillerbeck, line 1036:
```suggestion
    @apply p-2 text-gray-500 hover:text-gray-600 rounded-full hover:bg-gray-100 transition;
```

src/pages/research.astro
```diff
@@ -0,0 +1,1131 @@
...
+  .abstract-toggle-checkbox:checked ~ .abstract-content {
+    max-height: 2000px; /* Adjust as needed */
+    padding-top: 30px;
+    opacity: 1;
+  }
+  /* Rotate the SVG Icon When Checked */
+  .abstract-toggle-checkbox:checked + div .abstract-toggle svg {
+    transform: rotate(-180deg);
+  }
+</style>
```
thilobillerbeck, line 1124: If you use components, you could apply more tailwind classes, I'd prefer people sticking to tailwind instead of mixing up stuff like this. It's okay to use CSS if it is the last resort, but most of this can be done like the rest of this project does it.

src/content/menus/footer.yaml
```diff
@@ -31,6 +31,8 @@ sections:
         link: /blog/
       - name: Newsletter
         link: https://weekly.nixos.org/
+      - name: Research and scientific publications
```
thilobillerbeck, line 34: This link now looks pretty long in the footer, maybe we should come up with a shorter title

thilobillerbeck (changes requested):
Looks good, though I found one small thing where the colors did not apply. I added some other suggestions.

src/content/menus/footer.yaml
mschwaig: Maybe. I think 'Research publications' would be good, 'Scientific publications' would be OK as well.
I do like the current title the best, but if it does not look good I am fine with shortening it.

I do think that title of the page itself should stay the same though, and I am not sure we should change it in the footer anyways or not.

src/pages/research.astro
mschwaig: I do not want to spend a lot of time on putting every entry in its own file right now, because that interferes with my own personal workflow for adding and modifying entries and I would also have to add an index to keep them in the intended order.

Astro on the other hand does not seem to like collections where all the content is in one file though (that's what I would like to do at least for now), so that also does not work.

Basically I'd would prefer to keep this as is now, or just move all entries into another file, and then see later on if making it a collection with each entry in its own file makes sense.

src/pages/research.astro
mschwaig: Noted. I rarely do UI stuff at all, so I was not aware this is bad style. Maybe I should have looked at how other pages are written.

src/pages/research.astro
mschwaig: Thanks. 3c8f22f13578fb6d405576a918b54b57c7259b46 should fix this. To me it looked what I removed there (my earlier change) did not really have a visible impact.

src/pages/research.astro
mschwaig: In principle I agree with this suggestion.

mschwaig:
@thilobillerbeck thanks for the review.

Unless you insist, I would like to get this merged as is, without addressing your other suggestions first.

I do understand it would be better to write reusable components and use CSS as a last resort and probably using a content collection is a good idea as well, but to be frank I want to limit how much time I am spending on the details of this effort in the short term.

fricklerhandwerk (approved):

thilobillerbeck:
> @thilobillerbeck thanks for the review.
> 
> Unless you insist, I would like to get this merged as is, without addressing your other suggestions first.
> 
> I do understand it would be better to write reusable components and use CSS as a last resort and probably using a content collection is a good idea as well, but to be frank I want to limit how much time I am spending on the details of this effort in the short term.

I get your point, though it'd be great if those points could be adressed in the future, since otherwise this page will at some point be pretty hard to maintain. But I'll merge this for now. Sorry for bothering, I just want to keep this project as maintainable as it is. :D

thilobillerbeck (approved):

PR #1582, opened 2024-11-16
Refactoring Suggestion

SvMak:
Attention! This is just my vision of what this project could look like. These are not final improvements, but a stage along the way.

The goal is to improve the code base to further resolve existing issues and trying to make the UI more consistent.

- Used normal font weight, as using thin font is not a good practice, it is hard to read;
- Moved page sections into separate components;
- Minor UI/UX improvements;
- Added prettier plugins to improve code formatting.


thilobillerbeck:
This definitely contains some good changes overall, I hope I can spare some time soon to review this

thilobillerbeck:
It'd be great btw. to split those changes into smaller PRs since this PR now breaks with every subtle change made due to it touching a lot of moving parts

crertel (changes requested):
I think that, as other folks pointed out, smaller refactors here would help--and let's break out the easy-to-go-with stuff first.

A couple of relatively low-hanging fruit that would be good to extract out first:

* The additional prettier plugins
* Maybe the `tsconfig.json` tweaks
*  The `Link` util component

I'm not really a fan of the refactors to chop up the pages into smaller components; for what it's worth, when I started working on this not long ago I found it to be incredibly helpful to be able to copy over a single file and tweak it--I don't believe there's much extra utility, for example, in breaking up a single document like the values page or what have you into lots of single-use components.

SvMak:
> It'd be great btw. to split those changes into smaller PRs since this PR now breaks with every subtle change made due to it touching a lot of moving parts

I would be happy to get some feedback to understand your vision for the project as a whole.

We can use this PR as a set of changes, you can choose the ones you would like to see in the main branch and I'll make new PR/PR's.

SvMak:
Of course, the component approach is not necessary to use, it is useful for simplifying work with large pages and implementing/replacing sections on the page.

I can make a separate PR with changes without splitting the pages into components or even more granular changes. Let me know what would be preferable.

crertel:
My biggest argument against the component approach, based on the various React and LiveView apps I've maintained, is that if the components are not reusable (contrast the proposed subdivision of the values page for example with the proposed link utility) then it only adds another level of indirection. At some point, especially for this, each page really is mostly just a document, and breaking it into smaller pieces obscures that fact.

Being able to trivially add a new page by copying an existing one and editing it to-taste in basically one spot I found tremendously helpful as a new contributor.

SvMak:
> My biggest argument against the component approach, based on the various React and LiveView apps I've maintained, is that if the components are not reusable (contrast the proposed subdivision of the values page for example with the proposed link utility) then it only adds another level of indirection. At some point, especially for this, each page really is mostly just a document, and breaking it into smaller pieces obscures that fact.
> 
> Being able to trivially add a new page by copying an existing one and editing it to-taste in basically one spot I found tremendously helpful as a new contributor.

Sorry, but I'm not trying to 'sell' you a component approach to pages))

I'm here to improve the code base and solve issues.

I have developed and supported dozens of projects written in various languages ​​and frameworks, and for me any approach is a compromise.

For this project at this stage, I completely agree with you, so additional argumentation is unnecessary.

I hope to be able to make several PR based on the approved improvements in the near future.

crertel:
> For this project at this stage, I completely agree with you, so additional argumentation is unnecessary.

Sorry, sorry--was just trying to explain my position for posterity! No harm intended.

Thank you very much for your contributions!

SvMak:
> > For this project at this stage, I completely agree with you, so additional argumentation is unnecessary.
> 
> Sorry, sorry--was just trying to explain my position for posterity! No harm intended.
> 
> Thank you very much for your contributions!

I respect your position.

PR #1586, opened 2024-11-18
add prettier to dev shell

mschwaig:


flake.nix
```diff
@@ -222,6 +223,9 @@ rec {
                 npm run build
                 cd dist
                 netlify dev
+
+            To re-format the source code:
+                prettier --write ./src
```
crertel, line 228:
```suggestion
                npm run format
```

crertel (approved):
Looks good, see comment. We may already have some of this for you.

mschwaig:
Actually if npm manages what version of `prettier` is used, then adding the one from nixpkgs to the command line might not be a good idea, so I removed that from this PR again.

I got the idea from the log of the failing format check here: https://github.com/NixOS/nixos-homepage/actions/runs/11843965454/job/33006101651

crertel (approved):

thilobillerbeck (approved):

Issue #1035, opened 2023-04-11
The events funding page is very hard to discover


thufschmitt:
**Describe the issue**

The “events funding” page is very hard to access from the front page (go to https://nixos.org/community/, go nearly at the bottom an click the “[Event Funding](https://nixos.org/community/event-funding.html)” link).

We should find a way to make it more easily discoverable.

**Page links**

https://nixos.org/community/event-funding.html

cc @domenkozar

domenkozar:
We could make https://20th.nixos.org/ be a list of meetups and part of nixos.org and expose it on the front page. Thoughts?

fricklerhandwerk:
There already is a meetups page, and it could be separated out and simply kept up to date.

Event funding can be linked at the top or bottom, at least that would be related.

garbas:
I'll add this topic to the next marketing meeting, especially now that I hope website efforts are moving forward.

But I'd like to understand better and I'd love you to answer few more questions:
- Who is the audience that you expect to click on? I know meetup organizers, but who are they? How do they find us?
- Why do you think this is more (or less) important than other things on the front page? on the community page?
- What would be your the desired outcome? I'm looking for a user story ...

I think it would definitely help if we would have a page where all the events (especially next events) would be listed.

PR #1587, opened 2024-11-20
Break out teams page.

### LINKED ISSUE #1511: Make a separate "Teams" page
**Problem**
The [community page](https://nixos.org/community/) is currently about too many things:
- communication channels
- NixCon
- meetups
- RFC process
- teams
- foundation

**Proposal**

Split out the Teams part.

Ideally also move the foundation stuff on a separate page linked from the footer.
The RFC process diagram could go into the steering committee page.

I'd just do it myself, but the website has so much design stuff that one can't just move a bunch of text to a different place, and I can't make the time to figure out what to do until it looks correct. (Side note: It would be great if the contribution process would boil down to moving text around.)

**Additional context**
There were two notable discussions on the role of teams in 2023:
- https://discourse.nixos.org/t/what-do-you-think-of-teamification/34849
- https://discourse.nixos.org/t/nixcon-governance-workshop/32705/7

My takeaway is that we need to help people doing the technical work to get more visibility and access to those interested in helping out, and those interested in helping out a way to find people doing the technical work.

Yesterday I discussed with @imincik that we should emphasize the existence of active groups working on particular things, and  and that joining one is a fast track to making change happen without getting frustrated. A separate navigation item would be a first step in that direction.

crertel:
Closes #1511.

Quick stab at this, lemme know what you think.

Removed the teams page, which meant a color change for the foundation section:

image

And then the teams page, nice and clean:

image

Larger view:

image

Mobile:

image

### DIFF
--- a/src/content/menus/header.yaml
+++ b/src/content/menus/header.yaml
@@ -9,6 +9,8 @@ items:
     link: /values
   - name: Community
     link: /community
+  - name: Teams
+    link: /teams
   - name: Blog
     link: /blog
   - name: Donate
--- a/src/pages/community.astro
+++ b/src/pages/community.astro
@@ -411,111 +411,88 @@ import nixosFoundationLogo from "../assets/image/nixos-foundation-logo.svg";
     <Divider mirrorY />
   </div>
   <div class="bg-nix-blue py-8">
-    <Container>
-      <h2 class="text-4xl font-bold text-white mt-8" id="governance-teams">
-        Teams
-      </h2>
-      <ul class="flex flex-wrap gap-x-6 gap-y-16 mt-8 justify-stretch">
-        {
-          teams.map((team) => (
-            <li class="text-white [&>a]:inline-block basis-72 flex-grow flex flex-col items-center text-center md:items-start md:text-left gap-2">
-              <img
-                src={`/images/teams/${team.slug.split("_")[1]}.svg`}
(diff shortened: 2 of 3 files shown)


thilobillerbeck:
What is the practical outcome you want to achieve with this? I mean, yeah, it looks interesting, but now we have another page and the main nav is even larger than before, which now starts to work against the idea of a quick and concise way of navigation.

crertel:
@fricklerhandwerk filed an issue, here's a PR that resolves that issue. If we don't want to do it, that's cool.

The teams stuff is hidden pretty far down in the community section, and the navbar is a little large but also we could probably consider removing some things if that's a concern. Anyways, this is just to get a conversation started.

fricklerhandwerk:
@thilobillerbeck arguably we could move "Values" into the community page, that would fit thematically. Eventually we should also consolidate everything about events (related: https://github.com/NixOS/nixos-homepage/issues/1035#issuecomment-1504683858). There's already https://nix.ug/ which does a very good job at keeping information up to date, and could possibly be integrated more into official resources (wdyt @MatthewCroughan?) or at the very least linked to.

fricklerhandwerk:
Please mind, as is, this will break anchor links to the teams section. We have quite a few of them around, I remember running into the problem of that particular one breaking during the rework.

crertel:
Good catch, I'll do a pass on those when I pick this back up.

PR #1592, opened 2024-11-27
Use a badge that corresponds to an actual workflow

eiriksm:


thilobillerbeck (approved):
looks alright, thanks for contributing

PR #1597, opened 2024-11-30
Fix asciinema-player

SvMak:
Click title for asciinema-player doesn't run player, because logic is described outside the scope of the variable "player".


thilobillerbeck (approved):
looks alright, thanks for contributing

PR #1599, opened 2024-12-01
chore: convert flake to flake parts and add git-hooks

thilobillerbeck:
This PR adds

- flake-parts to the flake
- a git hook that checks formatting on push
- npm package installation if node_modules does not exist to prevent errors when pushing

The reason I set on-push for the hooks is, that people could get quite annoyed when having to check for proper formatting on every commit. Since we doe merge PRs anyway, we can require proper formatting on push and ease local development for contributors.

I'd like to hear feedback from other contributors here on what to add or to change before I merge this into main.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -232,7 +232,10 @@ rec {
           packages.demos = demos;
 
           pre-commit.settings.hooks = {
-            nixfmt-rfc-style.enable = true;
+            nixfmt-rfc-style = {
+              enable = true;
+              files = "\\.nix$";
+            };
             prettier-check = {
               enable = true;
               name = "check-formatting";
--- a/flake.nix
+++ b/flake.nix
@@ -238,6 +238,7 @@ rec {
               name = "check-formatting";
               entry = "${nodejs_current}/bin/npm run format:check";
               stages = [ "pre-push" ];
+              pass_filenames = false;
             };
           };
 
--- a/flake.nix
+++ b/flake.nix
@@ -69,6 +69,8 @@ rec {
         nix_stable = released-nix-stable.packages."${system}".nix;
         nix_unstable = released-nix-unstable.packages."${system}".nix;
 
+        nodejs_current = pkgs.nodejs_20;
+
         nixPills = import nix-pills {
           inherit pkgs;
         };
@@ -202,16 +204,17 @@ rec {
         pre-commit.settings.hooks.prettier-check = {
           enable = true;
           name = "check-formatting";
-          src = "src/**/*";
-          entry = "${pkgs.nodejs_20}/bin/npm run format:check";
+          entry = "${nodejs_current}/bin/npm run format:check";
           stages = [ "pre-push" ];
         };
 
         devShells.default = pkgs.mkShell {
           name = "nixos-homepage";
 
+          inputsFrom = [ config.pre-commit.devShell ];
+
           packages = with pkgs; [
-            nodejs_20
+            nodejs_current
             netlify-cli
           ];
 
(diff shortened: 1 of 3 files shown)


djacu (approved):
I'm down with all these changes. On-push hooks sounds like a nice QoL improvement!

flake.nix
```diff
@@ -27,17 +28,29 @@ rec {
...
     , released-nixpkgs-unstable
     , released-nixpkgs-stable
     , released-nix-unstable
     , released-nix-stable
     , nix-pills
     }:
-    flake-utils.lib.eachDefaultSystem (system:
-      let
+  flake-parts.lib.mkFlake { inherit inputs; } {
+    flake = {};
```
drupol, line 43: Let's remove the parts we don't need

flake.nix
```diff
@@ -27,17 +28,29 @@ rec {
...
+    imports = [
+      inputs.git-hooks-nix.flakeModule
+    ];
+    systems = [
```
drupol, line 47: This can be shortened with: https://github.com/loophp/nix-shell/blob/4a8220125e9537dfa316b49e2b9cf77e977c86af/flake.nix#L20

drupol (commented):
I just added minor and optional comments.

What I would do here is to format the file with nixfmt-rfc-style.

crertel (approved):
LGTM.

flake.nix
```diff
@@ -27,206 +28,267 @@ rec {
...
-            pname = "asciinema-scenario";
-            version = "0.2.0";
-            src = final.fetchCrate {
-              inherit pname version;
-              sha256 = "sha256-qMGi+myppWBapM7TkPeXC2g/M1FA1YGwESNrx8LVXkw=";
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      imports = [
+        inputs.git-hooks-nix.flakeModule
+      ];
+      systems = import systems;
```
drupol, line 47: You forgot to import systems in the inputs attribute (https://github.com/loophp/nix-shell/blob/4a8220125e9537dfa316b49e2b9cf77e977c86af/flake.nix#L9).

flake.nix
```diff
@@ -27,206 +28,267 @@ rec {
...
+      perSystem =
+        { config, system, ... }:
+        let
+          overlay = final: prev: {
+            asciinema-scenario = final.rustPlatform.buildRustPackage rec {
```
drupol, line 52: Is available in https://search.nixos.org/packages?channel=24.11&show=asciinema-scenario&from=0&size=50&sort=relevance&type=packages&query=asciinema

flake.nix
```diff
@@ -27,206 +28,267 @@ rec {
...
+              };
+              cargoSha256 = "0z4iwjm38xfgipl1pcrkl8277p627pls565k7cclrhxfcx3f513p";
             };
-            cargoSha256 = "0z4iwjm38xfgipl1pcrkl8277p627pls565k7cclrhxfcx3f513p";
           };
-        };
 
-        pkgs = import nixpkgs { inherit system; overlays = [ overlay ]; };
-        inherit (pkgs.lib) getVersion;
+          pkgs = import nixpkgs {
```
drupol, line 63: pkgs is already provided by the flake.parts framework, use it as such: https://flake.parts/overlays#consuming-an-overlay

drupol (commented):
I might submit a PR to fix this and some other minor improvements.

thilobillerbeck:
> I might submit a PR to fix this and some other minor improvements.

That'd be great, I thinks the flake can see some improvements anyway. I'm pretty new to flake-parts though, so it's good to have some people around who have used it for longer :D

PR #1602, opened 2024-12-02
Refactor research page

mschwaig:
Follow up to #1580 , to address the review comments by @thilobillerbeck.

As of now this should be functionally complete and make the code more maintainable, but I'm not in a hurry to double-check and get it merged due to other projects.

A few other things from the original issue are also still open, but we can decide if we want to address those preemptively or not:

* Since the original page was at http://nixos.org/docs/papers.html and this is at http://nixos.org/research, it would also be nice to set up a redirect, ~or change the URL~.
* Make it easier for people to add papers, by linking to an issue template in this repo with an appropriate label

Also, ed6b212f3646ed2e94831da72371531c66c79a83 introduced [src/content/research/papers.yaml ](https://github.com/NixOS/nixos-homepage/blob/ed6b212f3646ed2e94831da72371531c66c79a83/src/content/research/papers.yaml)for some reason.
Depending on how that happened it may or may not be easy to switch using content collections based on that file.
I tried that myself, when I was implementing the original issue, and it looked like Astro would not be happy about having all of the collection entries in one file.


src/components/papers/PaperActions.astro
```diff
@@ -0,0 +1,135 @@
...
+          {/* source https://heroicons.com v1.0.6, MIT licensed */}
+          <path d="M8 3a1 1 0 011-1h2a1 1 0 110 2H9a1 1 0 01-1-1z" />
+          <path d="M6 3a2 2 0 00-2 2v11a2 2 0 002 2h8a2 2 0 002-2V5a2 2 0 00-2-2 3 3 0 01-3 3H9a3 3 0 01-3-3z" />
+        </svg>
+      </button>
+    )
+  }
+  {
+    preprintUrl && (
+      <a
```
Maximilian-Maag, line 38: This link contains an image and would be hard to read with a screenreader. It should contain an aria-label.

src/components/papers/PaperEntry.astro
```diff
@@ -0,0 +1,98 @@
...
+    case 'thesis':
+      return `${pubInfo.thesisType} thesis, ${pubInfo.institution}, ${pubInfo.location}`;
+  }
+}
+---
+
+<article
+  class="shadow-sm group relative rounded-lg bg-white p-6 transition-all duration-200 hover:shadow-md"
+>
+  <input type="checkbox" id={getToggleAbstractId(paper)} class="peer hidden" />
```
Maximilian-Maag, line 31: The checkbox has no label, for screenreader users it could be hard to understand for what perpose this checkox must be used.

PR #1608, opened 2024-12-14
Small buttons update

SvMak:
- use cva(class-variance-authority) for variants;
- ~~made all buttons full rounded.~~


src/components/ui/Button.astro
```diff
@@ -1,9 +1,30 @@
...
+import { cva, type VariantProps } from 'class-variance-authority';
 
-interface Props {
-  color?: 'orange' | 'green' | 'lightblue' | 'semidarkblue';
-  size?: 'sm' | 'lg' | 'lg-full' | 'sm-mobfull';
+import SHA from '@/components/ui/SHA.astro';
+import Link from '@/components/util/Link.astro';
+
+export const buttonVariants = cva(
+  'font-bold text-white drop-shadow transition-colors rounded-full text-center',
```
thilobillerbeck, line 8: This is an interesting design change, though I'm not really a fan of it, this comes especially into play when seeing how buttons behave inside cards as seen on the learn site. The fully rounded corners do not really seem to fit into the more boxy card shape, while the old design aligns more with its parent design and at least in my opinion looks more coherent.

src/components/ui/Button.astro
```diff
@@ -1,9 +1,30 @@
...
+  {
+    variants: {
+      color: {
+        orange: 'bg-nix-orange hover:bg-nix-orange-hover',
+        lightblue: 'bg-nix-blue-light hover:bg-nix-blue-light-hover',
+        semidarkblue: 'bg-nix-blue-dark hover:bg-nix-blue-dark-hover',
+        green: 'bg-nix-green hover:bg-nix-green-hover',
```
thilobillerbeck, line 15: In my testing I couldn't really tell how the hover state applied, am I missing something or is it just to subtile? I'd guess the mentioned colors may need to be defined first. :D

thilobillerbeck (changes requested):
Introducing `class-variance-authority` is definitely a good idea, though I have two small nitpicks. :)

thilobillerbeck:
This is almost right. Though the original design had `rounded-2xl`  for `lg*` buttons and `rounded-3xl` for `sm*` buttons. This was not by accident and I'd appreciate if we keep it that way. We could discuss if we unify the 3xl buttons into a 2xl border radius, but especially in combination with other elements, 2xl makes more sense.

Sorry to bother you again. If I had the permission to push to this PR, I'd just fix it myself. :/

SvMak:
Sorry for late (((

thilobillerbeck (approved):

thilobillerbeck:
no problem :D looks alright now, thx for contributing :)

PR #1612, opened 2024-12-19
blog: show post names in titles

### LINKED ISSUE #727: Blog post names should be in the titles
**Introduce your stance**
Searching for pills on search engines like DuckDuckGo is rather hard since they are all titled the same, but searches often contain multiple pills in their results.

**Describe the issue**
The names in the search engine are all the same since the blog article names are not in the title of the corresponding webpage. It just says "NixOS - Nix Pills"

**Page links**
https://nixos.org/guides/nix-pills/why-you-should-give-it-a-try.html  
https://nixos.org/guides/nix-pills/install-on-your-running-system.html
and so on...

flyfloh:
The change in this PR shows the blog post title in the page title. Closes #727.

### DIFF
--- a/src/pages/blog/[category]/[year]/[id].astro
+++ b/src/pages/blog/[category]/[year]/[id].astro
@@ -20,7 +20,7 @@ const { entry } = Astro.props;
 const { Content } = await entry.render();
 ---
 
-<Layout title="Blog">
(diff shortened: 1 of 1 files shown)


thilobillerbeck (approved):
Well spotted, thanks for fixing this <3

PR #1627, opened 2025-01-16
sponsors: add oakhost

zimbatm:
OakHost generously offered the infra team 2 M2s, plus a discount for
future machines. Thanks OakHost!


zimbatm:
Btw, I wanted to run a local devshell to review the changes, but gave up after `npm run dev` took more than 15min, invoking nix multiple times, and asking me about the same untrusted caches each time. Isn't it possible to have the preview server startup instantly like on most nodejs stacks?

thilobillerbeck:
> the preview server startup instantly like on most nodejs stacks?

The problem is that all the CSS, Markdown etc. has to be compiled in order to display the webpage. Or what kind of preview server do you have in mind?

zimbatm:
After digging a bit, what's slowing things down is mostly the prebuilds. After that, it takes 2-3 seconds from running `astro dev` to getting a working website. I think I would commit the content of the prebuild since they aren't derived from the website content but are additional pieces. Or find a way to fetch the pre-built versions from somewhere.

thilobillerbeck:
Yeah, maybe I can also put some work into speeding things up. I guess adding some command to run `astro dev` alone would be a good idea.

thilobillerbeck:
okay, I guess we have to handle the doc inclusion differently, but I may have an idea for that

PR #1634, opened 2025-01-20
astro: don't set up file watchers for direnv

algorithmiker:
astro would try to set up inotify watchers for all files in direnv, including but not exclusive to a full nixpkgs, which would spam errors like the following in the console and refuse to start:

Error: ENOSPC: System limit for number of file watchers reached, watch 'nixos-homepage/.direnv/flake-inputs/nk036dp7q03yqkikxlfsapljhgnhxkka-source/default.nix' (for each file in nixpkgs).

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -34,5 +34,12 @@ export default defineConfig({
       theme: syntaxTheme,
     },
   },
+  vite: {
+    server: {
+      watch: {
+        ignored: ['**/.direnv/**']
+      }
+    }
+  },
   compressHTML: true,
 });


thilobillerbeck:
Issues with the development environment came up recently and we are aware and will investigate.

> Stop this madness.

This line wouldn't really have been necessary. ;)

thilobillerbeck:
> oops, this seems to disable inotify watchers for all files (or my system is broken).
> 
> it'd be appreciated if someone could help me figure out the correct regex.

I guess btw. its neither your system nor inotify. :D
https://github.com/NixOS/nixos-homepage/issues/1641

PR #1649, opened 2025-02-03
Migrate to Tailwind v4 & vite ignore .direnv

thilobillerbeck:
Tailwind 4 was release recently, this PR
- merges the latest dependabot update
- upgrades to tailwind 4
- replaces the astro tailwind plugin with vite
- migrates the sourcecode according to tailwinds changes
- ignores .direnv dir in vites dev server


thilobillerbeck:
So I hopefully removed all regression introduced by upgrading. Would be great if you especially have a look at if you notice any regressions in the built page compared to the current Tailwind 3 version. :)

Issue #1661, opened 2025-02-16
Some partner logos not showing in all browsers on the donate page.


djacu:
**Describe the glitch**
4TU and Equinix logos do not appear on the /donate page in Firefox on desktop or Firefox on Android mobile but they do in Chrome on desktop and Safari Mobile.

**Screenshots**
Firefox Desktop

![Image](https://github.com/user-attachments/assets/a6036409-267f-477f-8889-c64b74bb1451)

Chrome Desktop

![Image](https://github.com/user-attachments/assets/9460b2d2-f9dc-4b7a-b473-7d774c4040d4)

**Additional context**

PR #1666, opened 2025-02-17
Fix sponsor SVGs on firefox

jasonmishi:
Closes: #1661

Firefox actually renders the images but at 0x0 so it is not visible. Fixed by setting a width and height based on the svg's viewbox to display the image with a non 0x0 dimensions.


djacu (commented):
Logos look correct locally 

image

djacu (approved):

thilobillerbeck (approved):

PR #1672, opened 2025-02-28
Update packages count

AlexAntonik:
Updated according to https://search.nixos.org/


thilobillerbeck (approved):
lookgs good,
thanks for changing this <3

Issue #1614, opened 2024-12-22
endpoint for experimental installer


abathur:
Hi @NixOS/marketing-team,

Not sure what we'll need to close the gap(s) here, but we hope to get releases of https://github.com/NixOS/experimental-nix-installer hosted under nixos.org. (Not replacing existing installer endpoints--it should clearly be an experiment.)

There's background in the infra issue we opened earlier this year:
- https://github.com/NixOS/infra/issues/452

Reaching out as the infra team suggested in issue above now that we have a release workflow set up.

The [current tentative prerelease](https://github.com/NixOS/experimental-nix-installer/releases/tag/0.27.0) should be accessible via something like:
```shell
curl --proto '=https' --tlsv1.2 -sSf -L https://github.com/NixOS/experimental-nix-installer/releases/download/0.27.0/nix-installer.sh | sh -s -- --help
```

I'm not sure how ~controlled the sanctioning of a latest release should be, but I did also briefly un-toggle the prerelease status in order to confirm the latest release per github would also be accessible as:
```shell
curl --proto '=https' --tlsv1.2 -sSf -L https://github.com/NixOS/experimental-nix-installer/releases/latest/download/nix-installer.sh | sh -s -- --help
```

(I toggled it back to prerelease since we've already had people ~depend on this installer as we work, making me skittish about leaving a normal release sitting around for unknown amount of time without a clear go-ahead.)

cc @tomberek @roberth @mkenigs

Mic92:
Ok. Will see what it takes to make it happen.

Mic92:
@abathur should we just make this a redirect on the homepage?

abathur:
@Mic92 Jonas suggested a redirect in https://github.com/NixOS/infra/issues/452#issuecomment-2226855264, and I'm inclined to defer to anyone with a clear opinion on it for now :)

(I'm unsure if a redirect is how we'd want to handle this permanently if adopted as the official installer, but I think it is as good a place to start as any. I suspect it'll be easier to have a better version of this conversation nearer official adoption when more stakeholders are up to speed, if that makes sense.)

(also cc @idabzo @mkenigs @tomberek since we discussed this about an hour ago in the installer WG meeting)

Mic92:
Ok. I finished my research. First we need a redirect on the homepage:

This is just a file here:

```
[...]

PR #1677, opened 2025-03-03
add unstable iso image download

### LINKED ISSUE #635: NixOS unstable channel installer no longer findable
**Introduce your stance**
I have been using NixOS for a while, and each time I install it to a new machine I got to nixos.org to download the installer. For my personal machines, I've always used the unstable channel because I like having the newest software and am not too afraid of things being broken - NixOS has easy roll backs after all.

I understand this is not the recommended default path, but it always used to be possible to find with a few clicks.

**Describe the issue**
On https://nixos.org/download.html I can not find any reference to an unstable installer, nor even an "Advanced options" link where I could find that.

**Page links**
https://nixos.org/download.html#nixos-iso

**Additional context**
N/A

flyfloh:
This MR adds the download links to the unstable image again.

I chose to add an unstable ISO link to the NixOS sidebar on the download page (see screenshot below).

Screenshot From 2025-03-03 17-35-00

If this is confusing, we could also add a separate section for the unstable downloads. I am not 100% sure what the best way to approach this is.

Fixes #635.


thilobillerbeck:
Looks alright, though I'd opt we should put this in the ISO image tab and not in a separate one, since this kind of breaks the semantics of the download page now.

thilobillerbeck:
Notes from the meeting

- split the ISO image sections into Graphical and Minimal
- having a section for either stable and unstable downloads

flyfloh:
I've updated this with a section for the graphical / minimal installers and added headings for Stable and Unstable within those.

thilobillerbeck (approved):

ElvishJerricco:
This was merged at literally the worst possible time :P We were just about to change the ISOs on unstable so there's only one combined graphical ISO with both variants as boot entries: https://github.com/NixOS/nixpkgs/pull/355893

Once that's merged and the channel updates, these links will be invalid. I didn't realize there was an intention to add the unstable ISOs to the homepage, or else I would have warned about the upcoming change. I had expected I would coordinate with the 25.05 release manager about updating the homepage along with the 25.05 update.

I have a PR to account for the change: https://github.com/NixOS/nixos-homepage/pull/1700

Though, frankly, I wonder if it would be simpler to just revert this PR and redo it as part of the 25.05 release?

thilobillerbeck:
Hi

> This was merged at literally the worst possible time :P

Well, that's unfortunate :D, but we can fix this

> I had expected I would coordinate with the 25.05 release manager about updating the homepage along with the 25.05 update.

Maintaining and updating the homepage is at the end the marketing team's task, and we decided to address this improvement. Since this is not addressing a stable release, we were not aware of any possible changes. This is btw. one of the reasons we transparently document our coordination meetings, this issue was tracked over multiple weeks. A good example on how to prevent those situations is how the experimental nix-installer people handled it by opening a tracking issue https://github.com/NixOS/nixos-homepage/issues/1614.

> I have a PR to account for the change: https://github.com/NixOS/nixos-homepage/pull/1700

Awesome, let's coordinate timely merging then :)

> Though, frankly, I wonder if it would be simpler to just revert this PR and redo it as part of the 25.05 release?

I guess we can just merge this as soon as your changes go live

Issue #1611, opened 2024-12-17
Migrate to Content Layer API


thilobillerbeck:
Due to the upgrade to Astro v5 the Content Collections v2 API is now deprecated, to keep the site and all upgrades moving I temporarily enabled the legacy collection support. I postponed the migration because it requires some more work. This issue is used to track it. 

Migration guide: https://docs.astro.build/en/guides/upgrade-to/v5/#updating-existing-collections

PR #1687, opened 2025-03-10
Migrate to Astro 5 content layer API

albertchae:
Fixes: https://github.com/NixOS/nixos-homepage/issues/1611

Following https://docs.astro.build/en/guides/upgrade-to/v5/#updating-existing-collections, changes of the same type are grouped into commits

- turn off legacy.collections and define `src/content.config.ts`. Previously we were getting the content config autogenerated by directory. This is technically still available as a deprecated feature, but it wouldn't work for us anyway because it only globs on md/mdx files
  
  ```
  Auto-generating collections for folders in "src/content/" that are not defined as collections.
  This is deprecated, so you should define these collections yourself in "src/content.config.ts".
  The following collections have been auto-generated: blog, community, download, explore, landing, landing-features, learning-manuals, menus, research, sponsors, teams
  ```

  ```
  08:35:29 [WARN] [glob-loader] No files found matching "**/*{.md,.mdx},!**/_*/**/*{.md,.mdx},!**/_*{.md,.mdx}" in directory "src/content/community"
  08:35:29 [WARN] [glob-loader] No files found matching "**/*{.md,.mdx},!**/_*/**/*{.md,.mdx},!**/_*{.md,.mdx}" in directory "src/content/explore"
  08:35:29 [WARN] [glob-loader] No files found matching "**/*{.md,.mdx},!**/_*/**/*{.md,.mdx},!**/_*{.md,.mdx}" in directory "src/content/landing"
  08:35:29 [WARN] [glob-loader] No files found matching "**/*{.md,.mdx},!**/_*/**/*{.md,.mdx},!**/_*{.md,.mdx}" in directory "src/content/menus"
  08:35:29 [WARN] [glob-loader] No files found matching "**/*{.md,.mdx},!**/_*/**/*{.md,.mdx},!**/_*{.md,.mdx}" in directory "src/content/sponsors"
  08:35:29 [WARN] [glob-loader] No files found matching "**/*{.md,.mdx},!**/_*/**/*{.md,.mdx},!**/_*{.md,.mdx}" in directory "src/content/research"
  ```
- replace slug with id. Did not do the same for `src/pages/community/teams/[...slug].astro` because it wasn't strictly necessary and the example in the migration guide has an example with `[slug]` in the filename getting populated from a collection entry in `getStaticPaths`

- Switch to the new render() function.
  > Entries no longer have a render() method, as they are now serializable plain objects. Instead, import the render() function from astro:content.

- Manually sort collections when necessary. Previously we were relying on collection entries being sorted lexicographically from the directory, but this behavior has now changed. Except for blog entries which are already being sorted by date, we explicitly sort lexicographically (using [`localeCompare`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) because [array sort wants a comparator function that returns negative, 0 or positive to indicate order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

  > Sort order of generated collections is non-deterministic and platform-dependent. This means that if you are calling getCollection(), the order in which entries are returned may be different than before. If you need a specific order, you must sort the collection entries yourself.

### DIFF
--- a/src/content.config.ts
+++ b/src/content.config.ts
@@ -1,0 +1,66 @@
+import { defineCollection, z } from 'astro:content';
+import { glob } from 'astro/loaders';
+
+const blog = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.{md,mdx}', base: './src/content/blog' }),
+});
+
+const community = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/community' }),
+});
+
+const download = defineCollection({
+  loader: glob({
+    pattern: '**/[^_]*.{md,mdx}',
+    base: './src/content/download',
+  }),
+});
+
+const explore = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/explore' }),
+});
+
+const landing = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/landing' }),
+});
+
+const landingFeatures = defineCollection({
+  loader: glob({
+    pattern: '**/[^_]*.{md,mdx}',
+    base: './src/content/landing-features',
+  }),
+});
+
+const learningManuals = defineCollection({
+  loader: glob({
+    pattern: '**/[^_]*.{md,mdx}',
+    base: './src/content/learning-manuals',
+  }),
+});
+
+const menus = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/menus' }),
+});
+
+// TODO get `research` from astro collection instead of hardcoded json
+
+const sponsors = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/sponsors' }),
+});
+
+const teams = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.{md,mdx}', base: './src/content/teams' }),
+});
+
+export const collections = {
+  blog,
+  community,
+  download,
+  explore,
+  landing,
+  landingFeatures,
+  learningManuals,
+  menus,
+  sponsors,
+  teams,
+};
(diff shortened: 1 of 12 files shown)


src/content.config.ts
```diff
@@ -0,0 +1,66 @@
...
+    pattern: '**/[^_]*.{md,mdx}',
+    base: './src/content/learning-manuals',
+  }),
+});
+
+const menus = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/menus' }),
+});
+
+// TODO get `research` from astro collection instead of hardcoded json
```
albertchae, line 45: it looks like research is using this json inside the astro file https://github.com/NixOS/nixos-homepage/blob/c0b1510bde08beb21d69c7d837a76d2c04365286/src/pages/research.astro#L227 instead of the content collection. I see some discussion about this in a related PR https://github.com/NixOS/nixos-homepage/pull/1602

src/content.config.ts
```diff
@@ -0,0 +1,66 @@
...
+  loader: glob({ pattern: '**/[^_]*.{md,mdx}', base: './src/content/blog' }),
+});
+
+const community = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.yaml', base: './src/content/community' }),
+});
+
+const download = defineCollection({
+  loader: glob({
+    pattern: '**/[^_]*.{md,mdx}',
```
albertchae, line 14: should we glob every directory on md,mdx,yaml or is this fine? it might be annoying in the future if we add a new filetype to a collection and forget to edit this config

src/pages/community.astro
```diff
@@ -461,9 +463,7 @@ import nixosFoundationLogo from '../assets/image/nixos-foundation-logo.svg';
         <ul class="mx-auto mt-2 list-disc pl-8 md:w-72 md:pl-10">
           {
             teams
-              .filter(
-                (team) => team.slug.split('_')[1] === 'foundation-board',
-              )[0]
+              .find((team) => team.id.split('_')[1] === 'foundation-board')
```
albertchae, line 466: small refactor to use `find` instead of `filter(...)[0]`

src/content.config.ts
thilobillerbeck: yeah, lets tackle this in the PR you mentioned since it does not hurt content collection migration right now

thilobillerbeck (approved):
Awesome work! Thanks for working on this! :)

PR #1689, opened 2025-03-10
Add Sponsors to landing page

thilobillerbeck:
As discussed in the marketing meeting, we want to give sponsors more visibility. As such, we plan to move the sponsors to the landing page.


src/pages/index.astro
```diff
@@ -123,6 +124,25 @@ posts
       </Container>
     </div>
   </div>
+  <div>
```
jtojnar, line 127: The [donate page](https://nixos.org/donate/) clarified that “following companies and institutes have supported the development of NixOS and hosting for its infrastructure”. Here it is not clear what the logos mean. Landing pages often show users or endorsements so it might be important to the sponsors themselves for the relationship to be made clear.

src/pages/index.astro
```diff
@@ -123,6 +124,25 @@ posts
...
+    <Container
+      class:list={[
+        'flex flex-wrap items-center justify-center gap-8 pt-4 pb-16 md:justify-between',
+      ]}
+    >
+      {
+        sponsors.data.items.map((item) => (
```
jtojnar, line 134: The acknowledgements section on the donate page also includes institutions that have contributed to Nix in the past (at least we are not sure they do nowadays). Should the landing page continue to show all the past sponsors and supporters forever, only the foundational ones (Utrecht University), or just the current ones?

PR #1703, opened 2025-03-20
mention and link alternative implementations

mschwaig:
The project started by Nix has grown into a large and dynamic ecosystem.
We should start maintaining an overview of this ecosystem.
In particular, we should acknowledge the existence of related projects in the functional package management space.

### DIFF
--- a/src/pages/community.astro
+++ b/src/pages/community.astro
@@ -5,9 +5,9 @@ import { getCollection, getEntry } from 'astro:content';
 import Container from '../components/layout/Container.astro';
 import Divider from '../components/layout/Divider.astro';
 import PageHeader from '../components/layout/PageHeader.astro';
+import CommunityPageListItem from '../components/pages/community/CommunityPageListItem.astro';
 import NixConEntry from '../components/pages/community/NixConEntry.astro';
 import OfficialCommunityLink from '../components/pages/community/OfficialCommunityLink.astro';
-import UnofficialCommunityLink from '../components/pages/community/UnofficialCommunityLink.astro';
 import Banner from '../components/ui/Banner.astro';
 import Button from '../components/ui/Button.astro';
 import RfcDiagram from '../components/ui/RfcDiagram.astro';
@@ -161,6 +161,29 @@ const platformsOutside = [
   },
 ];
 
+const widerEcosystem = [
+  {
+    href: 'https://guix.gnu.org/',
+    name: 'Guix',
+  },
+  {
+    href: 'https://tvix.dev/',
+    name: 'Tvix',
+  },
+  {
+    href: 'https://lix.systems',
+    name: 'Lix',
+  },
+  {
+    href: 'https://determinate.systems/nix/',
+    name: 'Determinate Nix',
+  },
+  {
+    href: 'https://snix.dev/',
+    name: 'Snix',
+  },
+];
+
 import nixosFoundationLogo from '../assets/image/nixos-foundation-logo.svg';
 ---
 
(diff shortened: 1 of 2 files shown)


mschwaig:
Screenshot 2025-03-20 at 01-45-26 Community Nix   NixOS

Lassulus:
sounds good to me, I personally think we should work close with the wieder ecosystem anyway, so good to mention them. Maybe we can choose an alphabetic ordering or add a small comment, that they are in historical order if we want to keep it in the current way? :)

fricklerhandwerk (approved):

aforemny (approved):

thilobillerbeck:
Hey, thanks for implementing this change. I unfortunately had no chance to put in my perspective, so I'll leave some notes in here:

- Maybe we should add logos to the buttons, they look a bit out of place now in comparison to all other logos that have buttons
- Is the "Where to talk", the section where we have sources for social interaction, really be the section for social interaction?  Should we maybe have an extra section for that to give these projects some more room?
- This is not specifically related, but: how about extending `/download` in the long term to allow for choosing fork specific installers?

fricklerhandwerk:
@thilobillerbeck we discussed these issues as well, and came to the following conclusions:
- Yes, and can add that later, it would have blown up the scope of the change.
- Right, it should be a `<h2>`
- We thought about that, but considered it too intrusive. There are already too many entry points as it is.

thilobillerbeck:
> * Right, it should be a `<h2>`

I feel like separating this even more (giving it a dedicated section), but yeah, hierarchically it does not belong to socials :D

SomeoneSerge:
I believe this is a step in the right direction. Maintaining the silence about other projects, unintended in the first place, does not help the documentation and SEO of the NixOS project either. New users looking into "Nix" will surely hit search results for Lix and Snix and others as well. They might not be able to recover the context for the existence of all these independent efforts or the collaborations between them, in which case they'll end up reading uninformed and confused conversations on reddit.

IMO we should merge the current minimal version, but in future we should also include brief comments about each project, how it's positioned wrto "Nix", the core ideas and differences behind each effort. E.g. people will often find mentions that Guix "sets different goals and isn't directly comparable to Nix", but it's not easy to find what these "fundamental differences" are. NixOS-org is the most obvious space to maintain a discoverable "source of truth" for documenting these nuances.

Given our experiences with listing sponsorships and "unofficial communication channels", it is important that we communicate that these references are documentation and not endorsements. They are not meant to promote or advertise any particular project.

fricklerhandwerk:
@thilobillerbeck sure we can give it a dedicated section, but then we'll likely have to tear apart the whole site organisation. Many things are already hard to find because they're so far below the fold, and piling up more stuff taking lots of space on each page will not make that better. This is why I'd opt for a conservative approach until we find the courage/time to do bigger shifts.

SomeoneSerge:
We could also introduce a separate page featuring comparison and compatibility tables. E.g. to people not actively involved in the development it may not be obvious to which extent Nix, Snix and Lix are interoperable, or even that they can be used simultaneously. This could also be a more ambitious project aiming to map the whole design space of Nix-like solutions ("FPM"s).

Yet another axis to cover is the existence of alternatives to Nixpkgs and NixOS: aux, sixos, etc.

thilobillerbeck:
> IMO we should merge the current minimal version

Just to be clear, I'm not against this, I'm absolutely pro mentioning. I was just not really happy how this was handled and since I don't want to revert this I provided feedback nonetheless. 
We as the marketing team are responsible for this project, and we really don't want to annoy people with our reviews and iteration cycles, we just want to make sure everyone is happy with the final result and that it fits with our plans. Since we try to put a lot of effort into this, we'd be really happy to be involved discussing changes. I really try to be as responsive as possible.

fricklerhandwerk:
@thilobillerbeck thanks for tending to these change requests. Sorry, I was too trigger-happy today and merged it on a whim despite the group at Ocean Sprint having explicitly discussed yesterday that we would want these conversations to happen in public based on the code.

src/pages/community.astro
```diff
@@ -280,7 +303,28 @@ import nixosFoundationLogo from '../assets/image/nixos-foundation-logo.svg';
...
+              />
+            ))
+          }
+        </ul>
+      </p>
+      <h3 id="forks" class="font-heading text-nix-blue mt-6 text-3xl font-bold">
+        The wider ecosystem
+      </h3>
+      <p class="text-gray-700">
+        There are a few alternative implementations of the idea of purely
```
SomeoneSerge, line 321:
```suggestion
        There are a few independent implementations of the idea of purely
```

SomeoneSerge:
```suggestion
        There are several alternative implementations of the idea of purely
```

src/pages/community.astro
```diff
@@ -280,7 +303,28 @@ import nixosFoundationLogo from '../assets/image/nixos-foundation-logo.svg';
...
+        functional package managers (sorted chronologically by appearance):
```
SomeoneSerge, line 322:
```suggestion
        functional package management (sorted chronologically by appearance):
```

PR #1714, opened 2025-03-26
Restructure project files

thilobillerbeck:
This PR helps resolve some build time issues and live reload hickups by moving the core part of the website into its own workspace, called `/core`. Thus, the website's source does not conflict with anything that is introduced by the flake and its external builds. 

In short
- move the core website into `/core`
- Introduce npm workspaces
- more clearly separate the core website build and external builds such as the manual


thilobillerbeck:
btw. the CI will fail for this since code moved and this will try to build with the CI from the target branch

djacu (approved):

Issue #1648, opened 2025-02-03
Explain NixOS on the nixos.org landing page


flyfloh:
I have been thinking on how we can structure the content on the website to
create less surprises for people interested in NixOS, while also keeping the
importance of Nix (as the underlying & independent tool that makes it all possible).

We have had a number of issues (e.g. #734, #828, #993) and discussions on
Discourse about this topic, which at least hints to the fact that this is a
real issue.

In order to improve the situation I propose to highlight NixOS and *its* benefits
on the nixos.org landing page. This way, new users' expectations should be
better met. The current landing page is a really good landing page, but it explains in a lot of detail. This in turn creates the problem that new users are confused because they may not yet know what this "Nix" thing is.

It should help to talk more about NixOS on the landing page, but prominently mention `nix` as the thing that powers it all in the text(s).

See an **example proposal** (I have not spent much time word-smithing things yet) below:

![Image](https://github.com/user-attachments/assets/aac5c049-fcb5-4ed5-b5ca-802c5d129b7b)

On the `/learn` page, we should start at the top with an explanation of what
`nix`, `nixpkgs` and `NixOS` are. Again see the image below for an example text:

![Image](https://github.com/user-attachments/assets/bc78c0e1-bff6-4977-afc3-c0ee2c69fed5)

I think we should also help people find the NixOS download on the `/download`
page better. I am not sure what the right approach here is though. One option
would be to create two separate download pages for Nix and NixOS.

I'm willing to work further on this, but before I do I wanted to get a feel for
what the community thinks in general about this direction.

thilobillerbeck:
![Image](https://github.com/user-attachments/assets/b2d71f8f-7d8c-42c0-ad94-a9498cb0b25f)

I just quickly sketched my idea down as a wireframe. The goal in my opinion is to show people that both options exist. Let me know what you think. Maybe we can also continue this explanation further down the page, redoing the whole landing page.

gilice:
I had a similar idea previously: I think https://nixos.org/explore/ explains Nix(OS) better than the homepage, and maybe could be merged into the homepage as a whole.
I am curious what people think.

thilobillerbeck:
I think "merging" it would make the landing page pretty long, which is not really what we want to achieve. Though redesigning the homepage to be like explore is definitely something that'd be great imo.

fricklerhandwerk:
[...]

PR #1726, opened 2025-04-07
Explain NixOS on the landing page

flyfloh:
This PR adds an explanation of Nix and NixOS to the landing page. It also adds a bit of definition of Nix, NixOS and Nixpkgs to the `/learn` page.
Both changes were previewed and discussed in #1648.

### DIFF
--- a/core/src/pages/index.astro
+++ b/core/src/pages/index.astro
@@ -45,42 +45,52 @@ posts
       </Banner>
     )
   }
-  <div class="bg-nix-blue inline-svg-hero bg-cover bg-left bg-no-repeat">
-    <Container
-      class="grid grid-cols-1 gap-8 py-16 text-white lg:grid-cols-2 lg:py-24"
-    >
-      <div class="mx-0 my-auto flex flex-col gap-4">
+  <div
+    class="bg-nix-blue inline-svg-hero bg-cover bg-left bg-no-repeat text-white"
+  >
+    <Container class="pt-16">
+      <div class="mx-0 my-auto w-full gap-4">
         <h1 class="font-heading text-5xl leading-none font-bold">
-          Declarative builds<br class="hidden md:inline-block" /> and deployments.
+          Build reproducible development environments, simplify your
+          deployments!
         </h1>
+      </div>
+    </Container>
+
+    <Container class="grid grid-cols-1 gap-8 py-16 lg:grid-cols-2 lg:py-24">
+      <div class="mx-0 my-auto flex flex-col gap-4">
+        <h2 class="font-heading text-5xl leading-none font-bold">Nix</h2>
         <p class="text-lg leading-7 font-extralight">
-          Nix is a tool that takes a unique approach to package management and
-          system configuration. Learn how to make reproducible, declarative and
-          reliable systems.
+          The stand-alone package manager that ensures reproducible builds and
+          makes sure that everything that runs on one machine will also work on
+          another.
         </p>
         <div
           class="flex flex-col justify-center gap-4 text-center md:flex-row lg:justify-start"
         >
           <Button href="/download" color="green" size="lg">Download</Button>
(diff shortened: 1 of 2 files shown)


thilobillerbeck:
Looks good, though I feel like we should make both sides more distinguishable

flyfloh:
I've played around a bit with setting different background colors on both sides. Once I add a padding Astro makes them different sizes (which is weird). Other than that it looks pretty good. What do you think?

Screenshot From 2025-04-28 19-32-53

thilobillerbeck:
Hmm, I have multiple things here that I still don't like:

- the boxes don't really blend into the background
- the text looses contrast on light backgrounds
- The distinction is still not that noticeable, maybe we can add the respective Logos there or use icons that distinguish the usecase on first sight
- I think we should make the color differences blend in with the background

core/src/pages/index.astro
```diff
@@ -45,42 +45,52 @@ posts
...
         </p>
         <div
           class="flex flex-col justify-center gap-4 text-center md:flex-row lg:justify-start"
         >
           <Button href="/download" color="green" size="lg">Download</Button>
-          <Button href="/learn" color="lightblue" size="lg">Get started</Button>
+          <Button
+            href="https://nix.dev/tutorials/first-steps/"
+            color="lightblue"
+            size="lg">Get started</Button
```
fricklerhandwerk, line 75:
```suggestion
            size="lg">Get started with Nix</Button
```

Let's be specific what to get started with.

core/src/pages/index.astro
```diff
@@ -45,42 +45,52 @@ posts
...
+        <p class="text-lg leading-7 font-extralight">
+          Run your favourite webservice in 5 minutes: NixOS is a Linux
+          distrubition built on Nix and adds a powerful module system that
+          enables running your favourite application in no time.
+        </p>
+        <div
+          class="flex flex-col justify-center gap-4 text-center md:flex-row lg:justify-start"
+        >
+          <Button href="/download#nixos-iso" color="green" size="lg"
+            >Download</Button
```
fricklerhandwerk, line 90:
```suggestion
            >Download NixOS</Button
```

core/src/pages/index.astro
```diff
@@ -45,42 +45,52 @@ posts
...
-              [122.0, 'Build minimal docker image with Nix'],
-            ],
-          }}
-        />
+      <div class="mx-0 my-auto flex flex-col gap-4">
+        <h2 class="font-heading text-5xl leading-none font-bold">NixOS</h2>
+        <p class="text-lg leading-7 font-extralight">
+          Run your favourite webservice in 5 minutes: NixOS is a Linux
+          distrubition built on Nix and adds a powerful module system that
+          enables running your favourite application in no time.
```
fricklerhandwerk, line 84:
```suggestion
          NixOS is a Linux distribution that can be configured fully declaratively with the Nix language.
          It leverages Nix and the Nixpkgs software repository for correct deployments of even the most complex setups.
```

I don't think we can promise all that much more.

thilobillerbeck:
Disregarding what we put at the top, I'd go for something like this with the two options the user could choose from. These boxes clearly name what they present, what it is, and declare its use in one sentence. Maybe we could put this as a box under the hero and keep the hero more clean and reduce the sensory load.

image
(the flake background is missing here, but just imagine it'd be there :D)

core/src/pages/index.astro
fricklerhandwerk: I agree with focusing on the outcomes. But the module system is an implementation detail. Also it's not *that* special, it's "just" a dependently typed DSL for the Nix language. Which is indeed important, but a cryptic piece of information for most readers, so not very outcome-oriented. What would be an outcome of having a declarative, dependently typed language for configuring computers?

PR #1727, opened 2025-04-08
Reintroduce nix build functionality

thilobillerbeck:
This PR reintroduces `nix build` functionality for the homepage and also serves external dependencies like manuals, pills etc. via a development server.


djacu (changes requested):
Tested on NixOS and Linux Mint

`nix build .#core` works
`nix build .#default` works
After ^^^ I can spin up a simple HTTP server with Python and test out the built webpage. It looks correct.

The development workflow appears to work but with 1 regression. Hot reloading doesn't appear to refresh the page automatically on write. I have to manually refresh the page but the new content does load. It is a much better experience than before when you had to rebuild but it would be nice if refresh-on-write still worked.

djacu (approved):
:shipit:

PR #1753, opened 2025-05-12
Make install command more secure

### LINKED ISSUE #1707: Install command is insecure
**Introduce your stance**
Curl's flags in Nix (the package manager) `curl | bash` install command are not secure enough.

**Describe the issue**
Currently https://nixos.org/download suggests using

```
sh <(curl -L https://nixos.org/nix/install) --daemon
```

but this does not enforce neither HTTPS, nor TLS 1.2.

Although I don't like `-L` because such redirects may cause MITM attack, I assume HTTPS would prevent this. And if the server itself is compromised nothing stops the attacker from uploading their own malicious install script.

**Suggestion**
Update the install command to include `--proto '=https'` and `--tlsv1.2` flags. The same way [rustup](https://www.rust-lang.org/tools/install) does its `curl | bash`.

```shell
sh <(curl --proto '=https' --tlsv1.2 -L https://nixos.org/nix/install) --daemon
```

The command would no longer be as aesthetically pleasing, sure. But I'd argue it's much more important for it to be safe than pretty.

synalice:
Closes #1707.

### DIFF
--- a/core/src/content/download/01-nix-linux.mdx
+++ b/core/src/content/download/01-nix-linux.mdx
@@ -9,7 +9,7 @@ platform: linux
 Install Nix via the <span class="font-bold">recommended</span> [multi-user installation](/manual/nix/stable/installation/multi-user):
 
 ```bash
-$ sh <(curl -L https://nixos.org/nix/install) --daemon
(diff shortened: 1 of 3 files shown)


thilobillerbeck (approved):
Looks alright, thanks for implementing this change <3

PR #1762, opened 2025-05-24
learn: clarify that we link to *the* official wiki

stapelberg:
The previous wording was unnecessarily imprecise:
“a wiki” instead of “the wiki”, “user-maintained”, no mention of “official”.

The new wording keeps the connotation of welcoming user contributions (and makes that more explicit), but also makes it clear that this is the official one.

(Created as part of #ZurichZHF)

### DIFF
--- a/core/src/pages/learn.astro
+++ b/core/src/pages/learn.astro
@@ -42,7 +42,8 @@ const learningResources = [
   },
   {
     title: 'NixOS Wiki',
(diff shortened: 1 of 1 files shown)


thilobillerbeck (approved):

fricklerhandwerk:
Reviewed at #ZurichZHF (enabling a query for the report :)

PR #1765, opened 2025-05-25
update all buttons with better colors and icons

thilobillerbeck:
This PR basically removes all orange and green colors and as such improves contrast on buttons, text and other UI elements significantly.
Some buttons also now have icons to more clearly signify their function.


fricklerhandwerk (commented):
Nice! #ZurichZHF ftw

PR #1770, opened 2025-05-30
Marketing team -> Outreach team

fricklerhandwerk:
ther rename is intended shake off any vibes that someone is trying to sell snake oil around here

If accepted, this may need updating (adding to) the email alias and the GitHub team handle. Somewhat related, it would be good to update the actual team page contents to reflect the current reality of what the team does, why, and how it operates.


fricklerhandwerk:
Closed, as feedback on Discourse does not seem in strong favor of this, but feel free to reopen and use the change if it turns out to be desirable.

Issue #1715, opened 2025-03-26
commercial support page - list not randomized


Jappie3:
Just tried 4 browsers across 2 devices, the list on https://nixos.org/community/commercial-support/ doesn't seem to be randomized - even though this is stated at the top of the page.

thilobillerbeck:
Hi, they are. But not at every request but after every build. :)

for reference: https://github.com/NixOS/nixos-homepage/blob/f0ee6e65e96cd83c8df5f3be6364128fbacf9efc/src/pages/community/commercial-support.astro#L10-L26

Jappie3:
ah my apologies, seems like I jumped to conclusions - my knowledge of web frameworks is quite limited 😅

PR #1777, opened 2025-06-13
commercial-support: shuffle companies on each page load

cheriimoya:
This adds client side JavaScript to shuffle the companies on each page load as the page states.

This fixes https://github.com/NixOS/nixos-homepage/issues/1715

I left in the shuffle during build time for people blocking JavaScript.


fricklerhandwerk (approved):
Nice, looks good!

cheriimoya:
Added the "invisible" class so there is no flickering.

thilobillerbeck (changes requested):
Looks alright, though with the `invisible` class added, the site is pretty much unusable without JS enabled. A fallback for those without JS enabled would be pretty neat.

cheriimoya:
True, @thilobillerbeck I didn't think about that. Now, I tested this change with JS enabled and by setting the FF config option javascript.enabled to false. Both worked as intended:)

PR #1814, opened 2025-07-28
Update page colors to match NixOS/branding

thilobillerbeck:
This PR updates the hompage colors according to the https://github.com/NixOS/branding repo. It uses the exposed npm package as a workspace package loaded in via the flake. 

Furthermore this PR updates the commercial support page to bring it more in line with the overall homepage design.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "typescript-eslint": "^8.45.0"
   },
   "dependencies": {
-    "chalk": "^5.6.2",
-    "folder": "^0.0.4"
+    "chalk": "^5.6.2"
   }
 }


flake.nix
```diff
@@ -209,6 +213,15 @@ rec {
 
             npmConfigHook = pkgs.importNpmLock.npmConfigHook;
 
+            makeCacheWritable = true;
+
+            preConfigure = ''
+              mkdir -p @NixOS/branding
+              cp -R --no-preserve=mode,ownership ${
+                branding.hydraJobs.nixos-branding.${system}.npm-package
```
djacu, line 221: Rather than accessing `npm-package` directly through the flake input, wouldn't it be better to use the expose default overlay? This would add `nixos-branding` to the top-level of the import package set and then you could access `npm-package` via `nixos-branding.npm-package`.

flake.nix
```diff
@@ -304,6 +317,14 @@ rec {
               export PATH_PILLS="${pills}"
               export PATH_DEMOS="${demos}"
 
+              if [ -d @NixOS/branding ]; then
+                rm -rf @NixOS/branding
+              fi
+              mkdir -p @NixOS/branding
+              cp -R --no-preserve=mode,ownership ${
+                branding.hydraJobs.nixos-branding.${system}.npm-package
```
djacu, line 325: Same as the other comment

> Rather than accessing `npm-package` directly through the flake input, wouldn't it be better to use the expose default overlay? This would add `nixos-branding` to the top-level of the import package set and then you could access `npm-package` via `nixos-branding.npm-package`.

djacu (changes requested):
This is awesome! I looked around in the deploy preview and everything looks good. I briefly looked through the code changes and left a couple comments.

djacu (approved):
Looks like overlays are causing issues. Probably on my end because the branding repository has fallen behind. I'll look into it

PR #1817, opened 2025-08-02
Lower barrier to donate

tijlleenders:
Donating should require minimal effort.  
This PR removes the need to scroll down on the Donate page.  
Presenting the options to donate is the content that is most likely to be useful for someone clicking on 'Donate' in the top menu.


thilobillerbeck (changes requested):
The slopes moved too with the section. That should be fixed before this can get merged.

tijlleenders:
@thilobillerbeck Nice catch!  
How about this?:
image

thilobillerbeck (approved):
looks perfect now, thx for contributing this change <3

PR #1836, opened 2025-09-06
Add devices page

RossComputerGuy:
The hardware team is looking to add a page which lists devices that are supported. The support falls under 3 "tiers":
- Ships with NixOS
- Official Support
- Community Support

Ships with NixOS is a bunch of devices which ship with NixOS out of the box. This means the user has to only configure the system to their liking. No need for any installation ISO or changing secure boot options.

Official support are devices which the hardware team is able to "guarantee" support. These devices have been received by the hardware team and checked to be supported.

Community supported devices are ones which the hardware team is not able to provide support. A good example is Apple Silicon machines. Those machines are handled by a NixOS community repo.

The goal here is to incentivize vendors to ship with NixOS, encourage sponsorships from vendors, and provide users a central way to find out about hardware support. This has already been shown to @djacu and @Mic92 at NixCon.

### DIFF
--- a/core/src/pages/devices.astro
+++ b/core/src/pages/devices.astro
@@ -32,7 +32,7 @@ const devices = await getCollection('devices');
           </p>
         )}
         {category.data.devices && (
-          <ul class="mt-4 grid list-inside list-disc gap-4 space-y-2 md:grid-cols-3">
+          <ul class="my-4 grid list-inside list-disc gap-4 space-y-2 md:grid-cols-3">
             {category.data.devices.map((device) => (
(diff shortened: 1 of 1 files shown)


djacu:
lgtm
leaving the final review to you @thilobillerbeck

thilobillerbeck:
I just moved the content to its own collection for easier maintainability in the future and gave the device list some design touchups. If you are fine with that @RossComputerGuy I would merge this with my added changes. :)

RossComputerGuy:
Sweet, thanks for doing that. Looks easier to manage.

PR #1843, opened 2025-09-14
Fix double clicks on banner buttons

Erethon:
When clicking on an <a> element in a banner we open the link twice, once from the <a> and once from JavaScript. This is easily visible currently with the big "Take the Survey" button for the Nix Community Survey banner. I don't know if this solution is the best approach, it seems logical to me but I haven't used astro before.

### DIFF
--- a/core/src/components/ui/Banner.astro
+++ b/core/src/components/ui/Banner.astro
@@ -14,7 +14,11 @@ import Container from '../layout/Container.astro';
   const clickables = document.querySelectorAll('.clickable');
   clickables.forEach((clickable) => {
     clickable.classList.add('cursor-pointer');
-    clickable.addEventListener('click', () => {
+    clickable.addEventListener('click', (e) => {
+      // Don't interfere when a user clicks on a button
+      if (e.target.matches('a, button') || e.target.closest('a, button')) {
+        return;
+      }
       clickable.querySelector('a, button').click();
     });
   });


Erethon:
> 🚀 Deployed on https://68c7438b3f2e3d04985a645e--nixos-homepage.netlify.app

This needs a banner in order to be testable. I don't see a way to enable one in the GHA.

PS. I wonder if the deploy-pr action should by default enable a banner for testing in the future.

thilobillerbeck:
Hi, could you maybe tell me in which environment you encountered the bug your PR provides a fix for? I cannot reproduce it.

Erethon:
> Hi, could you maybe tell me in which environment you encountered the bug your PR provides a fix for? I cannot reproduce it.

Firefox 142.0.1 (latest in NixOS 25.05) and also the latest Firefox ESR in Debian 12. You might have to allow pop-ups from nixos.org to view this, because firefox prevented the second tab from opening the first time for me. Attached is a gif showing the double tabs opening.

Reproduced this on Vanadium (GrapheneOS' Chromium fork) v140.0.7339.123. One tab opens, but I get a notification about a pop-up being blocked.

Peek 2025-09-15 02-46

thilobillerbeck:
yup, I reproduced it (indeed it is the case without pop-up blocking), thanks for pointing this out and fixing it <3

PR #1862, opened 2025-10-06
teams(ngi): update team members and roles

imincik:
* Remove Valentin Gagarin
* Add role name for Ivan and Daniel


infinisil:
Can confirm that this is foundation approved :)

PR #1864, opened 2025-10-08
Bootstrap governance page

thilobillerbeck:
Convert the values page into something that is more clear about how the projects governance is working.

### DIFF
--- a/core/src/pages/governance.astro
+++ b/core/src/pages/governance.astro
@@ -51,8 +51,8 @@ const getMemberLink = (member) => {
           leadership.
         </span>
         <span>
-          <b>Responsibilities:</b> Technical and community decisions, delegation
-          and guidance on partnership decisions
+          <b>Responsibilities:</b> Technical and community decisions, delegation and
+          guidance on partnership decisions
         </span>
         <span>
           <b>Members:</b>
@@ -205,9 +205,8 @@ const getMemberLink = (member) => {
 
       <p>
         Values are <span class="italic">not</span> a Code of Conduct, they do not
-        define governance structures, and they do not provide specific policies.
-        Instead, values inform the decisions we make about these things. A good test
-        for <a
+        define governance structures, and they do not provide specific policies. Instead,
+        values inform the decisions we make about these things. A good test for <a
           href="https://medium.com/the-u-s-digital-service/our-values-1fc02b53598"
           >whether something is a value</a
         > is:
--- a/core/src/pages/branding.astro
+++ b/core/src/pages/branding.astro
@@ -34,9 +34,9 @@ import { Image } from 'astro:assets';
             download>NixOS Branding Guide</a
           > outlines the official visual identity of the NixOS project. It provides
           clear guidance on how to use the NixOS logo, colors, typography, and other
-          brand elements across digital and print media. Whether you're creating
-          a blog post, a presentation, a sticker, or a t-shirt, this guide helps
-          ensure your materials reflect the NixOS brand consistently and respectfully.
+          brand elements across digital and print media. Whether you're creating a
+          blog post, a presentation, a sticker, or a t-shirt, this guide helps ensure
+          your materials reflect the NixOS brand consistently and respectfully.
         </p>
         <Button
           color="semidarkblue"


core/src/pages/governance.astro
```diff
@@ -0,0 +1,315 @@
...
+    </p>
+    <h3
+      class="font-heading text-secondary-afghani-blue mt-4 text-3xl font-bold"
+    >
+      Steering Committee Elections
+    </h3>
+    <p class="mb-2">
+      Every year, the Nix community elects members to the steering committee.
+      Each member holds a two-year term, with elections held annually to fill
+      half of the seats. You can read more about the election process in the <a
```
infinisil:
```suggestion
      at least half of the seats. You can read more about the election process in the <a
```

core/src/pages/governance.astro
```diff
@@ -0,0 +1,315 @@
...
+    </h2>
+    <p class="mb-6">The Nix community is guided by two leadership bodies:</p>
+    <div class="mt-2 mb-3 grid gap-4 md:grid-cols-2">
+      <div
+        class="bg-secondary-afghani-blue-95 flex w-full flex-col gap-2 rounded-xl p-4"
+      >
+        <h3
+          class="font-heading text-secondary-afghani-blue text-3xl leading-none font-bold"
+        >
+          The Steering Committee
```
infinisil:
```suggestion
          The Steering Committee (SC)
```

Since it's used as an abbreviation so often

core/src/pages/governance.astro
```diff
@@ -0,0 +1,315 @@
...
+          7 elected individuals to represent the community
+        </span>
+        <span>
+          <b>Responsibilities:</b> Technical and community decisions and delegation
+        </span>
+        <span>
+          <b>Members:</b>
+          {
+            steeringCommittee.data.members
+              .map((member) => member.name)
```
infinisil: I think it would be better to use GitHub handles (with a link), because that's more recognisable

infinisil (changes requested):
Made some suggestions, but looking great otherwise!

core/src/pages/governance.astro
```diff
@@ -0,0 +1,315 @@
...
+          {foundationBoard.data.members.map((member) => member.name).join(', ')}
+        </span>
+        <a href="/community/teams/foundation-board">More Info</a>
+      </div>
+    </div>
+    <p>
+      Behind these bodies are many volunteers who contribute to the project in
+      various ways. Read more about our {teams.length} teams on the <a
+        href="/community">community page</a
+      >.
```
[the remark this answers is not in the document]
djacu: Woops something funny happened here @infinisil 
```
teams</a{teams.length},
```

core/src/pages/governance.astro
djacu: @infinisil I think you meant like this?
```suggestion
      at least half of the seats. You can read more about the election process in the <a
```

core/src/pages/governance.astro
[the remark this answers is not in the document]
djacu: Is this what you meant?

```suggestion
      Local decisions are made in a distributed way by the <a href="/community">{teams.length} teams</a>, among many other ones not listed on the website and many more individual specialists, responsible for various areas.
      From the leadership bodies down to individual contributors, almost everybody is here as a passionate volunteer.
```

djacu (approved):
Looks great! Left some comments on the suggestions

core/src/pages/governance.astro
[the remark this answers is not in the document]
infinisil: Whoops yes!

core/src/pages/governance.astro
infinisil: Yes thanks! I think I tested some new review UI that was kind of broken, because I'm pretty sure I selected both lines for the suggestion 😆

core/src/pages/governance.astro
```diff
@@ -0,0 +1,354 @@
...
+        <span>
+          <b>Members:</b>
+          {
+            steeringCommittee.data.members.map((member, idx) => {
+              const data = getMemberLink(member);
+              if (data[0]) {
+                return (
+                  <>
+                    {' '}
+                    <a href={data[0]}>${data[1]}</a>
```
infinisil:
```suggestion
                    <a href={data[0]}>@{data[1]}</a>
```

I think this is the right place, should be a `@` for usernames, not `$` :)

core/src/pages/governance.astro
```diff
@@ -0,0 +1,354 @@
...
+        <span>
+          <b>Members:</b>
+          {
+            foundationBoard.data.members.map((member, idx) => {
+              const data = getMemberLink(member);
+              if (data[0]) {
+                return (
+                  <>
+                    {' '}
+                    <a href={data[0]}>${data[1]}</a>
```
infinisil:
```suggestion
                    <a href={data[0]}>@{data[1]}</a>
```

infinisil (approved):
Looking good other than a minor issue!

I'd also like to ping @NixOS/steering to let them know of this

JulienMalka:
I don't know exactly why, but right now the page shows the incorrect names for SC members.

thilobillerbeck:
> I don't know exactly why, but right now the page shows the incorrect names for SC members.

This has been branched off before the sc teams page has been updated. Since the display of the SC members takes the data from the same source as the sc teams page it is simply using outdated data which will be no problem when main is either merged into this branch or this branch will at some point be merged into main. :)

thilobillerbeck:
@cafkafk I don't know how much a cleaned up commit stack would help you since this is pretty much about the end result and its contents only. This might have been missed but we usually **squash** our commits at the end with a proper commit message. This is why I do not really care about keeping a clean commit history in these PRs.

PR #1875, opened 2025-10-21
Update notice in banner for SC Election 2025.

djacu:
Banner now informs people that they can still register to vote and that
voting has started.

Update link to point to the repository README which has the latest
information.


djacu:
@thilobillerbeck I tried updating the election banner but in the deploy preview it reverts back to the nixos 25.05 release. I saw this locally too and thought it was just a local development bug. What am I doing wrong?

thilobillerbeck:
yeah, the PR workflow does not support editing banners. Maybe that could be an imporvement. Like, to detect if a banner was edited. Though I don't know how much of a vuln that could introduce.

PR #1914, opened 2025-12-11
Add Nix CN Meetup

jian-lin:
ref https://github.com/NixOS-CN/nixcn-web/issues/3

### DIFF
--- a/core/src/content/community/meetups.yaml
+++ b/core/src/content/community/meetups.yaml
@@ -37,3 +37,5 @@ asia:
     location: Singapore (Singapore)
   - href: http://www.meetup.com/Tokyo-NixOS-Meetup/
     location: Tokyo (Japan)
+  - href: https://meetup.nixos-cn.org/
+    location: China


thilobillerbeck (changes requested):
Please make the location more specific as seen in the rest of the meetup section entries

jian-lin:
Thanks for your review.

I have a few things to clarify:
- These meetups are hosted by and for the Nix CN community which consisting of people from all over China.
- The first and second meetups are in Shanghai.  However, there is no fixed location for future meetups.  They can be anywhere in China.  The meetup locations are mainly decided by the volunteers hosting those meetups, a bit similar to how the location of NixCon is decided.

That's why I follow the precedent of Netherlands.

Do you still think we should make the location more specific?

thilobillerbeck:
Seems I was a bit tired while reviewing, sorry. Yeah, you got a point there :+1:

jian-lin:
Thanks! I should have explained why the location is not very specific in the beginning.  That way, we can reduce some review rounds.

PR #2027, opened 2026-05-21
Add Wiki link to navigation menu

therealladiesman217:
The header menu currently lacks a direct link to the NixOS Wiki.
Users had to navigate imo 2 much to find it.

This adds a Wiki entry to the top nav, linking to https://wiki.nixos.org/wiki/NixOS_Wiki.
The link opens in a new tab since the Wiki has its own navigation.

### DIFF
--- a/core/src/content/menus/header.yaml
+++ b/core/src/content/menus/header.yaml
@@ -5,6 +5,8 @@ items:
     link: /download
   - name: Learn
     link: /learn
+  - name: Wiki
+    link: https://wiki.nixos.org/wiki/NixOS_Wiki
   - name: Governance
     link: /governance
   - name: Community


thilobillerbeck:
Hey, thanks for contributing. :) I like the idea, but I have some gripes with this approach:

1. First of all I'd opt for keeping the top navigation as clean and easy to grasp as possible, keeping as little links in there as we can. If we want to have more links we should consider some kind of submenu to open when clicking or hovering on different items.
2. While all links with text in the header nav will direct you to s new part of the homepage, this link does not. It is a good idea to open the wiki in a new tab but there is no context for the user to know that this link does something completely different than the others.
3. Currently we treat these links as folders. If you are looking for information you know that "Learn" is the only option that will fit this intention, once in, you have all official resources listed and ready to be navigated to. If we now add "Wiki" to the menu, there will likely be confusion. People will ask themselves where they will find the information. In an abstract manner we would basically have two menu entries that convey kind of the same intention.
4. When adding the wiki, we will soon also likely have to add the manuals etc. The top nav will be bloated at some point and loose its intention to communicate the overall structure of information.

I'm happy to discuss a solution, but I am not sure if adding the link to the current state of the navbar is a good practice. :)

therealladiesman217:
Hi,

i understand your points. Im willing to do a bit more substancial rework for some parts of the webpage, mainly the parts that are for new users. I found the page to be a bit frustrating, esp. the "learn" page, which has most prominently some Nix Package Manager stuff, not NixOS stuff. 

2. Regarding the indicator for 'new tab', there would be the option to add a symbol but that would need some changes to the header component.

4. I dont think you must link to manual additionally, as its pretty straight forward in the wiki. 

But i think it makes most sense to just do a rework of the "learn" page for what i want to achieve.

PR #2037, opened 2026-06-01
Automatically switch to pride theme in June

infinisil:
Was a bit surprised to learn that this had to be done manually!

PR #2040, opened 2026-06-03
Add deployement schedule for pride

Sigmanificient:
~~https://github.com/Sigmanificient/ci-test-delete-me/actions/runs/26915001049/job/79402465964~~

This pr was repurposed since it was duplicate of https://github.com/NixOS/nixos-homepage/pull/2041. See https://github.com/NixOS/nixos-homepage/pull/2041#discussion_r3352207881 for motivation

PR #2041, opened 2026-06-03
Fix prod deploy

infinisil:
Bash doesn't have `<=` apparently..?!

Follow-up to #2037, which broke the deploy: https://github.com/NixOS/nixos-homepage/actions/runs/26770861864/job/78909651322

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -52,7 +52,7 @@ jobs:
       - name: 'Build'
         run: |
           now=$(date +%s)
-          if [[ "$(date +%s --date="1 Jun")" <= "$now" && "$now" < "$(date +%s --date="15 July")" ]]; then
+          if [[ "$(date +%s --date="1 Jun")" < "$now" && "$now" < "$(date +%s --date="15 July")" ]]; then
             export THEME=pride
           else
             export THEME=default


.github/workflows/deploy-prod.yml
```diff
@@ -52,7 +52,7 @@ jobs:
       - name: 'Build'
         run: |
           now=$(date +%s)
-          if [[ "$(date +%s --date="1 Jun")" <= "$now" && "$now" < "$(date +%s --date="15 July")" ]]; then
+          if [[ "$(date +%s --date="1 Jun")" < "$now" && "$now" < "$(date +%s --date="15 July")" ]]; then
```
Sigmanificient, line 55: Shouldn't that skip Jun 1st tho? :o

infinisil: `date +%s` gives the current _seconds_ since unix epoch, so it's only one second. However in any case it will only be switched on the first CI deploy after that, which probably won't happen at 00:00:00 anyways, so with the current code we'll always be an indeterminate amount of time late :sweat_smile:

Sigmanificient:
```
on:
  schedule:
    - cron: '0 0 1 6 *'
    - cron: '0 0 15 7 *'
```
[https://crontab.guru/#0_0_15_7_*](https://crontab.guru/#0_0_15_7_*) :eyes:

infinisil: Heck yeah! Probably better in a separate PR though

Sigmanificient: Repurposed https://github.com/NixOS/nixos-homepage/pull/2040

Sigmanificient (approved):
Tested on https://github.com/Sigmanificient/ci-test-delete-me/actions/runs/26915394999/job/79403721230

thilobillerbeck (approved):

PR #2051, opened 2026-06-18
Add copy button to codeblocks

thilobillerbeck:
This PR adds a copy button to code lines and blocks that will be hydrated into the homepage when JS is enabled. If JS is disabled, the copy button will not be visible so no distraction is created.

Oh, and it removes htmx since I stumbled over it...


djacu:
uh oh
image

if i disable JS the code blocks look correct

thilobillerbeck:
whoops, yeah, I'll have a look

djacu:
Looks good on the how-nix-works page.

djacu (approved):
how-nix-works page looks good
approved pending the styling on the branding page is fixed

PR #2066, opened 2026-07-05
style(why-nix): clarify mail CTA

Sigmanificient:
I dont understand how i can fix the button padding other than using a `style` like this


thilobillerbeck:
thx. ftfy.

Btw. you can find more info on tailwind padding utilities here: https://tailwindcss.com/docs/padding :)

thilobillerbeck (approved):

PR #2068, opened 2026-07-08
Serve the NixOS community survey site under /surveys

djacu:
Integrates the [NixOS/surveys](https://github.com/NixOS/surveys) site into the homepage so survey results are served at `/surveys`.


thilobillerbeck:
I feel like we should restructure the footer next :D The distribution is quite uneven now haha

thilobillerbeck (approved):

PR #2085, opened 2026-07-31
content: add Bluesky to social links

orpheuslummis:
The NixOS Foundation posts on Bluesky as @nixos.org, but the account isn't linked anywhere on the site. Add it to the two places that list the project's social accounts: the footer's "Connect with us" icon row and the "Social media" section on /community.

The handle is domain-verified against nixos.org, so it can be trusted as the official account. simple-icons ships a `bluesky` icon, so the footer's name-derived icon lookup resolves without further changes.

### DIFF
--- a/core/src/content/menus/footer.yaml
+++ b/core/src/content/menus/footer.yaml
@@ -44,6 +44,8 @@ sections:
       - name: NixOS Weekly (Archive)
         link: https://weekly.nixos.org/
 social:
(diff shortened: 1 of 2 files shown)


cafkafk (commented):
i approve of the idea (will give proper approval when CI etc works)

thilobillerbeck (approved):

PR #2106, opened 2026-09-07
community: rework the teams cards

djacu:
Having "Read more" buttons on all the teams is silly. Turn all the teams into their own card which is now clickable and has similar style to other cards on the page.


thilobillerbeck:
I like the change.
Can we somehow convey they are clickable? Currently they don't convey that at all.

fricklerhandwerk:
@thilobillerbeck Good point. I juggled the background colors a bit in #2111

image

core/src/components/pages/community/TeamLinkItem.astro
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const { href, iconSource, name, description } = Astro.props;
+---
+
+<li
+  class:list={[
+    'relative flex-1 cursor-pointer rounded-xl p-4 transition-all',
+    'bg-secondary-afghani-blue-35 hover:bg-secondary-afghani-blue-25',
+    'before:icon-[mdi--external-link] before:block before:absolute before:w-6 before:h-6 before:top-6 before:right-6 before:text-primary-white',
```
fricklerhandwerk: Hm but that conveys "this is an external page", which it clearly isn't. And none of the other cards have that.

fricklerhandwerk: I guess the link icon is OK, though it does go beyond the scope of this change. Ideally we can add the right icons to all the links in one go in a different PR, but if you prefer this being a pilot, sure.

thilobillerbeck: My problem ist that even though the surfaces are dark now, there is no real clue that this is not a normal card. That is what the icon should convey. We could also use the chevron used on the link buttons, that'd be consistent

fricklerhandwerk: What I'm saying is that none of the other cards have this and they're all clickable just the same. I don't disagree that an indication would be good, I'm just saying maybe this is a different conversation than the one this PR is about.

thilobillerbeck: That's fair, let's discuss this in another PR

thilobillerbeck (approved):

